PR #136, opened 2020-10-15
Added first implementation of GDumb and test

AndreaCossu:
This is just a first implementation to showcase the way GDumb it is implemented. It passes the related test but I did not check any further.

### DIFF
--- a/tests/test_strategies.py
+++ b/tests/test_strategies.py
@@ -96,11 +96,11 @@ class StrategyTest(unittest.TestCase):
 
     def test_gdumb(self):
         model = SimpleMLP()
-        optimizer = SGD(model.parameters(), lr=1e-3)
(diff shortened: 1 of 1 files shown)


kaustubholpadkar:
> Retrain from scratch was missing in the test. Up to now it is user responsibility to recreate the model+optimizer in the loop over steps.
> I tested and performance is ok, you can merge.

Can you please add these instructions in the documentation that the user needs to recreate model and optimizer for each experience? After few hours of struggle, finally from this PR discussion I realized it. Adding this to documentation will save time for many users in future. 

PS. Great Job adding this method. Thank you!

AntonioCarta:
Thanks @kaustubholpadkar for bringing this up. I think we have to fix the strategy, adding it in the documentation is not enough. The strategy itself should reset the model after each new experience (or at least before eval is called).

PR #355, opened 2021-03-03
Added Max GPU / Disk Usage metrics 

AndreaCossu:
This PR introduces GPU metrics, following the same structure used for the other metrics.
The metric monitors the max GPU usage of a specific GPU provided by the user. It uses `gputil` package (installed with conda-forge) to get the usage percentage. A separate thread read the usage every `every` seconds. 
As for RAM metrics, activating this metric on each minibatch is discouraged since it can cause up to `every` seconds delay after each minibatch. I'll update the documentation accordingly after all metrics are ready.

There are also minor fixes in the previous metrics.
The eval plugin example has been updated.


AndreaCossu:
I added the Disk Usage metric. I removed support for monitoring disk I/O operations since I want this metric to be separated from `DiskUsage`, which monitors the size of certain disk paths.
The problem of delay between callbacks should now be solved. The delay occurs only after the entire training or eval phase (@AntonioCarta thanks!).

vlomonaco:
can you also add `gputil` to the setup.py so that it is installed as a dependency also in pip? thanks :)
Also, some tests are not passing it seems...

AndreaCossu:
I added `gputil` to setup, tests should now pass

PR #391, opened 2021-03-12
Updating Quick Example

christam96:
Update imports so code does not break.


vlomonaco:
Thank you @christam96! Much appreciated! :) 🥇

Issue #277, opened 2021-02-10
Make "strategies" a module


lrzpellegrini:
In the same spirit of #246, I think that `training/strategies/strategies.py` is growing too fast to keep all the strategies in the same file.

I propose to split `strategies.py` into several scripts, each containing a single strategy.

Issue #246, opened 2021-01-25
Make "plugins" a module


lrzpellegrini:
As of now, plugins implementation are all contained in a single script `plugins.py`, whose dimension is growing fast.

I propose to make plugins a module by splitting `plugins.py` into several scripts.

PR #415, opened 2021-03-15
FIX data loader and refactor plugins

### LINKED ISSUE #393: MultiTaskMultiBatchDataLoader does not reinitialize the data loaders
Using the replay plugin with `mem_size = 100`, together with the Naive strategy I noticed that when the `MultiTaskMultiBatchDataLoader` wants to reinitialize one of the data loaders (https://github.com/ContinualAI/avalanche/blob/master/avalanche/benchmarks/utils/data_loader.py#L129), it fails to do so. In fact, by monitoring the training using `train_mb_size = 16`, each epoch ends after no more than 10 mini-batches, namely after all the examples of the external memory have been used.

AntonioCarta:
Closes #393 #246 #277


AntonioCarta:
fixed.

PR #433, opened 2021-03-18
added joinedBatches dataloader and improved ReplayPlugin

### LINKED ISSUE #336: Alternative version / replacement of Replay Plugin
Currently, Replay plugin keeps an external memory of patterns which are concatenated to the current step dataset. Minibatches are then sampled from this new dataset.

We could provide an additional Replay version or substitute the current one by concatenating to each _minibatch_ a random subset of the external memory. In this way, there is no need to concatenate to the entire dataset and replay is more effective (even if it costs more in terms of computation). 
This can be done in `before_forward` which is only applied to the training phase.

MattiaSangermano:
Closes #336


AntonioCarta:
Thanks @MattiaSangermano.

If I understand your solution correctly, it seems that your solution changes the task label of the buffer to avoid clashes with the new data. However, changing the task labels may break some models (e.g. multi-heads). A possible solution would be to change the dataloader to take as input two separate datasets (new data and buffer) which will be iterated in parallel. The dataloader will need to concatenate batches from the new data and buffer together. Basically, the concatenation should happen inside the data loader and not the plugin, which will only instantiate the correct dataloader. Of course other solutions are possible, this is just a suggestion.

There are also a couple of things that may be useful:
- Can you update the `ReplayPlugin` docstring to give a high-level description of the new behaviour?
- It would be useful to have a couple of tests for the dataloader and plugin to check that the data is balanced correctly in different scenarios. There are lots of scenarios so it's easy to break things subtly.

MattiaSangermano:
> If I understand your solution correctly, it seems that your solution changes the task label of the buffer to avoid clashes with the new data. 

If you mean the task label in the `adapted_dataset`, yes the task label is changed only if the buffer has the same task id as the training data, i.e. in a class-incremental scenario.

> A possible solution would be to change the dataloader to take as input two separate datasets (new data and buffer) which will be iterated in parallel.

Do you mean two input dictionaries (the one that maps task_ids into datasets), one for the training data and one for the buffer? 

> Basically, the concatenation should happen inside the data loader and not the plugin, which will only instantiate the correct dataloader.

What do you mean by this? Because the concatenation already takes place inside the dataloader. What I do in the plugin is to prepare the `adapted_dataset` which currently contains both memories and training data.

MattiaSangermano:
I was looking at how the replay plugin memory changes over the experiences in a scenario where task_id is returned (I'm talking about the code before this PR). I noticed that the number of examples stored in memory for the different tasks is not the same. The pattern is the following:
- `task_id` 0: mem_size
- `task_id` 1: mem_size / 2
- `task_id` 2: mem_size / 3
and so on. 

Shouldn't every task have the same number of patterns in memory?  
If so I can change it in this PR

AntonioCarta:
@MattiaSangermano are you done? Looks complete to me.

MattiaSangermano:
@AntonioCarta Yes I'm done

AntonioCarta:
Thanks @MattiaSangermano.

PR #442, opened 2021-03-23
Explicit support for classification setups. Extended support for tensor scenario.

lrzpellegrini:
This PR brings an extended support for non-int targets for regression, segmentation and any other setup.

In addition, the new parameter `dataset_type` in `AvalancheDataset` allows the user to explicitly mark a dataset with one of the types CLASSIFICATION, REGRESSION, SEGMENTATION, UNDEFINED so that the conversion of targets can be better handled. `AvalancheDataset` currently supports the conversion of targets for the classification setup, while choosing other data types does not affect the targets field.

`AvalancheTensorDataset`, helpers and generators handling the creation of benchmarks instances from tensors have been extended to allow for more that 2 features (x, y, z, ...). Previously, only "x" and "y" were supported.

Added 2 deprecation warnings for legacy tensor scenario helpers.


AntonioCarta:
Thanks, looks good. Can I ask what is the purpose of the `dataset_type`? From a quick glance, it seems that as long as you provide a custom `collate_fn` and `targets_adapter` you should be fine.

lrzpellegrini:
> From a quick glance, it seems that as long as you provide a custom `collate_fn` and `targets_adapter` you should be fine.

Yes, the `dataset_type` is not particularly useful right now as it only "does something" when the type is `CLASSIFICATION`, but it may become useful when implementing object detection support. Providing a custom `collate_fn` and `targets_adapter` each time you create a dataset may make the manipulation of `AvalancheDataset`s a little bit tedious.

The idea was to allow the user to define a `dataset_type` (also by calling helpers like `as_classification_dataset`, `as_segmentation_dataset`, ...) that automatically defines `collate_fn` and `targets_adapter` if they are not explicitly provided. In this way the user can still set a custom `collate_fn` and `targets_adapter`, but I expect the vast majority of users to just provide the `dataset_type` that will do the job for them.

AntonioCarta:
Ok. If I understand correctly, the `dataset_type` is a default configuration for specific tasks that sets the correct functions. It seems a good idea. Maybe we should disallow setting `collate_fn` and `targets_adapter` when the user sets `dataset_type`?

Anyway, I think the PR can be merged.

lrzpellegrini:
Yes, makes sense to make them mutually exclusive! I'll add a commit to enforce this before merging the PR.

vlomonaco:
Ok @lrzpellegrini great job! Let us know when to merge then!

lrzpellegrini:
I updated the PR. AvalancheDataset and sub-classes now enforce the mutual exclusion between `dataset_type` and `collate_fn`+`targets_adapter`. I also added few unit tests. Unit tests are always good ;)

lrzpellegrini:
If it's ok for you, I think we can merge this

PR #444, opened 2021-03-24
Updated metric dictionary returned by the strategy on train and eval

### LINKED ISSUE #420: Add examples on how to use the standalone metrics
This is helpful to explain how to use metric off-the-shelf without integrating them in training and evaluation loops.

AndreaCossu:
This PR changes the way metric results are returned to the user after `train` and `eval` methods are called.
In case of `train`, the method returns a dictionary containing two keys: `train` and `eval`. Inside `train` the user finds a dictionary of training metrics: key is metric name, value is the last value recorded for that metric. Inside `eval` the user finds a list of dictionaries, one for each evaluation stream specified by the user in the `train` method call. Each dictionary reports has metric name as key and last recorded value as metric value.

If `collect_all` is set to True in the `EvaluationPlugin`, the user can access all the metrics by calling the method `get_all_metrics` of the `EvaluationPlugin`. 

I also updated the `eval_plugin` example.

This also closes #420 since I added an example on how to use standalone metrics.


AntonioCarta:
I think it's better to avoid being too verbose, which is already a problem with metric names. Maybe instead of a dictionary with two keys we can return a single dictionary (since metric names are unique) or a tuple `train_res, eval_res`?

AndreaCossu:
The problem with a single dict is being able to quickly retrieve metric computed during evaluation or training. The user should filter the dictionary by looking at phase name, which is not ideal in my opinion. Having two outputs instead of a single dict could cause compatibility problems if we need to add more outputs in the future. Apart from this reason I don't have strong preferences on the type of output reported by `train`.

vlomonaco:
A single dict with a flat list of metrics results makes sense to me, but I don't have a strong preference... let's just make sure this is well documented with examples as well.

AntonioCarta:
> The problem with a single dict is being able to quickly retrieve metric computed during evaluation or training. 

I don't get this point. The majority of the usecases will be:
- save all metrics to process them later.
- plot a *single* curve/value.

Either way, a single dictionary is enough. It is also possible to split them, but return a tuple `train_res, eval_res`. I don't like using a dictionary when it's not strictly necessary (breaks IDE, more verbose, harder to understand/debug than a tuple). However, this is mostly a stylistic preference.

AndreaCossu:
Ok, its' fine by me. I'll try and implement a single dictionary with the last recorded value for each metric.

AndreaCossu:
Done! If you don't have any additional comments, I'll merge this PR tomorrow.

Issue #472, opened 2021-03-31
Consistent Download Location for Benchmarks Data


vlomonaco:
See feature-request here: https://github.com/ContinualAI/avalanche/discussions/470

ggraffieti:
I'll take care of this

PR #477, opened 2021-03-31
Dataset download only in ~/.avalanche/data folder

ggraffieti:
All the dataset now are downloaded into `~/.avalanche/data` folder.
This solve #470 and #472


vlomonaco:
It looks good to me, merging now!

Issue #463, opened 2021-03-30
Lazy scenario construction for Permuted MNIST


AntonioCarta:
I'm trying to run PMNIST using a large number of experiences. This is slow because all the permutations are built inside the constructor by copying and permuting the data repeatedly. Can we speed up this process with lazy computations by computing them only as needed? 

Maybe there is also a general need for some support for lazy generation of experiences whenever there are a large number of them.

Mattdl:
I've experienced the same issue, also with RotatedMNIST. I think we can just use a fixed transform (permutation for PMNIST or rotation for RMNIST) which is applied by the generator when fetching the data.

PR #478, opened 2021-03-31
Performance improvement for AvalancheConcatDataset. Prevent dataset duplication in MNIST benchmarks.

lrzpellegrini:
A small performance improvement for `AvalancheConcatDataset`, which will now use bisect to obtain the correct dataset to use (similar to PyTorch `ConcatDataset`)


lrzpellegrini:
I updated the PR to include a fix for #463. MNIST is no longer duplicated by Permuted/RotatedMNIST benchmarks!

vlomonaco:
Perfect! I will also add here a link to this feature-request which I think is relevant for the future: https://github.com/ContinualAI/avalanche/discussions/426

PR #481, opened 2021-04-01
Added global iteration counter for the metric x values

### LINKED ISSUE #375: logging with global "x_values"
Metrics emit `<name, x, y>` values. Currently, `x` is updated inside the callback. Each callback can use its own `x` values. I propose to use a single global counter, incremented after every training iteratior. Basically, this global counter uniquely identify the model at a certain point in time. I think this is easier to implement and more intuitive (tensorflow does this automatically if I remember correctly). Also, now that we have the `eval` enabled every `k` training epochs, the `x` values depend on the frequency of the evaluation, which seems weird to me.

AndreaCossu:
This PR adds a global iteration counter which starts from zero and it is increased by 1 at every training or eval iteration. 
The counter is managed by the `PluginMetric` class. All metrics have been updated to use the global counter.
I also updated the examples with the recent `StreamForgetting` modifications.

The main modifications the user will perceive are:
- longer distances between a metric x value and the next one (depending on how many iterations have been performed between the two)
- larger absolute values for the x values (total number of iterations)

If you agree with this system we can merge the PR.

This closes #375


vlomonaco:
it looks good to me! Feel free to merge when the actions complete

Issue #482, opened 2021-04-01
Notebooks to Doc Automation


vlomonaco:
We should create a new action that:

1. Converts notebooks in the `notebooks` dir to .md
2. copy them to the GitBook doc.
3. Commit eventual changes in the mirrored notebooks on [colab](https://github.com/ContinualAI/colab).

PR #483, opened 2021-04-01
W&B Logger Documentation updates

digantamisra98:
Minor documentation changes for W&B logger in gitbook. Requesting edit access to the [demo colab notebook](https://colab.research.google.com/github/ContinualAI/colab/blob/master/notebooks/avalanche/loggers.ipynb) to showcase the functionalities for W&B.


AndreaCossu:
Thank you @digantamisra98. Currently, the From Zero to Hero tutorial is updated through the notebooks you see in the `notebooks` folder. Can you modify that notebook instead of directly the gitbook? Also, the colab notebooks will be kept in sync automatically once you modified the notebook here on avalanche. See #482

digantamisra98:
@AndreaCossu for the colab notebooks, I believe I will have to run them and save the cell output as well so it would make sense if I run them on a W&B Teams account for Continual AI, which would serve as the central dashboard.

PR #495, opened 2021-04-05
fix-agem. resolved memory filling and grad proj

JonasFrey96:
A-Gem is now working (fixed gradient projection computation) and nearly matching the reported performance.
Updated the example providing the new numbers.

Best Jonas

### DIFF
--- a/avalanche/training/plugins/agem.py
+++ b/avalanche/training/plugins/agem.py
@@ -53,22 +53,24 @@ class AGEMPlugin(StrategyPlugin):
         """
         Project gradient based on reference gradients
         """
-
         if self.memory_x is not None:
-            for (n1, p1), (n2, refg) in zip(strategy.model.named_parameters(),
-                                            self.reference_gradients):
+            current_gradients = [p.grad.view(-1)
+                for n, p in strategy.model.named_parameters() if p.requires_grad]
+            current_gradients = torch.cat(current_gradients)
 
-                assert n1 == n2, "Different model parameters in AGEM projection"
-                assert (p1.grad is not None and refg is not None) or \
-                       (p1.grad is None and refg is None)
+            assert current_gradients.shape == self.reference_gradients.shape "Different model parameters in AGEM projection"
 
-                if refg is None:
-                    continue
-
-                dotg = torch.dot(p1.grad.view(-1), refg.view(-1))
-                dotref = torch.dot(refg.view(-1), refg.view(-1))
-                if dotg < 0:
-                    p1.grad -= (dotg / dotref) * refg
+            dotg = torch.dot( current_gradients, self.reference_gradients)
+            if dotg < 0:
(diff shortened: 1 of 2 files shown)


AntonioCarta:
Thanks @JonasFrey96! your changes look good but there are some syntax errors. I will merge the code as soon as you fix them.

ggraffieti:
`avalanche/training/plugins/agem.py` line 61. 
A comma after the asset's comparison and before the message is missing. 
Fix the problem and the PR should be good to be integrated inside avalanche! :tada:

JonasFrey96:
Thanks a lot. Best Jonas

AndreaCossu:
Hi @JonasFrey96 . Checks are still failing. Please, run from the `avalanche` root folder the command `pycodestyle avalanche` to make your code PEP8 compliant. 
Then, move to the `tests` folder and run `USE_GPU=false; FAST_TEST=true python -m unittest`. These should show you the errors.

EdenBelouadah:
If you use Pycharm, it has an auto PEP-8 corrector, it can help you to check your mistakes and reformat the file.

ggraffieti:
Thanks, @EdenBelouadah for pointing this out, the pep8 checking of the project is configured to be exactly the same as pycharm. Maybe we could have mentioned that PyCharm is the preferred editor to use in a better way (maybe in the "How to contribute" section of the documentation).

Apart from that seems that there is also an error in a test: 
`File "/home/runner/work/avalanche/avalanche/avalanche/training/plugins/agem.py", line 61, in after_backward
    assert current_gradients.shape == self.reference_gradients.shape , "Different model parameters in AGEM projection"
AttributeError: 'list' object has no attribute 'shape'`

To run the test locally (and not have to rely only on GitHub actions) you can use this command: 
`FAST_TEST=True USE_GPU=False python -m unittest`

vlomonaco:
Hi @JonasFrey96, this PR has been pending for a while, do you plan to make you amendment soon? Otherwise we can take care of it. Thanks again for your help!

AndreaCossu:
I'll take care of polishing this up.

AndreaCossu:
I've fixed the PEP8 but it seems that I cannot push directly on the `fix-agem` branch of @JonasFrey96 's fork. I can open a PR from my fork. 

I also fixed the `self.reference_gradients` tensor list which was treated as a tensor, thus causing errors in `.shape` call. I don't know how the example has been run to update the accuracy values.
Moreover, I am not able to reproduce the performance stated in the example for SMNIST. I obtain ~30% less final average accuracy.

vlomonaco:
@AndreaCossu @AntonioCarta how do we want to proceed on this? I don't remember what we decided.

AndreaCossu:
We still have to convert A-GEM to use the multi-head models by using Avalanche custom model.

PR #516, opened 2021-04-08
Support for custom streams

### LINKED ISSUE #332: Multiple Streams per Scenario
As already discussed in #262 it would be nice to have also a `validation_stream` other than the train and test ones. 
However, it turns out some benchmarks like Stream-51 (#315) would need additional had-hoc streams. 

I was wondering if other than the standard train, valid, test streams it would be better to have an additional attribute in the scenario object with a list of additional streams.

### LINKED ISSUE #443: Untied size for train and test stream
At the moment Avalanche supports creating scenarios using a lot of helpers/generators. However, these tools constraint the available choice of streams (only "train" and "test" are supported, see #332) and the amount of experiences in the streams.

Currently the following combinations are supported:

- N train experiences, N test experiences
- N train experiences, 1 test experience

This not sufficiently general and doesn't allow for the implementation of various benchmarks that use a different amount of experiences in different streams (for instance, the CV part of A-GEM).

lrzpellegrini:
This PR implements the support for custom streams. Custom streams (validation, out-of-distribution, ...) are used by some benchmarks but as of now Avalanche only supports the 'train' and 'test' streams.

This also introduces the support for streams with varying lengths. Now Avalanche only supports train and test streams with the same amount of experiences (or 1 experience for test set). With this PR stream lengths will be unconstrained.

Resolves #332
Resolves #443

### DIFF
--- a/avalanche/benchmarks/scenarios/generic_cl_scenario.py
+++ b/avalanche/benchmarks/scenarios/generic_cl_scenario.py
@@ -1,97 +1,118 @@
 import copy
+import re
 from abc import ABC
 from typing import Generic, TypeVar, Union, Sequence, Callable, Optional, \
-    Dict, Any, Iterable, List, Set
+    Dict, Any, Iterable, List, Set, Iterator, Tuple, NamedTuple
+
+from torch.utils.data.dataset import Dataset
 
 from avalanche.benchmarks.scenarios.generic_definitions import \
-    TExperience, ScenarioStream, TScenarioStream, Experience, TScenario, \
-    TrainSet, TestSet
-from avalanche.benchmarks.utils import AvalancheDataset, \
-    AvalancheSubset
+    TExperience, ScenarioStream, TScenarioStream, Experience, TScenario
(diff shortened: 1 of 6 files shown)


vlomonaco:
This is great @lrzpellegrini! Can you update the doc accordingly? both the 5 minutes and the fzth from the notebooks! (just a few hints on this very useful features) thanks!

AntonioCarta:
Looks good to me. Thanks Lorenzo!

AndreaCossu:
Thanks @lrzpellegrini , I think we can merge :)

lrzpellegrini:
> This is great @lrzpellegrini! Can you update the doc accordingly? both the 5 minutes tutorial from gitbook and the fzth from the notebooks! (just a few hints on this very useful features) thanks!

I think I'll update them after this PR is merged. However, consider that the support for custom streams is now available through the :class:`GenericCLScenario` class but we still don't have a proper set of helper methods supporting it.

vlomonaco:
ok, I'll create a new issue to keep track of that

PR #518, opened 2021-04-09
Add histogram and figure logging, fixed confusion matrix bug

### LINKED ISSUE #513: StreamConfusionMatrix Stalling Problem
**Describe the bug**
The script provided below stalls (or it is super slow) to move to the next experience. 

**To Reproduce**

```
import torch
import torch.optim as optim
import torchvision.transforms as transforms
from avalanche.benchmarks.classic import CORe50
from avalanche.evaluation.metrics import ExperienceForgetting, accuracy_metrics, loss_metrics, timing_metrics, \
    cpu_usage_metrics, StreamConfusionMatrix, disk_usage_metrics
from avalanche.logging import InteractiveLogger, TextLogger, TensorboardLogger
from avalanche.training.plugins import EvaluationPlugin
from avalanche.training.strategies import Naive
from torchvision.models import resnet34

train_transform = transforms.Compose([
        transforms.ToTensor()
])
eval_transform = transforms.Compose([
    transforms.ToTensor()
])
scenario = CORe50(
    scenario="ni",
    train_transform=train_transform,
    eval_transform=eval_transform)
model = resnet34(pretrained=True)

tb_logger = TensorboardLogger(tb_log_dir="experiment")
interactive_logger = InteractiveLogger()

eval_plugin = EvaluationPlugin(
    accuracy_metrics(minibatch=True, epoch=True, epoch_running=True, experience=True, stream=True),
    loss_metrics(minibatch=True, epoch=True, epoch_running=True, experience=True, stream=True),
    StreamConfusionMatrix(num_classes=50, save_image=True), # <---- disable this to make the problem disappear
    loggers=[interactive_logger, tb_logger]
)
optimizer = optim.SGD(model.parameters(), lr=1e-3, momentum=0.9)
criterion = torch.nn.CrossEntropyLoss()
batch_size = 16
epochs = 1
device = torch.device("cuda")

cl_strategy = Naive(
        model=model,
        optimizer=optimizer,
        criterion=criterion,
        train_mb_size=batch_size,
        train_epochs=epochs,
        eval_mb_size=batch_size,
        evaluator=eval_plugin,
        device=device
)

print("Starting experiment...")
print(len(scenario.train_stream))

for experience in scenario.train_stream[:2]:
    print("Start of experience ", experience.current_experience)
    print("Current classes: ", experience.classes_in_this_experience)

    cl_strategy.train(experience, num_workers=8)
    print("Training complete.")

    print("Computing accuracy on the test set")
    cl_strategy.eval(scenario.test_stream, num_workers=8)
    print("End of experience ", experience.current_experience)
```

**Expected behavior**
The script should go over the next experience immediately. 

**Additional context**
See discussion here: https://github.com/ContinualAI/avalanche/discussions/439#discussioncomment-585228

AndreaCossu:
This PR brings the following changes:

1. WandbLogger and Tensorboard logger now logs also matplotlib Figure and histogram
2. Confusion matrix is logged as a figure, not image. By default, the confusion matrix does not show numbers in the cell, to accommodate for medium / large number of classes. 
3. Confusion matrix now uses `num_classes` to fix the size of the confusion matrix for all the training and evaluation. This value is strict. The class documentation has been updated to explain the change.
4. The evaluation logger returns the plain metric dictionary instead of a copy. This is because matplotlib Figures cannot be deepcopied. This should not be a problem because loggers logs metrics instantly.

This closes #513

### DIFF
--- a/avalanche/evaluation/metric_utils.py
+++ b/avalanche/evaluation/metric_utils.py
@@ -13,6 +13,7 @@ import io
 from typing import Tuple, TYPE_CHECKING
 
 import matplotlib.pyplot as plt
+import numpy as np
 from PIL import Image
 from sklearn.metrics import ConfusionMatrixDisplay
 from torch import Tensor
@@ -29,19 +30,16 @@ TRAIN = "train"
 
 def default_cm_image_creator(confusion_matrix_tensor: Tensor,
                              display_labels=None,
-                             include_values=True,
-                             xticks_rotation='horizontal',
+                             include_values=False,
+                             xticks_rotation=0,
+                             yticks_rotation=0,
                              values_format=None,
                              cmap='viridis',
-                             dpi=100,
                              image_title=''):
     """
-    The default Confusion Matrix image creator. This utility uses Scikit-learn
-    `ConfusionMatrixDisplay` to create the matplotlib figure. The figure
-    is then converted to a PIL `Image`.
-
-    For more info about the accepted graphic parameters,
-    see `Scikit learn <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.plot_confusion_matrix.html>`_ # noqa
+    The default Confusion Matrix image creator.
+    Code adapted from
+    `Scikit learn <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.plot_confusion_matrix.html>`_ # noqa
 
     :param confusion_matrix_tensor: The tensor describing the confusion matrix.
         This can be easily obtained through Scikit-learn `confusion_matrix`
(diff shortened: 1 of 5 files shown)


AntonioCarta:
> The evaluation logger returns the plain metric dictionary instead of a copy

Doesn't this breaks collecting the metrics iteratively? For example `[strategy.train(step) for step in scenario.train_stream]` does not work anymore because the dictionary inside the list are all the same object.

AndreaCossu:
Yes, that way of accumulating metrics would not be supported. The user could either use the `all_metric_results` when available, or manually copying the dictionary if he wants to accumulate it for future use.  
Alternatively, we could remove the matplotlib Figure from the dictionary and then deepcoping it (maybe keeping its dictionary key and adding None as value, just to keep track of which metrics are actually computed).  
Do you see other solutions?

AndreaCossu:
Maybe we could `copy` instead of `deepcopy`. This works also for Figure, and should not introduce bugs unless we modify nested metric values in place. Currently we only re-assign values to the keys, which should be preserved by `copy`. What do you think?

AntonioCarta:
`copy` should be enough.

Issue #490, opened 2021-04-02
SplitFMNIST error when enumerating the dataset


AndreaCossu:
On `SplitFMnist` is not possible to loop over the experience dataset.

To reproduce:
```
from avalanche.benchmarks import SplitFMNIST

scenario = SplitFMnist(5)
dataset = scenario.train_stream[0].dataset
for el in dataset:
  pass
```

returns 

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/avalanche_dataset.py", line 306, in __getitem__
    return TupleTLabel(manage_advanced_indexing(
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/dataset_utils.py", line 320, in manage_advanced_indexing
    single_element = single_element_getter(int(single_idx))
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/avalanche_dataset.py", line 1035, in _get_single_item
    return self._process_pattern(self._dataset[idx], idx)
  File "/home/cossu/miniconda3/envs/avalanche-env/lib/python3.9/site-packages/torch/utils/data/dataset.py", line 272, in __getitem__
    return self.dataset[self.indices[idx]]
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/avalanche_dataset.py", line 306, in __getitem__
    return TupleTLabel(manage_advanced_indexing(
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/dataset_utils.py", line 320, in manage_advanced_indexing
    single_element = single_element_getter(int(single_idx))
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/avalanche_dataset.py", line 659, in _get_single_item
    return self._process_pattern(self._dataset[idx], idx)
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/avalanche_dataset.py", line 306, in __getitem__
    return TupleTLabel(manage_advanced_indexing(
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/dataset_utils.py", line 320, in manage_advanced_indexing
    single_element = single_element_getter(int(single_idx))
  File "/home/cossu/avalanche/avalanche/benchmarks/utils/avalanche_dataset.py", line 1035, in _get_single_item
    return self._process_pattern(self._dataset[idx], idx)
[...]

PR #529, opened 2021-04-13
Fixed FMNIST transformations

AndreaCossu:
FMNIST transformations were the ones used in CIFAR-10. They produced an error when looping through the dataset. I kept only the `ToTensor` transformation, since I don't think FMNIST requires specific preprocessing.
@lrzpellegrini do I need to add also a normalization like in MNIST? If so, do you have the normalization coefficients? You can merge the PR as soon as you are ok with the changes. Let me know if there are any issues.

This solves #490

### DIFF
--- a/avalanche/benchmarks/classic/cfashion_mnist.py
+++ b/avalanche/benchmarks/classic/cfashion_mnist.py
@@ -20,18 +20,12 @@ from torchvision import transforms
 from avalanche.benchmarks import nc_scenario
 from avalanche.benchmarks.utils import train_eval_avalanche_datasets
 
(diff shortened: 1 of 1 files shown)


lrzpellegrini:
Hi @AndreaCossu, I found this for FMNIST: https://github.com/facebookresearch/Adversarial-Continual-Learning/blob/master/src/dataloaders/mulitidatasets.py

It seems that mean = 0.2190 and std = 0.3318. Can you add this normalization after ToTensor? An example should be in that script, lines 149-150.

AndreaCossu:
I'll do that, thanks!

lrzpellegrini (approved):

PR #535, opened 2021-04-15
New Model API

### LINKED ISSUE #189: Remove model as input of MultiHeadPlugin
at the moment the `MultiHeadPlugin` needs the model as a parameter.
However, it can recover it from the strategy after the first call, which has the model reference. 

I think it would be nicer to not pass the model as a parameter to the plugin and eventually make its checks (like the name of the output layer) on-the-fly when an inner method is called. What do you think?

### LINKED ISSUE #328: separate incremental classifier and multi head functionality
From our discussion in #306. 

**Steps**:
- [ ] separate head expansion and multi-head functionality.
- [ ] switch to head expansion in the examples with single-incremental tasks.

If anyone wants to tackle I can support you.

### LINKED ISSUE #468: Potential Issue w/ AGEM with multi_head plugin
**Description**
When looking at the code for agem.py (in the before_training_iteration function),  it appears that the reference gradient is computed w.r.t the head of the current task as opposed to being computed w.r.t to the head of corresponding the the task to which each sample belongs.  It does not seem like the before_training_iteration checks whether or not the multi_head plugin is present.

In this code snippet (from the agem before_training_iteration function), the task descriptor for each member of x_ref is not accounted for and it doesn't seem like there is a check  for the multi_head plugin anywhere:

image

I think that this is inconsistent with the original AGEM implementation, and may therefore lead to a degradation in performance.

### LINKED ISSUE #523: New Models API
As you noticed, we have some issues with the `MultiHeadPlugin`. The source of this issue is the dynamic change of the head using the labels. Right now, this is done by the plugin. This means that if you call `model(x)` without calling the plugin's callback you get the wrong output because the model does not change the head.

As a solution, I would like to make a custom `DynamicModule`. This is a pytorch `nn.Module` with two additional functionalities.
- a `model_adaptation` phase to add new heads, units and make other architectural changes.
- a `forward` methods that takes `task_labels` as an additional argument.

This approach makes it easy to define dynamic architectures as common in continual learning, such as multi-head or architectural methods that require incremental expansion/freezing of units (e.g. PNN).

There is an important caveat here. Since dynamic modules take an additional argument we can't use standard pytorch modules and substitute the classifier anymore. This is the current API:
```
    model = SimpleMLP(input_size=32*32*3, num_classes=10)
    strategy = Naive(model=model, optimizer=optimizer, criterion=criterion, plugins=[MultiHeadPlugin])
```

This is what will happen with the proposed change:
```
    class MTModel:
        def __init__(feature_extractor):
             self.feature_extractor = feature_extractor
             self.classifier = MultiHeadClassifier(...)

        def forward(x, task_labels):
            x = self.feature_extractor(x)
            return self.classifier(x, task_labels)

    model = MTModel(feature_extractor)
    strategy = Naive(model=model, optimizer=optimizer, criterion=criterion, plugins=[MultiHeadPlugin])
```

The advantages are that:
- multi-task and dynamic models become standalone.
- you can save the model after training with `torch.save` and use `model(x, task_labels)` without the strategy.
- better separation of code which will probably simplify the `BaseStrategy`.
- easier to understand, less bugs that are hard to discover. If you don't pass the `task_labels` you get an error instead of silent feature like we currently have.

The disadvantages are:
- you can't use a pre-made model as is.
- you need to implement a custom forward. It should be easy and mostly boilerplate code but it may be slightly more than we have right now.

Let me know if I should progress on this.

AntonioCarta:
This large PR proposes an overhaul of the current models. It provides classes and utilities for:
- implementing dynamic modules i.e. module that change their architecture over time (new classification heads, adding/freezing units...)
- reset of the optimizer's state to support dynamic modules
- support inside the `BaseStrategy` for the model's adaptation phase and optimizer's update
- multi-task modules now support sample-level task labels, in line with `AvalancheDataset`
- splitting multi-task minibatches is done automatically, users only implement `single_task_forward`, which is usually much easier to write.

Future work:
- better support for implicit pass of task labels, which should make it easier to seamlessly convert single task models into multi-task models.
- now that multi-task models support sample-level labels, we should simplify the dataloaders to return a single mini-batch.

I'll leave this open for discussion. If there are no remarks, I will merge it in a couple of days.

Closes #523 
Closes #468
Closes #328
Closes #189.


vlomonaco:
Impressive work, thanks @AntonioCarta! Can we add also something about this in the models documentation?

As I understand, native pythorch models at the moment cannot be used anymore with the multi-head plugin.
You are forced to change the forward pass, inherit from `MultiTaskModule` and use `MultiHeadClassifier`, correct?

We should clarify this as well into the training doc I believe.

AntonioCarta:
@vlomonaco yes, but I need some time to update the notebook. I will open an issue about it.

AndreaCossu:
Let's try this out :smile:

vlomonaco:
Ok then, merge it when you are ready @AntonioCarta!

Issue #543, opened 2021-04-19
Possible Replay bug


Mattdl:
Hi all,

I'm working on implementing Class-Balanced and Experience-balanced Replay (See #479).
I want to provide the opportunity for the user to select a fixed capacity prior or adaptive to all seen experiences so far.

However, when looking at the Replay code I find it hard to follow: 

    single_task_mem_size = min(self.mem_size, len(curr_data))
    h = single_task_mem_size // (strategy.training_exp_counter + 1)

    remaining_example = single_task_mem_size % (
        strategy.training_exp_counter + 1)
    # We recover it using the random_split method and getting rid of the
    # second split.
    rm_add, _ = random_split(
        curr_data, [h, len(curr_data) - h]
    )

In the first line:
https://github.com/ContinualAI/avalanche/blob/9cf3c53d83ffc1b3dfe4400d43356ebcd901cfc9/avalanche/training/plugins/replay.py#L64
What if len(curr_data) is smaller than self.mem_size? It would occur to me that 'h' is not correct anymore. Even though the current data length ` len(curr_data)` may not equal the full memory size, it can still have capacity for a  `mem_size/n_observed_exps` portion of the memory.
So `h` should become `self.mem_size // (strategy.training_exp_counter + 1)`?

When taking the random_split however, then we should take the 'min' operation into account:

     h2 = min(capacity_per_exp, len(curr_data))
      rm_add, _ = random_split(
          curr_data, [h2, len(curr_data) - h2]
      )

Am I missing something here or is this a bug?

AntonioCarta:
If I understand correctly the bug happens only whenever `len(curr_data) < mem_size and len(curr_data) > (mem_size / (strategy.training_exp_counter + 1))`. Definitely a bug, we should fix it and add a test to cover this corner case.

vlomonaco:
Thanks @Mattdl for reporting, if you have already a general fix in place, would you be able to tackle this opening a PR? thanks :)

Mattdl:
Yes sure, I'll add the fix as soon as possible! (:
I'm also implementing the ExperienceBalancedStoragePolicy and ClassBalancedStoragePolicy. Should I incorporate those at once?

AntonioCarta:
A single PR is ok.

Mattdl:
I'm finishing up on the PR, but one issue remains if you want to update the replay memory in each minibatch.
[...]

Issue #360, opened 2021-03-04
Support for data incremental scenario


AntonioCarta:
We need to support the data incremental scenario with no task labels, no task boundaries. The problem is that experiences always have boundaries between them. After a bit of discussion I think there are two possible solutions:

1 - **single experience**, where the entire scenario is a single `Experience` carefully loaded by a data loader that does not reuse data.
2 - **a separate experience for each mini-batch**, where boundaries happen at the mini-batch level, which makes each mini-batch independent.

Personally, I prefer option (2). It seems natural to me to represent the passage of time using the stream of experiences. When you have a stream of experiences, it's clear that you either store them in a buffer or you will not encounter anymore. Option (1) is possible, and maybe slightly more efficient, but it's more error-prone. If you don't use a proper data loader it's not data incremental anymore. If you have multiple epoch it's not data incremental. And so on.

@lrzpellegrini what do you think? Is it possible to have such a large number of experiences?

@Mattdl what do you think of this solution?

See also [previous discussion](https://github.com/ContinualAI/avalanche/issues/336#issuecomment-789172633).

Mattdl:
That's a great solution @AntonioCarta , I would also prefer option 2. The problem is though that right now the 'experiences' of the dataset creation and the 'experience' in the Strategy flow are entangled. We should still be able to create a datastream in a set of steps (or experiences, such as e.g. typical Split-MNIST), while the algorithm does batch-wise processing (if the batch would be an experience, this becomes a problem). It should also be possible to process the same batch multiple times as is often done in the online data incremental scenarios.

vlomonaco:
I think we already support (2) it's just that it is probably not that efficient, but I'll let @lrzpellegrini brief us on this.

AntonioCarta:
> The problem is though that right now the 'experiences' of the dataset creation and the 'experience' in the Strategy flow are entangled. 

Sorry, what do you mean?

> It should also be possible to process the same batch multiple times as is often done in the online data incremental scenarios.

This is already possible, and it works also in the data incremental scenario with option (2).

lrzpellegrini:
Having an experience for each mini-batch (2) is totally doable. It shouldn't add too much overhead.

A custom scenario can already generate the required experiences if correctly implemented, even experiences containing 1 single pattern. Also, experiences are now created on the fly in a lazy way and the underlying dataset is generated based on lists of indexes, so this is why expect little to no overhead.

[...]

PR #553, opened 2021-04-22
Replay with StoragePolicies and CoPE strategy

Mattdl:
This PR implements the following:

Replay:
- ReplayPlugin now uses an abstract StoragePolicy to determine which sample store in the replay memory.
- 2 implementations of the StoragePolicy are provided: ClassBalancedStoragePolicy & ExperienceBalancedStoragePolicy.
- Test for the different storage policies. 
- The implementation of ExperienceBalancedStoragePolicy solves the bug in #543 
-  Replay memory is fully filled with the available data: remaining data after dividing equally over experiences/classes, is divided at random taking into account the data available per experience/class.

CoPE:
- The CoPE plugin for online data incremental learning is implemented.
- The CoPE strategy wrapper.
- Testing for the CoPE strategy.
- A working example of the CoPEPlugin. This is a class-incremental setup example. The online data incremental setup still is to be implemented as discussed in #543, #360.

SimpleMLP
- Small update where the number of layers can be easily configured as well.


Mattdl:
The PEP8 error didn't show locally. Also don't know how `l` is ambiguous in any way in the SimpleMLP.
Any ideas why this happens?

lrzpellegrini:
Alas, it seems that PEP8 doesn't really like variables, classes, etc. named ‘l’, ‘O’, or ‘I’ no matter the context.

AntonioCarta:
Thanks @Mattdl, everything looks good. Can I merge it?

Mattdl:
Thanks! I've no further updates for this PR, so you can merge (:

PR #554, opened 2021-04-22
Fix wrong normalization statistics for CIFAR-100

imirzadeh:
CIFAR-100 has different normalization stats than CIFAR-10. However, in the current implementation, the same stats were used for CIFAR-10 and CIFAR-100.
This commit fixes this issue.

Source for calculating stats: https://gist.github.com/weiaicunzai/e623931921efefd4c331622c344d8151


vlomonaco:
Thanks @imirzadeh! :) merging now!

PR #555, opened 2021-04-22
Fix wrong normalization statistics for Fashion MNIST

imirzadeh:
Fashion MNIST normalization stats are wrong in the current implementation.  

Source for calculating stats: https://gist.github.com/weiaicunzai/e623931921efefd4c331622c344d8151

### DIFF
--- a/avalanche/benchmarks/classic/cfashion_mnist.py
+++ b/avalanche/benchmarks/classic/cfashion_mnist.py
@@ -22,12 +22,12 @@ from avalanche.benchmarks.utils import train_eval_avalanche_datasets
 
 _default_fmnist_train_transform = transforms.Compose([
     transforms.ToTensor(),
-    transforms.Normalize((0.2190,), (0.3318,))
+    transforms.Normalize((0.2860,), (0.3530,))
 ])
 
 _default_fmnist_eval_transform = transforms.Compose([
     transforms.ToTensor(),
-    transforms.Normalize((0.2190,), (0.3318,))
+    transforms.Normalize((0.2860,), (0.3530,))
 ])
 
 


vlomonaco:
Hi @imirzadeh! Thanks for the heads-up! :) The link refers to the cifar dataset. What's your source for the proposed change? 
@ggraffieti do you remember where you took the original numbers here?

AndreaCossu:
I changed the Fashion MNIST transformations in #529 following @lrzpellegrini advice. 
Values are taken from https://github.com/facebookresearch/Adversarial-Continual-Learning/blob/master/src/dataloaders/mulitidatasets.py

lrzpellegrini:
@AndreaCossu, I confirm that the values computed by @imirzadeh are correct. I don't know why https://github.com/facebookresearch/Adversarial-Continual-Learning/blob/master/src/dataloaders/mulitidatasets.py uses such values.

See the example i prepared: https://colab.research.google.com/drive/1EQIpTDuXAbm4F6TYgzdo2Vkzm9u1NmL4?usp=sharing

lrzpellegrini:
I'll merge this PR. Thanks @imirzadeh for the fix!

imirzadeh:
> Hi @imirzadeh! Thanks for the heads-up! :) The link refers to the cifar dataset. What's your source for the proposed change?

Hi @vlomonaco ! In the comments below the link the values for other datasets were calculated as well.

imirzadeh:
> @AndreaCossu, I confirm that the values computed by @imirzadeh are correct. I don't know why https://github.com/facebookresearch/Adversarial-Continual-Learning/blob/master/src/dataloaders/mulitidatasets.py uses such values.
> 
> See the example i prepared: https://colab.research.google.com/drive/1EQIpTDuXAbm4F6TYgzdo2Vkzm9u1NmL4?usp=sharing

I guess the normalization stats do not matter much in the final performance. Even without any specific normalization, the torchvision's `ToTensor()` transform will automatically scale down the image values from [0, 255] to [0, 1], which should be enough alone.

I added this patch just to be precise.

PR #556, opened 2021-04-23
pytorchcv wrapper

### LINKED ISSUE #430: Add Models Available in PyTorchCV
We should add direct import for pytorchcv (pre-trained) models through the models module.

VerwimpEli:
Hi,

Closes #430 . This PR implement a pytorchcv wrapper to load various models. A general wrapper is available through `get_model`. For VGG, Resnet, DenseNet and Pyramidnet direct wrappers are provided.

A bit of explanation for the design choices:
- Initially I wanted to make sure that imagenet pre-trained models could be easily used for Cifar etc. However, this isn't as simple as changing input and output layers. Eg in Resnet, the structure of the convolutional layers also change because of the smaller input size. Since there is no straight-forward way, I would leave it up to the user to change models if required, such that no unexpected behavior is introduced.
- The choice of the networks (vgg, densenet, pryamidnet..)  is mostly based on the nets available for cifar10/100. 
- Not all options (eg growth rate for densenet, alpha in pyramidnet, bottlenecks...) are available through the direct wrappers. I don't think this is necessary, as those wrappers should be convenient and quick. If someone wants a very specific model, it's probably quicker to use the general wrapper anyway.
- As mentioned in the issue #430 discussion,  a slimmed down Resnet18 is a often used net, but not available in `pytorchcv`. In the future I would either expand these functions to also load our own implemented networks, which includes a slimmed down resent.

### DIFF
--- a/avalanche/models/pytorchcv_wrapper.py
+++ b/avalanche/models/pytorchcv_wrapper.py
@@ -1,0 +1,99 @@
+################################################################################
+# Copyright (c) 2021 ContinualAI.                                              #
+# Copyrights licensed under the MIT License.                                   #
+# See the accompanying LICENSE file for terms.                                 #
+#                                                                              #
+# Date: 1-05-2020                                                              #
+# Author(s): Eli Verwimp                                       #
+# E-mail: contact@continualai.org                                              #
+# Website: www.continualai.org                                                 #
+################################################################################
+
+"""
+This module wraps pytorchcv models, with some convient wrappers.
+"""
+
+from pytorchcv.model_provider import get_model as ptcv_get_model
+from torch.nn import Module
+
+
+def vgg(depth: int, batch_normalization=False, pretrained=False) -> Module:
+    """
+    Wrapper for VGG net of verious depths availble in pytorchcv. Only availabe for imagenet.
+    :param depth: Depth of the model, one of (11, 13, 16, 19)
+    :param batch_normalization: include batch normalizaion layers
+    :param pretrained: loads model pretrained on imagnet
+    """
+    available_depths = [11, 13, 16, 19]
+    if depth not in available_depths:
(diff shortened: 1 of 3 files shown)


avalanche/models/pytorchcv_wrapper.py
```diff
@@ -0,0 +1,145 @@
+################################################################################
+# Copyright (c) 2021 ContinualAI.                                              #
+# Copyrights licensed under the MIT License.                                   #
+# See the accompanying LICENSE file for terms.                                 #
+#                                                                              #
+# Date: 1-05-2020                                                              #
+# Author(s): Eli Verwimp                                       #
```
vlomonaco, line 7: The "#" is not aligned with the others :)

avalanche/models/pytorchcv_wrapper.py
```diff
@@ -0,0 +1,145 @@
...
+# E-mail: contact@continualai.org                                              #
+# Website: www.continualai.org                                                 #
+################################################################################
+
+"""
+This module wraps pytorchcv models, with some convient wrappers.
```
vlomonaco, line 13: Can you expand this a little with the ideas that you put already in the description of the PR? This may be useful to understand the scope of the module better!

vlomonaco (changes requested):
Thanks @VerwimpEli for the PR, it looks great! :)
Before merging I'd ask you to:

- Add an example in the `examples` directory on the usage of this methods
- Add a unittest in the `tests` directory to check if everything works as expected 

Thanks!

avalanche/models/pytorchcv_wrapper.py
VerwimpEli: Alright, updated now!

vlomonaco (approved):

vlomonaco:
Thanks again @VerwimpEli! Merging now!

PR #562, opened 2021-04-23
Avalanche Subset/ConcatDataset flattening

### LINKED ISSUE #560: Bad recursion management in AvalancheSubset/ConcatDatasets
`AvalancheSubset` and `AvalancheConcatDataset` currently manage subsetting and concatenation in a way which is very similar to PyTorch `Subset` and `ConcatDataset`.

This however may lead to recursion errors in some corner cases (see #543).

In order to fix this, I'm planning on flattening the tree when creating instances of those classes.

### LINKED ISSUE #565: Wrapping AvalancheDataset(s) doesn't inherit the initial transformation group
Initializing an `AvalancheDataset` by passing another `AvalancheDataset` leads to the unexpected behavior of not inheriting the initial transformations group as 'train' is always used as the initial transformation group.

To reproduce the error:

```python
original_dataset = MNIST(
    root=expanduser("~") + "/.avalanche/data/mnist/", download=True
)

def plus_one_target(target):
    return target+1

transform_groups = dict(
    train=(ToTensor(), None),
    eval=(None, plus_one_target)
)

x, y = original_dataset[0]
dataset = AvalancheDataset(original_dataset, transform_groups=transform_groups)

dataset_eval = dataset.eval()
dataset_inherit = AvalancheDataset(dataset_eval)

x2, y2, _ = dataset_inherit[0]
self.assertIsInstance(x2, PIL.Image.Image)  # Fails
self.assertIsInstance(y2, int)
self.assertEqual(y + 1, y2)
```

It should be easy to implement an hotfix by checking for the existing active group of the input dataset.

### LINKED ISSUE #566: AvalancheSubset break transformation groups
Wrapping an `AvalancheDataset` in an `AvalancheSubset` breaks transformation groups mechanics. When switching transformation group in a subset, the underlying dataset's "current group" is not changed to the required group.

Code to reproduce the error:
```python
original_dataset = MNIST(
    root=expanduser("~") + "/.avalanche/data/mnist/", download=True
)

def plus_one_target(target):
    return target+1

transform_groups = dict(
    train=(ToTensor(), None),
    eval=(None, plus_one_target)
)

x, y = original_dataset[0]
dataset = AvalancheDataset(original_dataset,
                           transform_groups=transform_groups)

dataset_sub_train = AvalancheSubset(dataset)
# Here .eval() doesn't actually change the internal fork of 'dataset' to
# use the 'eval' group -> train transforms are still used
dataset_sub_eval = dataset_sub_train.eval()  
dataset_sub = AvalancheSubset(dataset_sub_eval, indices=[0])

x5, y5, _ = dataset_sub[0]
self.assertIsInstance(x5, PIL.Image.Image)  # Fails
self.assertIsInstance(y5, int)
self.assertEqual(y + 1, y5)
```

Fix on its way!

lrzpellegrini:
This PR implements the flattening of the hierarchy of datasets created using `AvalancheSubset` or `AvalancheConcatDataset`.

This will improve instance retrieval time and prevent max recursion errors (see #560)

Update: in the process, I also detected and fixed #565 and #566.

Fixes #560 
Fixes #565 
Fixes #566

### DIFF
--- a/avalanche/benchmarks/classic/ccifar100.py
+++ b/avalanche/benchmarks/classic/ccifar100.py
@@ -24,14 +24,14 @@ _default_cifar100_train_transform = transforms.Compose([
     transforms.RandomCrop(32, padding=4),
     transforms.RandomHorizontalFlip(),
     transforms.ToTensor(),
-    transforms.Normalize((0.4914, 0.4822, 0.4465),
-                         (0.2023, 0.1994, 0.2010))
+    transforms.Normalize((0.5071, 0.4865, 0.4409),
+                         (0.2673, 0.2564, 0.2762))
 ])
 
 _default_cifar100_eval_transform = transforms.Compose([
     transforms.ToTensor(),
-    transforms.Normalize((0.4914, 0.4822, 0.4465),
-                         (0.2023, 0.1994, 0.2010))
+    transforms.Normalize((0.5071, 0.4865, 0.4409),
+                         (0.2673, 0.2564, 0.2762))
 ])
 
 
--- a/avalanche/benchmarks/classic/cfashion_mnist.py
+++ b/avalanche/benchmarks/classic/cfashion_mnist.py
@@ -22,12 +22,12 @@ from avalanche.benchmarks.utils import train_eval_avalanche_datasets
 
 _default_fmnist_train_transform = transforms.Compose([
     transforms.ToTensor(),
-    transforms.Normalize((0.2190,), (0.3318,))
+    transforms.Normalize((0.2860,), (0.3530,))
 ])
 
 _default_fmnist_eval_transform = transforms.Compose([
     transforms.ToTensor(),
-    transforms.Normalize((0.2190,), (0.3318,))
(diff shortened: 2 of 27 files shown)


vlomonaco:
Perfect, thanks!

lrzpellegrini:
This is not ready yet as I still need to tackle the Subset part (which is a little bit more tricky).

AntonioCarta:
@lrzpellegrini is this really possible to implement easily? What happens if the concatenated datasets have different transformations? If I'm not mistaken, with the current implementation the two datasets keep the different transformations (and it's possible to compose additional transformations). However, this is one of the reasons why (I guess) it's hard to flatten the concatenated datasets.

lrzpellegrini:
Hi @AntonioCarta, in the last hours I worked specifically on that point. It now also works for the subset (I added a ton of unit tests ;) )

lrzpellegrini:
This will also include an hotfix for #565, please be patient!

PR #567, opened 2021-04-27
BaseStrategy support for arbitrary mini-batches

### LINKED ISSUE #539: Update dataloader to return a single minibatch
Dataloaders now split data by task labels. Now that models support sample-level task labels this is not necessary anymore. Dataloader should return a single mini-batch, possibly with different task labels.

AntonioCarta:
This PR updates dataloaders. Now BaseStrategy supports pytorch dataloaders. Custom dataloaders are provided for CL-specific features:
- TaskBalancedDataLoader
- GroupBalancedDataLoader
- ReplayDataLoader

BaseStrategy (and dataloaders) now supports arbitrary mini-batches of the form `<x, y, ..., t>`.

Closes #539.

### DIFF
--- a/avalanche/benchmarks/utils/data_loader.py
+++ b/avalanche/benchmarks/utils/data_loader.py
@@ -8,7 +8,12 @@
 # E-mail: contact@continualai.org                                              #
 # Website: avalanche.continualai.org                                           #
 ################################################################################
-from copy import copy
+"""
+    Avalanche supports data loading using pytorch's dataloaders.
+    This module provides custom dataloaders for continual learning such as
+    support for balanced dataloading between different tasks or balancing
+    between the current data and the replay memory.
+"""
 from itertools import chain
 from typing import Dict, Sequence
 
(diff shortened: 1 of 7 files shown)


vlomonaco:
That's great @AntonioCarta! I think a brief docstring explaining the rationale behind the three main CL data loaders we provide would be a great add to the `data_loader.py` module incipit. Also, it would be great to add an example showcasing the difference/features offered from these loaders. We can later on create an How-to guide related to data loading based on this.

Also, where do you think we should add some hints about data loading in the FZTH tutorial? in the Benchmark or Training chapter?

AntonioCarta:
@vlomonaco yes, I will add the doc and example.

> Also, where do you think we should add some hints about data loading in the FZTH tutorial? in the Benchmark or Training chapter?

Like in pytorch, I think dataloaders should be part of the benchmarks. However, I believe they are an advanced topic that should not be explained in the FZTH. Most users don't need to care about dataloaders because they are already managed by the strategies.

PR #569, opened 2021-04-28
Docker image with pre-downloaded datasets used in test

### LINKED ISSUE #454: Use docker image that contains dataset in github actions
Github actions continue to fail to download MNIST dataset. 
Create and use a Docker container that already contains all the needed data, so we don't have to download it every time.

ggraffieti:
This PR improves the automatic testing of the project: 
- Using a docker image with pre-downloaded datasets, data no longer has to be downloaded every time the software is tested. This solves download issues (broken links, servers down, slow connection..) making the testing procedure more reliable. 
- Using a docker image with only python + conda installed increase the validity of testing since the environment is minimal and does not contain any unnecessary package. 
- Conda is now runned directly in the shell without relying on any additional action. This seems to have solved the slow environment creation with python 3.6 
- Fix download location for the data used during tests.

Now in the docker image are only contained the MNIST and tiny imagenet, cifar10 and cats_and_dogs will be inserted in the docker image as soon as possible (in any case the insertion of the datasets only requires the upload of a new container on DockerHub, so it's totally independent from Avalance). 

This PR resolves #454

### DIFF
--- a/.github/workflows/test-coverage-coveralls.yml
+++ b/.github/workflows/test-coverage-coveralls.yml
@@ -72,7 +72,9 @@ jobs:
           conda activate avalanche-env
           FAST_TEST=True USE_GPU=False coverage run -m unittest
       - name: Upload coverage data to coveralls.io
-        run: coveralls --service=github
+        run: |
+          conda activate avalanche-env
+          coveralls --service=github
         env:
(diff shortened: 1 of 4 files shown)


vlomonaco:
Perfect, thanks!

PR #583, opened 2021-05-03
Added W&B built-in Confusion Matrix and Deep SLDA

### LINKED ISSUE #564: Confusion Matrix for W&B
We should try to see how much code changes is actually needed to implement the builtin confusion matrix for W&B Logger.

Some points to note:

1. Metrics are meant to be disconnected from the loggers, so it is not ideal to have two different implementation of the CM. It is also true that the current one provides the CM directly, while the other would provide information on logits, true labels etc which combined can be used to produce the CM. Alas, there is no way around this if we want to support the builtin CM.
2. We could create additional CM metrics for W&B and expose only the usual utility function to build the metrics with the param `wandb=True/False`. 
3. We have to assess the impact on the existing loggers

I'll make a draft PR to sketch an implementation and we will see how much this violates the Avalanche design.

AndreaCossu:
This PR brings the W&B built-in confusion matrix.
The ConfusionMatrix can be instantiated by using the `confusion_matrix_metrics` function, as for the other metrics. The metric gathers logits and targets and build the matrix at the end of the evaluation stream.

@digantamisra98 I am able to log the matrix but it is not possible to see the results from different streams, I'll need your help to clarify this point.
Also, W&B automatically produces a Table, not only the confusion matrix visualization. Is this an expected outcome? Tables are correctly visualized for different streams.

This closes #564


AndreaCossu:
Current results for Deep SLDA example (average accuracy over all the test stream at the end of training on experience i=0,...,8):
0.1895 0.2663 0.3371 0.3974 0.4496 0.5088 0.5689 0.6420 0.6845

AntonioCarta:
Thanks @AndreaCossu! I have a couple comments:
- We need a test for the strategy.
- `ModelWrapper` is a generic component. Maybe we could put it inside the `models` module since there are other strategies that need it (AR1 for example does something similar).
- Can you add a link to the paper in the docstring?
- The results seems lower than what Tyler expected.

@tyler-hayes do you have any comments? Can you explain the difference in performance? We are probably using different hyperparameters somewhere.

AndreaCossu:
I agree on all your points :smile: I already contacted Tyler to get a feedback on this. I'll add the `ModelWrapper` in the `utils` file under `models`. I think we could also rename it to better clarify its purpose, but I don't have a good alternative, yet.

AndreaCossu:
After a thorough check with @tyler-hayes we came to the conclusion that the only difference between the two implementations is the way the use data. Here, we rely on the default Avalanche data loading while in the original PR (#500) the example loop over the experience dataset by wrapping a dataloader around it and computing metrics by hand. 
Unfortunately, I am not able to say what could be the problem here. The 10% reduction in accuracy we see could be further investigated in the reproducibility project backed by Avalanche. Maybe a model selection could bridge the gap with the other results.

Therefore, I think we can merge this PR (once the W&B commits are ready) since the strategy implementation should be correct.

avalanche/training/strategies/deep_slda.py
```diff
@@ -0,0 +1,287 @@
...
+import torchvision.models as models
+
+from avalanche.training.plugins import StrategyPlugin
+from avalanche.training.strategies import BaseStrategy
+from avalanche.training import default_logger
+from avalanche.models.dynamic_modules import MultiTaskModule
+from avalanche.models import ModelWrapper
+
+
+class SLDAResNetModel(nn.Module):
```
vlomonaco, line 16: I'd move this class to the models package...

avalanche/training/strategies/deep_slda.py
```diff
@@ -0,0 +1,287 @@
...
+
+class StreamingLDA(BaseStrategy):
+    """
+    Deep Streaming Linear Discriminant Analysis.
+    This strategy does not use backpropagation.
+    Minibatches are first passed to the pretrained feature extractor.
+    The result is processed one element at a time to fit the
+    LDA.
+    Original paper:
+    https://openaccess.thecvf.com/content_CVPRW_2020/papers/w15/Hayes_Lifelong_Machine_Learning_With_Deep_Streaming_Linear_Discriminant_Analysis_CVPRW_2020_paper.pdf
```
vlomonaco, line 77: I'd add the textual reference as we do in a paper

avalanche/training/strategies/deep_slda.py
AndreaCossu: If we believe this to be a model useful also in other cases I can move it there. It currently uses `pool_feat` which is something a bit too specific in my opinion. Its purpose is to reproduce SLDA results, not to be useful at a general level. 
It really depends on the purpose of `models` module, I think.

vlomonaco:
@AndreaCossu and @tyler-hayes I think it would be important to investigate a little more why this degradation is happening since it may be related about bugs in the data loading or the evaluation. Can you check if you can obtain the same results with the avl data loader but the custom accuracy computation?

avalanche/training/strategies/deep_slda.py
vlomonaco: In "models" we have already specific models used by other strategies... maybe we can have a specific sub-package there for this kind of models. But I'd prefer to maintain all the models in a single place.

AndreaCossu:
I already did this by adding the manual accuracy computation in the base strategy and I obtained the same results produced by the loggers.

AndreaCossu:
Also @tyler-hayes was proposing `FeatureExtractorBackbone` instead of `ModelWrapper`. It makes sense to me, what do you think?

vlomonaco:
ok great, can you check the impact of the data ordering on the final results? (just by re-creating the benchmark with a different seed?)
If this does not change much... then there's probably a bug somewhere in the data loading / dataset. Is the preprocessing done in the same way?

vlomonaco:
> Also @tyler-hayes was proposing `FeatureExtractorBackbone` instead of `ModelWrapper`. It makes sense to me, what do you think?

Yes, much better!

AndreaCossu:
The preprocessing is the same for both examples (Z-normalization only). As for the seed, `CORe50` does not have a `seed` parameter like `SplitMNIST`. Is there a way around this?

vlomonaco:
> The preprocessing is the same for both examples (Z-normalization only). As for the seed, `CORe50` does not have a `seed` parameter like `SplitMNIST`. Is there a way around this?

You can use the "run" parameter instead. Also, check if shuffling is enabled in the data loader.

AndreaCossu:
For the data loader I tried with both `shuffle=True` and `shuffle=False` by changing it in the base strategy. We both use `run=0` but I'll try to change it.

AntonioCarta:
@AndreaCossu can you show me the example that is giving you the correct results?

tyler-hayes:
> ok great, can you check the impact of the data ordering on the final results? (just by re-creating the benchmark with a different seed?)
> If this does not change much... then there's probably a bug somewhere in the data loading / dataset. Is the preprocessing done in the same way?

@AndreaCossu @vlomonaco I re-ran the Deep SLDA implementation from my original PR using `run=1` (instead of `run=0`) for the NC ordering of CORe50 and obtained the following:
```
[0.19227519345370453, 0.2780841412434404, 0.3584230187672329, 0.4274215067152895, 0.5103620030241038, 0.5772480654629547, 0.6617006137152006, 0.7352352574935516, 0.7970737347683002]
```

The final performance should match my original run identically, but there are some minor differences for other batches (due to the different ordering).

I believe we are both processing data in the same way. I am using the following:
```
_mu = [0.485, 0.456, 0.406]  # imagenet normalization
_std = [0.229, 0.224, 0.225]
transform = transforms.Compose([
  transforms.Resize((224, 224)),
  transforms.ToTensor(),
  transforms.Normalize(mean=_mu,
                       std=_std)
])
```

tyler-hayes:
> @AndreaCossu can you show me the example that is giving you the correct results?

@AntonioCarta I sent over the two files I have been working with on the Continual AI Slack. If it is helpful for me to put them here, I could do that as well.

avalanche/training/strategies/deep_slda.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # setup weights for SLDA
+        self.muK = torch.zeros((num_classes, input_size)).to(self.device)
+        self.cK = torch.zeros(num_classes).to(self.device)
+        self.Sigma = torch.ones((input_size, input_size)).to(self.device)
+        self.num_updates = 0
+        self.Lambda = torch.zeros_like(self.Sigma).to(self.device)
+        self.prev_num_updates = -1
+
+    def forward(self, return_features=False):
```
AntonioCarta, line 81: The `forward` method should be:
```
    def forward(self, return_features=False):
        self.model.eval()
        if isinstance(self.model, MultiTaskModule):
            feat = self.model(self.mb_x, self.mb_task_id)
        else:  # no task labels
            feat = self.model(self.mb_x)
        out = self.predict(feat)
        if return_features:
            return out, feat
        else:
            return out
```

AntonioCarta (commented):
Good news, I was able to reproduce the results by fixing the forward method.

You can check the results using the same dataloaders with a custom strategy:
```
    class MySLDA(StreamingLDA):
        def make_train_dataloader(self, num_workers=0, shuffle=True,
                                  pin_memory=True, **kwargs):
            self.dataloader = DataLoader(self.experience.dataset,
                                         batch_size=args.batch_size,
                                         shuffle=False, num_workers=8)

        def make_eval_dataloader(self, num_workers=0, pin_memory=True,
                                 **kwargs):
            self.dataloader = DataLoader(scenario.test_stream[0].dataset,
                                         batch_size=args.batch_size,
                                         shuffle=False, num_workers=8)
```

@AndreaCossu as soon as you fix the forward I will merge it ;)

AntonioCarta:
BTW, is there any particular reason why `train_epochs` is an argument? Shouldn't it be fixed to 1?

AndreaCossu:
Thanks @AntonioCarta ! I'll take a look at it and implement the changes! 
For the number of epochs, even if it is true that the strategy has been proposed as a streaming learning algorithm, I think that in principle it might also be used with multiple epochs if the user wants. It is like GEM and A-GEM, which have been proposed in online learning settings but you can also use larger minibatch sizes.

digantamisra98 (approved):
Looks good to me. I think the table creation as an artifact when logging custom charts is not expected but I will look into it in more detail today. Additionally I also think that custom chart do not offer iteration stream for viewing like normal panels do but just offer the latest snapshot but this can be changed I believe. 

As of now this can be merged.

AndreaCossu:
Ok, since this has been opened for a while I'll merge it and we can check out more in-depth the confusion matrix later.

digantamisra98:
Sure yes, I will work on this through this week.

PR #591, opened 2021-05-04
Implementation of data incremental benchmarks (helper)

### LINKED ISSUE #360: Support for data incremental scenario

lrzpellegrini:
This PR adds the support of data incremental benchmarks.

The new helper `data_incremental_benchmark` accepts a benchmark instance and returns another instance in which experiences are split. By default train experiences will be split according to a fixed size. This behavior can be changed by passing a custom splitting strategy. Experiences from other streams can be split as well.

Fixes #360


vlomonaco:
Up to now we had only benchmarks generators starting from one or multiple datasets. If we want to keep them all of them in the same sub-package I think we need to update the doc accordingly and generalize a bit the concept of generators at least as we used it until now. What do you think?

AntonioCarta:
I think it's ok to keep them together. These are heterogeneous helpers and over time we will probably need a lot of them that don't fit a predefined structure. Having a subpackage for each of them seems more dispersive and harder to search.

Maybe we should ask the feedback of some users that asked for this feature before merging. @Mattdl @tyler-hayes what do you think? Do you need anything else?

Mattdl:
First of all, this is an amazing PR. Good job @lrzpellegrini !
I think this is a really elegant way to implement the data incremental setup.
Because 1 batch is now a single experience, it's also possible to do multiple 'epochs' on this batch, as is common in online CL literature.
This really shows the strength of the modularity in Avalanche 😄 👍🏼 

I see though that the splitting is not lazy, which will only give problems for long sequences (e.g. 20 tasks of rotated MNIST). 
This is rather an edge case that can probably be solved later on in another issue.

lrzpellegrini:
Thank you @Mattdl!

The problem with the lazy splitting is that currently streams have a size (expose a `__len__` method) so we need to know how many experiences it contains. In the future we'll support lazy streams as well.

PR #598, opened 2021-05-06
Fix for #593

### LINKED ISSUE #593: Bug in classes_in_this_experience on test set
The `classes_in_this_experience` does not work when applied to the complete test set of CoRE50. The test stream correctly contains all the 50 classes but the parameter returns only the first 10. 

I think the problem may be related to the fact that `LazyClassesInExps` only works with training experiences (which impacts on the experience factory generating the test stream). @lrzpellegrini can you provide help on this?
```
scenario = CORe50(scenario='nc')
print(len(scenario.test_stream))
print(scenario.test_stream[0].classes_in_this_experience)
```
Output:
```
»»» 1
»»» [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```
Expected:
```
»»» 1
»»» [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, ..., 49]
```

lrzpellegrini:
The `classes_in_this_experience` field of `GenericCLScenario` by default returned the list linked to the training stream. This made it difficult to obtain the list of classes for non-training experiences.

This PR introduces another way to use that field, which can now be used as:

```python
benchmark_instance.classes_in_experience[stream_name][exp_id]
```

The previous way to use this field is deprecated. The following:

```python
benchmark_instance.classes_in_experience[exp_id]  # Deprecated
```

still works, but a warning will be printed to stdout.

Fixes #593

### DIFF
--- a/avalanche/benchmarks/scenarios/generic_cl_scenario.py
+++ b/avalanche/benchmarks/scenarios/generic_cl_scenario.py
@@ -2,8 +2,10 @@ import copy
 import re
 from abc import ABC
 from typing import Generic, TypeVar, Union, Sequence, Callable, Optional, \
-    Dict, Any, Iterable, List, Set, Iterator, Tuple, NamedTuple
+    Dict, Any, Iterable, List, Set, Iterator, Tuple, NamedTuple, Mapping, \
+    overload
 
+import warnings
 from torch.utils.data.dataset import Dataset
 
 from avalanche.benchmarks.scenarios.generic_definitions import \
@@ -176,6 +178,15 @@ class GenericCLScenario(Generic[TExperience]):
         self._make_stream_fields()
 
     @property
+    def streams(self) -> Dict[str, 'GenericScenarioStream['
+                                   'TExperience, TGenericCLScenario]']:
+        streams_dict = dict()
+        for stream_name in self.stream_definitions.keys():
+            streams_dict[stream_name] = getattr(self, f'{stream_name}_stream')
+
+        return streams_dict
+
+    @property
     def n_experiences(self) -> int:
         """  The number of incremental training experiences contained
         in the train stream. """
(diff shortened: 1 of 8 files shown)


vlomonaco:
You mean 

`benchmark_instance.classes_in_experience[stream_name][exp_id]`

not:

`benchmark_instance.classes_in_this_experience[stream_name][exp_id]`

right?

lrzpellegrini:
Yes, sorry it's `classes_in_experience`!

lrzpellegrini:
There also seem to be problems with the experience level field. I'll fix it ASAP.

AntonioCarta:
After this change it seems that the attribute belongs to the experience. Maybe it shouldn't be in the `GenericCLScenario` at all? it seems weird to use the dictionary instead of `benchmark_instance.stream_name.classes_in_this_experience[exp_id]`.

lrzpellegrini:
Users can currently access the list of classes through

```python
benchmark_instance.classes_in_experience[stream][exp]
```

or 

```python
benchmark_instance.xxx_stream[exp].classes_in_this_experience
```

The result is the same. For the moment I'd go with this plain fix. In the future we may consider re-organizing this mechanism.

vlomonaco:
It's ok for me. What's really important IMHO is that `experience.classes_in_this_experience` works as expected!

AndreaCossu:
Thanks @lrzpellegrini, ok for me!

AntonioCarta:
> 
> 
> Users can currently access the list of classes through
> 
> ```python
> benchmark_instance.classes_in_experience[stream][exp]
> ```
> 
> or
> 
> ```python
> benchmark_instance.xxx_stream[exp].classes_in_this_experience
> ```
> 
> The result is the same. For the moment I'd go with this plain fix. In the future we may consider re-organizing this mechanism.

Right, I forgot you could already do it. I would probably remove `benchmark_instance.classes_in_experience` since it's redundant, but it's fine if you prefer it this way.

PR #599, opened 2021-05-06
Googledrivedownloaded replaced by Gdown

### LINKED ISSUE #533: Update/Remove Current Auto-Download Features from Google Drive
Currently we have `googledrivedownloader` but we don't use it since downloads from Drive are usually corrupted or interrupted for big files. We should remove it if we don't use it anymore.

In this repo they use another method to download gdrive files: https://github.com/salmank255/IJCAI-2021-Continual-Activity-Recognition-Challenge/blob/main/Data/download_data.sh

We should check if this is something we can integrate into AVL.

vlomonaco:
This PR replaces `googledrivedownloader` with `gdown` as the main dep to download data from google drive. 

1. Now download works even with large documents hosted on GDrive.
2. OpenLoris and CUB200 have been updated accordingly.

This PR closes #533.


vlomonaco:
Even if this fixes the gdrive download problem google still limits automatic download so I cannot test this properly:

```
Too many users have viewed or downloaded this file recently. Please
        try accessing the file again later. If the file you are trying to
        access is particularly large or is shared with many people, it may
        take up to 24 hours to be able to view or download the file. If you
        still can't access a file after 24 hours, contact your domain
        administrator.
```

Google drive should be used as the last possible resort.

PR #601, opened 2021-05-06
Validation stream creation helper

### LINKED ISSUE #262: Split Streams "horizontally": Create validation stream from train stream
We should provide a simple way to split the training data into train and validation splits.

lrzpellegrini:
This PR introduces and helper that can be used to obtain a benchmark instance with a proper validation stream.

This simple helper `benchmark_with_validation_stream` takes an existing benchmark instance as an input and splits each experience in the train stream to obtain a valid stream with the same amount of experiences.

As of now no class balancing is used (as the helper must be general enough to handle non-classification tasks), but a custom splitting strategy can be employed as well.

The size of validation experiences can be set by passing an absolute size (int) or a relative size (float between 0 and 1).

Fixes #262

### DIFF
--- a/avalanche/benchmarks/generators/benchmark_generators.py
+++ b/avalanche/benchmarks/generators/benchmark_generators.py
@@ -16,7 +16,8 @@ the generic ones: filelist_benchmark, tensors_benchmark, dataset_benchmark
 and paths_benchmark.
 """
 from functools import partial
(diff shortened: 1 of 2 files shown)


AndreaCossu:
You really are :smile:

AntonioCarta:
Great!

PR #602, opened 2021-05-07
Added csvlogger

AndreaCossu:
This PR brings the `CSVLogger`: a logger which can be used to produce two files.
The training file lists for each epoch and experience the training and validation accuracies and losses.
The evaluation file lists for each training and evaluation experience the accuracy, loss and forgetting.

Important points which may reduce the value of this logger: 
this is quite a specific logger which assumes 
1) logging at least accuracy explicitly for each epoch and experience (maybe the logger name could be changed into `CSVClassificationLogger` but I don't like it very much). 
2) evaluation during training is performed only on the eval experience corresponding to the training one.
The example `eval_plugin` shows how to use it and I could provide a separate one if needed. We could also think about making it more general so that the metrics are not pre-defined in the logger.
I would like to know your opinion on this: do you believe this to be something useful?

@AntonioCarta I also proposed a rough fix so that the periodic eval is not run twice if `eval_every` is a multiple of `epochs`. Let me know if you have better solutions to this problem.

### DIFF
--- a/avalanche/logging/__init__.py
+++ b/avalanche/logging/__init__.py
@@ -16,5 +16,4 @@ from .tensorboard_logger import *
 from .wandb_logger import *
 from .text_logging import TextLogger
 from .interactive_logging import InteractiveLogger
-
-from avalanche.logging import InteractiveLogger
+from .csv_logger import CSVLogger
--- a/avalanche/logging/csv_logger.py
+++ b/avalanche/logging/csv_logger.py
@@ -1,0 +1,176 @@
+################################################################################
+# Copyright (c) 2021 ContinualAI.                                              #
+# Copyrights licensed under the MIT License.                                   #
+# See the accompanying LICENSE file for terms.                                 #
+#                                                                              #
+# Date: 2020-01-25                                                             #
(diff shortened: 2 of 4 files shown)


vlomonaco:
good job @AndreaCossu!

But I have a couple of questions: 

1. Do we really need csv files that are lively populated? 
2. What's the end usage of these csv files?
3. Why can't we just create utilities that given the metrics results dict at the end of the training converts everything to csv files?

AndreaCossu:
1. Yes and no, I prefer to have them updated interactively because for long runs I'd like to see what's going on before waiting until the end. It's true that the other loggers can be used for this purpose but in this way I got a simpler view. It's just a matter of preference I think
2. I use these files to check some specific metrics I am particularly interested in during experiments (I log more metrics because it may be useful but I am not always interested in all of them) and to produce plots basically.
3. From the dictionary it is not easy to find the values of evaluation during training for example. This is because currently they are mixed up with the metrics resulting from direct `strategy.eval` calls (evaluation during training and evaluation outside training calls the same `eval` function).

PR #605, opened 2021-05-11
Fix bug in GDumbPlugin

lebrice:
Hello there, thanks for this cool package! :)

I think I've spotted a bug in GDumbPlugin, it fails to concatenate the samples due to a shape mismatch between `pattern` and `target`.  

Also, I think the current implementation is quite inefficient: for each new item,
it performed an entire concatenation with the current dataset.
This uses lists instead, and only concatenates once.

It also uses the task labels from each sample in the dataset, rather
the task label from the current experience, as there might be more
than one task in the dataset.

Signed-off-by: Fabrice Normandin <fabrice.normandin@gmail.com>

### DIFF
--- a/avalanche/training/plugins/gdumb.py
+++ b/avalanche/training/plugins/gdumb.py
@@ -41,20 +41,24 @@ class GDumbPlugin(StrategyPlugin):
         # for each pattern, add it to the memory or not
         assert strategy.experience is not None
         dataset = strategy.experience.dataset
-        pbar = tqdm.tqdm(dataset, desc="Exhausting dataset to create GDumb buffer")
-        for pattern, target_value, task_id in pbar:
-            target = torch.as_tensor(target_value)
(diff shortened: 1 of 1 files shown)


AntonioCarta:
Thanks @lebrice!

Your fix is only about the shape error, right? I may be wrong but from a quick look GDUMB seems to ignore the original task labels even with your implementation

Of course you are also correct about the other points. Unfortunately, most of the old strategies were implemented before the benchmark support for multi-task/concatenation/subsampling. We are trying to update them over time.

lebrice:
@AntonioCarta  I meant that the `task_id` here is a variable in the for-loop, rather than being `strategy.experience.task_label` , and is used to fetch a (possibly different)  `current_mem` and `current_counter` to be used for each sample in the dataset.

AntonioCarta:
Ok, got it. Thanks again for the PR.

PR #613, opened 2021-05-15
Refactored metrics to follow a GenericPluginMetric class

AndreaCossu:
This PR refactors most of the existing metrics following the newly introduced `GenericPluginMetric`. This change does not affect neither the metric behavior nor the user interface.

Forgetting and confusion matrix metrics are still implemented through the `PluginMetric` class since they have a slightly more complex flow.

Now the generic plugin can be used to easily build new metrics.


AndreaCossu:
I'll leave this open for few days, just in case anybody has any comments :smile:

Issue #616, opened 2021-05-17
Make CoPE example data incremental


Mattdl:
Putting the new data-incremental interface to use for the CoPE example, which is now still class incremental.
Can someone assign this issue to me? Thanks!

Mattdl:
I get an error when implementing the data incremental loader.

    -- >> Start of training phase << --
    -- Starting training on experience 242 (Task 0) from train stream --
    
    [...] RecursionError: maximum recursion depth exceeded while calling a Python object

Should I make a PR to have a working example?

vlomonaco:
Yes so that we can provide help in case you need it!

Mattdl:
I've added the PR to reproduce the error. See #543 which had the same error.
@lrzpellegrini you found a fix for #543, do you have an idea what may cause the error in this case?

lrzpellegrini:
Hi @Mattdl, sorry for the delay. I reproduced the bug on the cope example of your fork. The recursion bug is very similar to the one I previously fixed, but this seems to be linked to a different corner case. I'll fix it asap.

lrzpellegrini:
The last PR should fix the recursion issue and it will be merged soon. Let me know if it works!

Mattdl:
Hi thanks for the effort! The recursion bug seemed to be solved, but now after 201 iterations another problem in the corner case arises. It seems to happen when concatting for the replay buffer, i.e. in AvalancheConcatDataset.
This is the stack trace, it should be an index out of range exception:

    Traceback (most recent call last):
      File .../avalanche/training/plugins/cope.py", line 88, in before_training_exp
        AvalancheConcatDataset(self.replay_mem.values()),
      File .../avalanche/benchmarks/utils/avalanche_dataset.py", line 1616, in __init__
        super().__init__(ClassificationDataset(),  # not used
      File .../avalanche/benchmarks/utils/avalanche_dataset.py", line 306, in __init__
        self._flatten_dataset()
      File .../avalanche/benchmarks/utils/avalanche_dataset.py", line 1849, in _flatten_dataset
        flattened_list += self._flatten_subset_concat_branch(dataset)
      File .../avalanche/benchmarks/utils/avalanche_dataset.py", line 1904, in _flatten_subset_concat_branch
        last_c_targets.append(dataset.targets[idx])
    IndexError: list index out of range

I'll make another PR to make it reproducable.

Mattdl:
[...]

PR #623, opened 2021-05-18
Data incremental example CoPE

Mattdl:
Support needed for #616 , the Max recursion depth error was also a problem in a previous issue, where spawning new AvalancheDatasets per batch gets a too deep calling tree.


vlomonaco:
Hi @Mattdl for sharing a code example next time I suggest to create a commit on your personal fork (and share the link) rather than opening a PR here! :-)

Mattdl:
Hi @vlomonaco, makes sense! Other collaborators suggested making a PR with a [WIP] tag to show it's a draft PR.
So the policy has changed on this?

vlomonaco:
Yes, @Mattdl actually I may myself have pointed you towards this earlier, sorry :) But given that now a few PR are stalling it would be good to have only future integrations here and not examples, thanks!

Mattdl:
Ok I see! (: I actually wanted to merge this, because the CoPE example was still in the class-incremental setting.
Only afterwards the data-incremental setting was implemented, so I'm trying to update the example with this PR.

lrzpellegrini:
Hi Matt, thank you for the great work!

Apart from the performance issues pointed out in #650 and #653 (which seem related to Avalanche internals more than CoPe itself), do you think that this PR ready to merge?

AntonioCarta:
Before merging we should check that the data incremental example reaches the correct performance values. RIght now, it doesn't #650

Mattdl:
Problem is that I can't yet reproduce the results when applying the data incremental generator, and the setup is now too cumbersome to debug 😄 But if you want to close this PR, there is a TODO in the example and I can make a new PR when the other issues are resolved (although it might be harder to have a clear history on all the PRs/issues together)

PR #628, opened 2021-05-22
[TensorboardLogger] Add support for AWS and Gcloud buckets

Mathieu4141:
PyTorch's `SummaryWriter` support writing logs directly to AWS S3 buckets and Google Cloud buckets if the path provided starts by `s3://` or `gs://`. (GS buckets are only supported if tensorflow is also installed https://github.com/pytorch/pytorch/issues/24468#issuecomment-549415705_)

We were not able to use this functionality previously because of the Path conversion. 

🐛  I also noticed that the last logs were often missing (flushing is done automatically every 2 min, and if we don't do it manually at the end then the non flushed logs are lost). Closing the writer in the logger destructor fixes that issue.

### DIFF
--- a/avalanche/logging/tensorboard_logger.py
+++ b/avalanche/logging/tensorboard_logger.py
@@ -60,6 +60,9 @@ class TensorboardLogger(StrategyLogger):
         self.writer = SummaryWriter(tb_log_dir,
                                     filename_suffix=filename_suffix)
 
+    def __del__(self):
+        self.writer.close()
+
     def log_metric(self, metric_value: MetricValue, callback: str):
         super().log_metric(metric_value, callback)
         name = metric_value.name
--- a/avalanche/logging/tensorboard_logger.py
+++ b/avalanche/logging/tensorboard_logger.py
@@ -13,6 +13,7 @@
 Avalanche experiments. """
 
 from pathlib import Path
+from typing import Union
 
 from PIL.Image import Image
 from torch import Tensor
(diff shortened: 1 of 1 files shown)


vlomonaco:
Thanks @Mathieu4141 for this! :) @AndreaCossu will check it out and provide eventual feedback!

AndreaCossu:
Thank you @Mathieu4141 ! The PR looks good. Maybe you could check if the writer is still open before closing it in the `__del__`. The user may close it by hand resulting in an error. A try block should be enough.

Also, can you fix the errors raised by the tests (let me know if you need any help)? I'll merge the PR after these changes :smile:

Mathieu4141:
Thanks for reviewing!

`FileWritter` has a check to prevent errors on multiple closes (https://github.com/pytorch/pytorch/blob/master/torch/utils/tensorboard/writer.py#L1032-L1038)
```python
    def close(self):
        if self.all_writers is None:
            return  # ignore double close
        for writer in self.all_writers.values():
            writer.flush()
            writer.close()
        self.file_writer = self.all_writers = None
```
I can still add a double safety on avalanche side if you think it is worth it

For the unittests I'll wait for your investigation on the other PR, I think it is the same issue indeed :)

AndreaCossu:
Great, no need to double check then, you can leave it as is, thanks again!

AndreaCossu:
Hi @Mathieu4141, can you please update also this branch so that tests are run with new version? Thanks!

PR #629, opened 2021-05-23
[Replay] Add some memory selection strategies

Mathieu4141:
This extends the `ExperienceBalancedStoragePolicy` to be able to use different selection strategies. It expect the strategy to sort the exemplars, and so only need to be called the first time they are seen.

I implemented 3 strategies:
 - Random selection (which was the previous behavior)
 - Herding (as used in the [iCaRL paper](https://arxiv.org/pdf/1611.07725.pdf))
 - Closest to center (in feature space)

[RWalk](https://arxiv.org/pdf/1801.10112.pdf) introduced other strategies, like entropy or distance to boundary decision, but herding and random seems to be the more used / best strategies.

### DIFF
--- a/avalanche/benchmarks/utils/avalanche_dataset.py
+++ b/avalanche/benchmarks/utils/avalanche_dataset.py
@@ -35,11 +35,11 @@ from .dataset_definitions import ITensorDataset, ClassificationDataset, \
 
 try:
     from typing import List, Any, Iterable, Sequence, Union, Optional, \
-        TypeVar, Protocol, SupportsInt, Generic, Callable, Dict, Tuple, Literal
+    TypeVar, Protocol, SupportsInt, Generic, Callable, Dict, Tuple, Literal, \
+    Collection
 except ImportError:
     from typing import List, Any, Iterable, Sequence, Union, Optional, \
         TypeVar, SupportsInt, Generic, Callable, Dict, Tuple
-    from typing_extensions import Protocol, Literal
 
 T_co = TypeVar('T_co', covariant=True)
 TTargetType = TypeVar('TTargetType')
@@ -1494,7 +1494,7 @@ class AvalancheConcatDataset(AvalancheDataset[T_co, TTargetType]):
     (if they are subclasses of :class:`AvalancheDataset`).
     """
     def __init__(self,
-                 datasets: Sequence[SupportedDataset],
+                 datasets: Collection[SupportedDataset],
                  *,
                  transform: Callable[[Any], Any] = None,
                  target_transform: Callable[[int], int] = None,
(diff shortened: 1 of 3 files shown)


avalanche/benchmarks/utils/avalanche_dataset.py
```diff
@@ -33,13 +33,8 @@
 from .dataset_definitions import ITensorDataset, ClassificationDataset, \
     IDatasetWithTargets, ISupportedClassificationDataset
 
-try:
-    from typing import List, Any, Iterable, Sequence, Union, Optional, \
-        TypeVar, Protocol, SupportsInt, Generic, Callable, Dict, Tuple, Literal
-except ImportError:
-    from typing import List, Any, Iterable, Sequence, Union, Optional, \
-        TypeVar, SupportsInt, Generic, Callable, Dict, Tuple
-    from typing_extensions import Protocol, Literal
```
AntonioCarta, line 42: Some imports may seem useless but they are needed for type checking (not necessarily these lines), which is why the CI is currently failing.

avalanche/training/plugins/replay.py
```diff
@@ -30,7 +35,8 @@ class ReplayPlugin(StrategyPlugin):
     in the external memory.
```
AntonioCarta, line 35: storage_policy argument is missing from the docstring.

avalanche/training/plugins/replay.py
```diff
@@ -162,8 +176,9 @@ def __init__(self, ext_mem: Dict, mem_size: int, adaptive_size=True,
         :param total_num_classes: If adaptive size is False, the fixed number
                                   of classes to divide capacity over.
```
AntonioCarta, line 177: selection_strategy argument missing from the docstrin.

AntonioCarta (changes requested):
Thanks for the PR! I added a couple of minor comments. I will merge this as soon as they are fixed.
Feel free to open issues/feature requests for any missing selection strategy that you are interested in.

avalanche/benchmarks/utils/avalanche_dataset.py
Mathieu4141: From the logs in the CI it seems that `BaseStrategy` was the missing import, so it should be better now
Is there a way I can run the `check_syntax` job locally to test everything is ok before pushing ?

Also I think that the unittests require too much memory to be run from my fork. I'm not sure if there is something I can do about it ?
Screen Shot 2021-05-24 at 12 16 41 PM

Mathieu4141:
Thanks for your review! :)

avalanche/benchmarks/utils/avalanche_dataset.py
AntonioCarta: I've never done the syntax check locally but I think it's using flake8, according to this action: https://github.com/cclauss/Find-Python-syntax-errors-action

The disk failure is unrelated to your PR. We are working to fix it ASAP.

Issue #627, opened 2021-05-21
Class order in Confusion Matrix


Mathieu4141:
# Existing behavior

When using the Confusion Matrix in avalanche the classes are sorted per class id.

Here is a CM that can be obtained in avalanche now (0 and 3 were the last classes to be learn):
<img width="313" alt="Screen Shot 2021-05-21 at 5 46 11 PM" src="https://user-images.githubusercontent.com/23531109/119164250-83725d80-ba5c-11eb-9fae-e5f3b65a53ab.png">

# Desired behavior

In a Class-Incremental Learning setting, we usually want the classes sorted by the time they were observed in training. This allow to quickly assert if there is a bias toward the newly added classes, or toward the old classes.

For eg, here are the results showed in BiC and ICaRL papers:
<img width="340" alt="BiC" src="https://user-images.githubusercontent.com/23531109/119163843-16f75e80-ba5c-11eb-8e2c-7a1460f9ee76.png">
<img width="340" alt="ICaRL" src="https://user-images.githubusercontent.com/23531109/119163852-1959b880-ba5c-11eb-852c-e345bad29a44.png">

# Possible implementations

I think that this should only be implemented for the image rendered and not the tensor also returned by the metric, since we would loose class ids on the tensor

We could:

1. Have an optional parameter `classes_order` in the `StreamConfusionMatrix` metric

```python
        if self._save_image:
            if self.classes_order:
                cm_image = self._image_creator(exp_cm[self.class_order][:, self.class_order], display_labels=self.class_order)
            else:
                cm_image = self._image_creator(exp_cm)
            metric_representation = MetricValue(
                self, metric_name, AlternativeValues(cm_image, exp_cm),
                plot_x_position)
```

2. Provide an image creator that takes the order into account

```python 
class SortedCMImageCreator:
    def __init__(self, class_order: List[int]):
        self.class_order = class_order

    def __call__(self, tensor: Tensor) -> Image:
        return default_cm_image_creator(tensor[self.class_order][:, self.class_order], display_labels=self.class_order)
```

Option 2 seems cleaner, but more difficult to discover. Option 3 that combines the 2 could be to use a `SortedCMImageCreator` class by default when some `class_order` parameter is passed to `StreamConfusionMatrix` but no `image_creator` is set ?

What do you think?

If it is ok I'd like to make a PR for this 🙂 

EDIT: I can't seem to change the label, but this is not a bug :)

AntonioCarta:
[...]

PR #632, opened 2021-05-24
[evaluation] (CM) implement relative class order in confusion matrix

Mathieu4141:
Implementing class order in CM #627 cc @AndreaCossu 

Here is an exemple of CM obtained with the new defaults:
Screen Shot 2021-05-24 at 4 01 22 PM

## W&B

I tried to implement it for W&B too, but it seems that the UI re-order the classes alphabetically, so it doesn't work. We could prefix with the relative number (0. Class 3 / 1. Class 5 / 2. Class 9, ...) to get it sorted correctly

### DIFF
--- a/avalanche/evaluation/metrics/confusion_matrix.py
+++ b/avalanche/evaluation/metrics/confusion_matrix.py
@@ -8,9 +8,10 @@
 # E-mail: contact@continualai.org                                              #
 # Website: www.continualai.org                                                 #
 ################################################################################
-
+from numpy import arange, ndarray
 from typing_extensions import Literal
-from typing import Callable, Union, Optional, Mapping, TYPE_CHECKING, List
+from typing import Callable, Union, Optional, Mapping, TYPE_CHECKING, List, \
+    Sequence
 
 import wandb
 import numpy as np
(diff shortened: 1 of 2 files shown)


avalanche/evaluation/metrics/confusion_matrix.py
```diff
@@ -243,8 +245,8 @@ def __init__(self,
                  num_classes: Union[int, Mapping[int, int]] = None,
                  normalize: Literal['true', 'pred', 'all'] = None,
                  save_image: bool = True,
-                 image_creator: Callable[[Tensor], Image] =
-                 default_cm_image_creator):
+                 image_creator: Callable[[Tensor, Sequence], Image] =
```
Mathieu4141, line 248: Is it ok to make the assumption that `image_creator` expects the CM and the labels?
The issue is that in this `__innit__` we can't get the `class_order`, so we need to tell the image creator later what the labels are

AndreaCossu: Yes, I think this is a reasonable assumption.

AndreaCossu:
Thanks again @Mathieu4141 ! This seems ok (I'll just wait to the answer of my review before merging). After this integration we can tackle the same issue in the W&B matrix with the help of @digantamisra98 .

PR #637, opened 2021-05-27
added icarl plugin

andrew-r96:
Implementation of iCaRL algorithm with an Avalanche plugin.

Issues:
- Should I add to Avalanche the ResNet, data augmentation and main.py needed to reproduce the paper results?
- There are some utilities classes needed by the plugin that are currently stored inside the same file (e.g. DistillationLoss, NCMClassifier). Maybe they could be moved somewhere else.
- Still working to avoid having to pass `diff_transform` to the plugin (the difference between test_transform and train_transform). It should be easy but some of the solution I've tried led to a decrease in performances.

Apart from that the plugin works and achieves the results reported by the paper.

### DIFF
--- a/avalanche/models/__init__.py
+++ b/avalanche/models/__init__.py
@@ -14,3 +14,5 @@ from .mobilenetv1 import MobilenetV1
 from .dynamic_modules import *
 from .utils import *
 from .slda_resnet import SLDAResNetModel
+from .icarl_resnet import *
+from .ncm_classifier import NCMClassifier
--- a/avalanche/models/dynamic_modules.py
+++ b/avalanche/models/dynamic_modules.py
@@ -33,7 +33,7 @@ class DynamicModule(torch.nn.Module):
         """
         super().__init__()
 
-    def adaptation(self, dataset: AvalancheDataset):
+    def adaptation(self, dataset: AvalancheDataset = None):
         """ Adapt the module (freeze units, add units...) using the current
         data. Optimizers must be updated after the model adaptation.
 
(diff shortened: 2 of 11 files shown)


AntonioCarta:
Thanks @andrew-r96.

Are you sure about the distillation loss? I don't understand why you are using the `BCELoss`. Maybe your computations are equivalent to the common equations but from a quick glance they look different.

The rest of the implementation looks good. Having ICaRL as a plugin is a bit weird, I would prefer a separate strategy. The other components (distillation, NCM) can already be used indipendently.
For a quick fix, you can make a strategy `ICaRLPlugin` using the plugin and define the plugin in the same file, making it a private class for that file (`_ICaRLPlugin`).

I would move the distillation and modules in their appropriate places:
- modules go inside `models`
- we don't have a place for `DistillationLoss` yet. I would put it inside `training` in a separate file `training.losses`.

Remember to add a small test.

AntonioCarta:
> Still working to avoid having to pass `diff_transform` to the plugin (the difference between test_transform and train_transform). It should be easy but some of the solution I've tried led to a decrease in performances.

`diff_transform` are the buffer transformations? I'm also passing them manually in my own projects, as you are already doing.
It's not easy to use the `train_transform` because you have to remove the preprocessing steps (normalization, ToTensor), and leave the data augmentation (rotate, zoom, shift ...). We could try to provide a function that removes the transforms we know are not needed, but it could lead to weird and hard to find bugs. I would go with explicit arguments for this, at least the user is aware of the choice.

andrew-r96:
> Are you sure about the distillation loss? I don't understand why you are using the BCELoss. Maybe your computations are equivalent to the common equations but from a quick glance they look different.
Yes the result is actually different from `log_softmax + nll `. The reason I used it is that it's what they do in the original Theano implementation and in Lorenzo's PyTorch implementation. I'd assume it would work with the common equation as well (maybe even better?) but I'm not sure whether we should functionally modify the original implementation. 

By the way, the implementation of distillation is somewhat not standard too. Normally, one would compute the current loss and the KD loss on all the output units and sum them together. In this case, the one-hot target for the current sample is "merged" with the output of the old model on the same sample in the positions corresponding to the old classes. This works well in the class incremental setting but is problematic in the domain incremental one (or in any setting where old classes are observed again in future experiences).

> diff_transform are the buffer transformations?
Yes basically if `train_transform = [preprocessing, agumentation]` and `test_transform = [preprocessing]` then `diff_transform = [augmentation]`. To do herding or compute the class mean I need the elements to be just preprocesses but, when I use them for replay I need the augmentation part too. My idea was to select the samples from the dataset like this:

`memory[i] = AvalancheSubset(dataset, torch.where(pick == 1)[0])`

so that I could use eval() or train() on the dataset to get the transformation that I need. Sadly, when I tried this solution the performance dropped. I'll leave the explicit argument as you suggest.

AntonioCarta:
Ok, thanks for the explanation. I guess it doesn't make sense to make the DistillationLoss publicly available since it's customized for ICaRL. Maybe you can rename it to `ICaRLDistillationLoss` and explain in the docstring that it's not the classic KD loss.

Mathieu4141:
It's great to have iCaRL 🎉 

Maybe the plugin could be split into 3, to separate distillation, NCM and replay ? And iCaRL would be a base strategy with those 3 plugins?
There already are some replay plugins in avalanche (the part with `diff_transform` would need to be added though).

This way it would be easy to implement the hybrids used in iCaRL (by removing some of the plugins), and we could try NCM on other strategies easily

andrew-r96:
I've moved the different utilities (loss, NCMClassifier, etc.) in different files and wrote a test for each of them. I've also added a test for the icarl strategy and an icarl example in the examples directory. Should be good to go!

AntonioCarta:
Thanks @andrew-r96, great work!

PR #651, opened 2021-06-06
OpenLORIS GDrive Links Changed to Avl VPS

### LINKED ISSUE #617: Mirroring OpenLoris and Add new link for Auto-Download
We need to mirror OpenLORIS Data on our VPS. We had consent from the original authors.

vlomonaco:
This PR closes #617. It uses the new continualai vps with openloris mirrors to download the data.
This still needs to be tested since the vps is currently down.

### DIFF
--- a/avalanche/benchmarks/datasets/openloris/openloris.py
+++ b/avalanche/benchmarks/datasets/openloris/openloris.py
@@ -43,7 +43,7 @@ class OpenLORIS(DownloadableDataset):
         for name in data2download:
             if self.verbose:
                 print("Downloading " + name[1] + "...")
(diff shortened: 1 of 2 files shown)


vlomonaco:
@lrzpellegrini what do you think? For now I maintained only the GDrive links but I've removed the gdrive download method... I don't think it's usable for this big dataset.

lrzpellegrini:
I agree, OpenLORIS is huge and a direct download seems to be a better choice! Did you check that the dataset could be properly downloaded and unpacked using the new code?

vlomonaco:
I finally tested and it works now as expected. Anything to add? otherwise we can merge this.

Issue #487, opened 2021-04-02
Add metrics to inspect dataset and patterns


AndreaCossu:
By exploiting the callbacks `before_forward` and similar it is very easy to log samples of the training dataset, number of patterns, features statistics and other metrics.

We can discuss in here what it could be useful to add.

Mathieu4141:
I'd be interested to work on that :)

I'm not sure to know what you mean by number of patterns @AndreaCossu  ?

I was thinking that we could also visualize the evolution of the tasks/labels we train on

PR #652, opened 2021-06-06
[Metrics] dataset inspection

Mathieu4141:
I started to write some metrics for this feature #487 

I didn't added the doc yet, I wanted to get some feedback first :)

I implemented 2 different kind of metrics to inspect the dataset:

1. LabelsRepartition: track the repartition & evolution of labels (grouped by task) during training.
Screen Shot 2021-06-06 at 1 11 26 PM

2. ImagesSamples: plot some examples of images. They can be grouped by tasks, labels, or taken at random in the dataset.
Screen Shot 2021-06-06 at 1 13 40 PM

### DIFF
--- a/avalanche/evaluation/metric_definitions.py
+++ b/avalanche/evaluation/metric_definitions.py
@@ -52,7 +52,6 @@ class Metric(Protocol[TResult]):
 
         :return: The value of the metric.
         """
-        pass
 
     def reset(self) -> None:
         """
@@ -60,7 +59,6 @@ class Metric(Protocol[TResult]):
 
         :return: None.
         """
-        pass
 
 
 class PluginMetric(Metric[TResult], StrategyCallbacks['MetricResult'], ABC):
@@ -212,11 +210,11 @@ class GenericPluginMetric(PluginMetric[TResult]):
         super(GenericPluginMetric, self).__init__()
         assert mode in {'train', 'eval'}
         if mode == 'train':
-            assert reset_at in {'iteration', 'epoch'}
-            assert reset_at in {'iteration', 'epoch'}
+            assert reset_at in {'iteration', 'epoch', 'experience', 'stream'}
+            assert emit_at in {'iteration', 'epoch', 'experience', 'stream'}
         else:
-            assert reset_at in {'experience', 'stream'}
-            assert reset_at in {'experience', 'stream'}
+            assert reset_at in {'iteration', 'experience', 'stream'}
+            assert emit_at in {'iteration', 'experience', 'stream'}
         self._metric = metric
         self._reset_at = reset_at
         self._emit_at = emit_at
@@ -293,6 +291,12 @@ class GenericPluginMetric(PluginMetric[TResult]):
         if self._reset_at == 'experience' and self._mode == 'eval':
             self.reset()
 
+    def after_eval_iteration(self, strategy: 'BaseStrategy'):
+        super().after_eval_iteration(strategy)
+        self.update(strategy)
+        if self._reset_at == 'iteration' and self._mode == 'eval':
+            self.reset()
+
     def after_eval_exp(self, strategy: 'BaseStrategy'):
         super().after_eval_exp(strategy)
         if self._emit_at == 'experience' and self._mode == 'eval':
@@ -307,12 +311,6 @@ class GenericPluginMetric(PluginMetric[TResult]):
         super().before_eval_iteration(strategy)
         if self._emit_at == 'iteration' and self._mode == 'eval':
             return self._package_result(strategy)
-
-    def after_eval_iteration(self, strategy: 'BaseStrategy'):
-        super().after_eval_iteration(strategy)
-        self.update(strategy)
-        if self._reset_at == 'iteration' and self._mode == 'eval':
-            self.reset()
 
 
 __all__ = ['Metric', 'PluginMetric', 'GenericPluginMetric']
--- a/avalanche/evaluation/metric_definitions.py
+++ b/avalanche/evaluation/metric_definitions.py
@@ -262,7 +262,8 @@ class GenericPluginMetric(PluginMetric[TResult]):
 
     def after_training_iteration(self, strategy: 'BaseStrategy') -> None:
         super().after_training_iteration(strategy)
-        self.update(strategy)
+        if self._mode == 'train':
+            self.update(strategy)
         if self._emit_at == 'iteration' and self._mode == 'train':
             return self._package_result(strategy)
 
@@ -293,7 +294,8 @@ class GenericPluginMetric(PluginMetric[TResult]):
 
     def after_eval_iteration(self, strategy: 'BaseStrategy'):
         super().after_eval_iteration(strategy)
-        self.update(strategy)
+        if self._mode == 'eval':
+            self.update(strategy)
         if self._reset_at == 'iteration' and self._mode == 'eval':
             self.reset()
 
--- a/avalanche/evaluation/metric_definitions.py
+++ b/avalanche/evaluation/metric_definitions.py
@@ -52,6 +52,7 @@ class Metric(Protocol[TResult]):
 
         :return: The value of the metric.
         """
+        pass
 
     def reset(self) -> None:
         """
@@ -59,6 +60,7 @@ class Metric(Protocol[TResult]):
 
         :return: None.
         """
+        pass
 
 
 class PluginMetric(Metric[TResult], StrategyCallbacks['MetricResult'], ABC):
--- a/avalanche/evaluation/metric_utils.py
+++ b/avalanche/evaluation/metric_utils.py
@@ -302,5 +302,7 @@ __all__ = [
     "get_metric_name",
     "stream_type",
     "bytes2human",
+    "default_history_repartition_image_creator",
     "repartition_pie_chart_image_creator",
+    "repartition_bar_chart_image_creator",
 ]
--- a/avalanche/evaluation/metric_utils.py
+++ b/avalanche/evaluation/metric_utils.py
@@ -9,10 +9,12 @@
 # Website: www.continualai.org                                                 #
 ################################################################################
 
-from typing import Tuple, TYPE_CHECKING
+from typing import Tuple, TYPE_CHECKING, Dict, Union, Iterable
 
 import matplotlib.pyplot as plt
 import numpy as np
+from matplotlib.axes import Axes
+from numpy import ndarray, arange
 from torch import Tensor
 
 if TYPE_CHECKING:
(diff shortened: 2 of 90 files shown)


vlomonaco:
Hi @Mathieu4141! This is awesome, really! thank you so much! I remember dreaming about these kind of features when I was still a PhD :-) I think that making CL less of a black box through automatic visualizations and metrics will be fundamental for the feature of this field and the loggers/dashboards side of avalanche can really help with this!

As for the PR I'll let @AndreaCossu review it. Just a couple of suggestion on my side:

1. It would be nice to provide also a visualization like [this](https://d1fq16qvu9tpz9.cloudfront.net/uploads/ckeditor/pictures/109/content_Screen_Shot_2017-08-10_at_2.48.52_PM.png), to see how the classes representation varies over time (in each experience).
2. For the `ImagesSamples` I'd suggest also to let them be grouped by experiences as well!

Mathieu4141:
Thanks Vincenzo 😄 

I really liked the visualization you suggested, I tried to implement that:
Screen Shot 2021-06-06 at 5 40 54 PM

AndreaCossu:
Hi @Mathieu4141 ! Thanks for the PR. Results look really good, I like these features very much.
As for the code, I'll review it as soon as possible and provide you some feedback. 
Much appreciated :smile:

avalanche/evaluation/metric_definitions.py
```diff
@@ -212,11 +210,11 @@ def __init__(self, metric, reset_at='experience', emit_at='experience',
         super(GenericPluginMetric, self).__init__()
         assert mode in {'train', 'eval'}
         if mode == 'train':
-            assert reset_at in {'iteration', 'epoch'}
```
AndreaCossu, line 215: The idea behind the asserts was that if the metric works at training time it should reset and emit values only during training stages like `iteration` and `epoch`. Correspondingly, if the metric works at eval time it should reset and emit at `experience` or `stream`. Do you need to emit or reset in different stages?

avalanche/evaluation/metric_utils.py
```diff
@@ -216,10 +296,11 @@ def get_metric_name(metric: 'PluginMetric',
...
-    'get_metric_name',
-    'stream_type',
-    'bytes2human'
+    "default_cm_image_creator",
+    "get_task_label",
+    "phase_and_task",
+    "get_metric_name",
+    "stream_type",
+    "bytes2human",
+    "repartition_pie_chart_image_creator",
```
AndreaCossu, line 305: I think you should add also the other 2 newly created functions in the `__all__` since they are imported later on (`default_history_repartition_image_creator` and `repartition_bar_chart_image_creator`)

avalanche/evaluation/metrics/dataset_inspection.py
```diff
@@ -0,0 +1,144 @@
...
+    modes = []
+    if on_eval:
+        modes.append("eval")
+    if on_train:
+        modes.append("train")
+
+    return [
+        LabelsRepartitionPlugin(image_creator=image_creator, mode=mode)
+        for mode in modes
+    ]
```
AndreaCossu, line 144: can you add the `__all__`?

avalanche/evaluation/metrics/images_samples.py
```diff
@@ -0,0 +1,259 @@
...
+    return plugins
+
+
+def make_un_normalize(
+    means: Iterable[float], std: Iterable[float]
+) -> Normalize:
+    return Normalize(
+        mean=[-mean / s for mean, s in zip(means, std)],
+        std=[1 / s for s in std],
+    )
```
AndreaCossu, line 259: can you add the `__all__`?

AndreaCossu (commented):
The code seems ok in general. I left some specific comments.

AntonioCarta:
Two quick comments:
- this is a personal opinion, but I like to be able to decide how to visualize the data by myself (e.g. if I want to publish a figure). Therefore, I would prefer if it was possible to retrieve the raw data used to visualize LabelsRepartition
- ImagesSamples should probably bypass transformations. Right now, you can see that the images are distorted by the training transforms

avalanche/evaluation/metric_definitions.py
Mathieu4141: Yes for eg the LabelRepartition needs to be emited at stream / experience, it would be unnecessary to emit it every epoch, as we would emit the same thing each time, it only depends on the dataset

The methods within the class already allow that
For eg the implementation of `GenericPluginMetric .after_training` expects `self._emit_at == 'stream' and self._mode == 'train'`

avalanche/evaluation/metric_definitions.py
```diff
@@ -52,15 +52,13 @@ def result(self) -> Optional[TResult]:
 
         :return: The value of the metric.
         """
-        pass
```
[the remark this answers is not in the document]
Mathieu4141: Sorry my linter removed it and I forgot to add it back, will do :)

vlomonaco:
> this is a personal opinion, but I like to be able to decide how to visualize the data by myself (e.g. if I want to publish a figure). Therefore, I would prefer if it was possible to retrieve the raw data used to visualize LabelsRepartition

I totally agree, isn't this already possible through the "alternative values" current definition?

@Mathieu4141 if I understand correctly the `repartition_bar_chart_image_creator` is the plot without history but it's not a pie anymore (a barh instead). Is that correct? :)

avalanche/evaluation/metric_definitions.py
AndreaCossu: Since you need to emit at `experience` or `stream` wouldn't be better to set this as a `eval` metric? In this way we can maintain these lines as before. This would also bypass the training transform problem I think, since you would not operate during training mode.

AndreaCossu:
To address point 1 of @AntonioCarta I think the user could specifiy if he wants plot or raw data as output. Alternatively, the metric could always emit both and the loggers could choose the appropriate one.

avalanche/evaluation/metric_definitions.py
Mathieu4141: Does that mean only getting the info from the eval stream?
The eval and train label repartitions won't be the same, and having the train imbalance (due to limited/no memory) is something we want to be able to plot

AndreaCossu: You are right, I think we can proceed like you did, thanks.

Mathieu4141:
> > this is a personal opinion, but I like to be able to decide how to visualize the data by myself (e.g. if I want to publish a figure). Therefore, I would prefer if it was possible to retrieve the raw data used to visualize LabelsRepartition
> 
> I totally agree, isn't this already possible through the "alternative values" current definition?

Yes the alternate value will hold the raw data and the image.
Maybe we can have a boolean to create or not an image (as in the Confusion Matrix plugin), to avoid this step if unnecessary.

> @Mathieu4141 if I understand correctly the `repartition_bar_chart_image_creator` is the plot without history but it's not a pie anymore (a barh instead). Is that correct? :)

Yes, `repartition_bar_chart_image_creator` is a barh without history (you need to use TB slider for that).
There is also a `repartition_pie_chart_image_creator` that is a pie chart (without history) if preferred 🙂 

> ImagesSamples should probably bypass transformations. Right now, you can see that the images are distorted by the training transforms

I agree it would be great to have both possible (sometimes it can be interesting also to see the transformations) but I'm not sure how to bypass that in the training plugin

vlomonaco:
> I agree it would be great to have both possible (sometimes it can be interesting also to see the transformations) but I'm not sure how to bypass that in the training plugin

What about adding a new "transform_group" in the [AvalancheDataset](https://avalanche-api.continualai.org/autoapi/avalanche/benchmarks/utils/avalanche_dataset/?highlight=avalanchedataset#avalanche.benchmarks.utils.avalanche_dataset.AvalancheDataset) with void transformations? than you can lead the example from there. Do you have a better idea @lrzpellegrini?

AndreaCossu:
Hi @Mathieu4141 ! Do you need any help in completing the PR? The solution of `replace_transforms` suggested by @vlomonaco should solve the problem. Let me know :smile:

Mathieu4141:
Yes thanks, I think I understand how to do it with @vlomonaco 's suggestion, I'll update the PR during the weekend 😄

Mathieu4141:
I talked too soon, I'm actually not sure how to do that @AndreaCossu @vlomonaco 

The Replay plugin modifies the DataLoader, not the dataset. So if we want to have images from the full training set (new instances + memory), without the transforms, I'm not sure how to do it, unless we do something very custom for the plugin. But then if the user uses its own dataloader it won't work...

```python
    def before_training_exp(self, strategy: "BaseStrategy",
                            num_workers: int = 0, shuffle: bool = True,
                            **kwargs):
        """
        Dataloader to build batches containing examples from both memories and
        the training dataset
        """
        if len(self.ext_mem) == 0:
            return
        strategy.dataloader = ReplayDataLoader(
            strategy.adapted_dataset,
            AvalancheConcatDataset(self.ext_mem.values()),
            oversample_small_tasks=True,
            num_workers=num_workers,
            batch_size=strategy.train_mb_size,
            shuffle=shuffle)
```

Mathieu4141:
I added the doc and fixed Andrea's comments.

> this is a personal opinion, but I like to be able to decide how to visualize the data by myself (e.g. if I want to publish a figure). Therefore, I would prefer if it was possible to retrieve the raw data used to visualize LabelsRepartition

For that I made the `image_creator` optional, and if not given only the raw data are sent in the metric. (but the raw data is still accessible when the image_creator is used, from the `AlternativeValues` object.

---

If I didn't forget anything, bypassing the transformations for the images sampling is the only thing missing on that PR.

I'll wait if you have suggestions on how to do that from the dataloader 🙂

AndreaCossu:
Great, thanks @Mathieu4141 ! I see the problem for replay, but  I think that for the moment we can focus on the original dataset only and remove transforms via `dataset_without_transforms = strategy.experience.dataset.replace_transforms()`. 
We can think about replay in a later stage (just make this point clear in the documentation, so that the user knows what the metric is returning). As you said, a completely custom replay may not work with the metric but we may be able to isolate a subset of assumptions covering the most used replay cases. I think @AntonioCarta can give us useful tips here :smile:

vlomonaco:
After a brief discussion with @lrzpellegrini one option for the future could be to have in the BaseStrategy a custom attribute always updated with the concatenation of the current_exp + ext_mem dataset to be used for this case and others in the future.

AntonioCarta:
There are three points where attributes that correspond to training data:
- the experience's dataset
- the strategy's `adapted_dataset`
- replay plugins buffer

I expected this plugin to look only in the first two and ignore the replay buffer. I would ignore the buffer for now, if someone wants to log the buffer's content they can easily adapt the metric. Alternatvely, the metric could have an optional parameter (a closure) which implements a `get_current_data`. Example:
```
def get_current_data():
    return my_replay_plugin.buffer
```

Mathieu4141:
As suggested I changed to use `BaseStratey.adapted_dataset`, and made it clear in the doc where the plugin get the data.

One great side effect of using `replace_transforms` is that we don't need to reverse the normalization anymore, so the usage is more simple this way :)

Here are some ex from the train metric:
image

Issue #664, opened 2021-06-10
ADD early stopping plugin


AntonioCarta:
Use `stop_training` to implement a basic early stopping plugin.

TomVeniat:
Since early stopping was needed for #561, I took a stab at it and proposed an implementation [here](https://github.com/TomVeniat/avalanche/blob/3e3152a77941f3bf2029afd40aeb9efd015dca9b/examples/simple_ctrl.py#L37).
For now, it is just a utility class used in the simple_ctrl.py example but I can create a proper PR to integrate it into the training module.

Before that I would need your help to fix the 2 main problems I see with the current  implementation:
- It does the checks inside of `after_training_epoch` instead of `after_eval_exp`, causing a lot of unnecessary checks when the `eval_every` parameter of the strategy is high. The reason is that I didn't find a way to make the distinction between an evaluation triggered during training (the periodic evaluations) and an evaluation triggered by the user calling `strategy.eval()`. Since we don't want to perform any early-stopping checks in the user-triggered evaluations, I used `after_training_epoch`.
- It will always be off-by-one, even if we use `after_eval_exp`. Since the evaluator is added to the plugins [in the strategy constructor](https://github.com/ContinualAI/avalanche/blob/a2e2fb09f77eaecad8dbbe74b4b78ab737b7e464/avalanche/training/strategies/base_strategy.py#L125), the `EarlyStopperPlugin` callbacks are called before the metrics callbacks, causing the early-stopping checks to always happen before the results of the metrics are available. 

Do you have any suggestions on how to overcome these issues?

AntonioCarta:
> It does the checks inside of after_training_epoch instead of after_eval_exp, causing a lot of unnecessary checks when the eval_every parameter of the strategy is high. The reason is that I didn't find a way to make the distinction between an evaluation triggered during training (the periodic evaluations) and an evaluation triggered by the user calling strategy.eval(). Since we don't want to perform any early-stopping checks in the user-triggered evaluations, I used after_training_epoch.

Using `after_training_epoch` is correct. The unnecessary checks should not be a problem since they are not expensive. I would prefer to keep the periodic evaluation logic and the early stopping logic as independent as possible.

> It will always be off-by-one, even if we use after_eval_exp. Since the evaluator is added to the plugins in the strategy constructor, the EarlyStopperPlugin callbacks are called before the metrics callbacks, causing the early-stopping checks to always happen before the results of the metrics are available.

[...]

PR #670, opened 2021-06-17
Implement early-stopping

TomVeniat:
Implementation of #664.

Maybe we could add the option to define the patience in terms of iterations instead of epochs, but I think that it can't be easily implemented for now since there is no simple way to keep track of the iteration count in the current training procedure.


AntonioCarta:
Thanks @TomVeniat! I will add the iteration counter in a separate PR.

Issue #600, opened 2021-05-06
Lazy definition of benchmarks


AntonioCarta:
It would be useful to be able to define lazy streams. Long streams are quite slow to load because they instantiate everything at init time. If `__len__` is necessary, we could just allow lazy streams but with a fixed length as a first solution.

Low priority since this is mainly a performance improvement.

AntonioCarta:
Fixed by #977.

PR #671, opened 2021-06-17
Support for lazily created streams

lrzpellegrini:
This PR introduces the support for lazy streams (discussed in #600)

This PR introduces some changes on the internals of `GenericCLScenario` and adds proper helpers to support the creation of streams based on generator expressions or functions.

This also introduces the support for dropping the references to previous experiences to allow for a more fine-grained memory management (to be documented).

Beware that the helper used to create data incremental benchmarks and the helper used to create a validation stream will still work in non-lazy mode (will load all experiences at once). This behavior will be fixed in the future.

The behavior for non-lazy benchmarks will not be affected by these changes.

### DIFF
--- a/avalanche/benchmarks/generators/benchmark_generators.py
+++ b/avalanche/benchmarks/generators/benchmark_generators.py
@@ -335,6 +335,7 @@ dataset_benchmark = create_multi_dataset_generic_benchmark
 filelist_benchmark = create_generic_benchmark_from_filelists
 paths_benchmark = create_generic_benchmark_from_paths
 tensors_benchmark = create_generic_benchmark_from_tensor_lists
+lazy_benchmark = create_lazy_generic_benchmark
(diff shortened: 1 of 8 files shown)


vlomonaco:
Hi @lrzpellegrini this is awesome, thanks! We are looking forward to see this integrated in the benchmark generators (and manipulators). Do you thing this could be done with a simple flag? Also, what's the disadvantage of making the whole system lazy?

lrzpellegrini:
For the moment there are no disadvantages in supporting lazy streams. However, for the vast majority of helpers adding a lazy option won't help in reducing times. The helpers that can benefit the most from a lazy approach are the ones that load the whole dataset in memory, but we currently don't have any of those. In the future, we may want to implement a path-based tensor dataset, in which case the laziness may help a lot.

PR #675, opened 2021-06-19
Added support for metric computation separated for each task label and added metric plugin test 

AndreaCossu:
This PR brings new functionalities for the performance metrics (accuracy, loss, forgetting, backward transfer).  
* The accuracy is capable of monitoring performance for each task label separated. In particular, task labels defined at the level of patterns are supported.
* The loss is able to monitor performance for different task labels **only** if task labels are defined for the entire experience, not for each pattern. This is due to the fact that loss computation is bundled within base strategy and cannot be computed for each pattern but only for the whole minibatch.
* The Forgetting/backward transfer have a more general API (they can be implemented with custom metrics and not only accuracy by exploiting the general class provided for both of them). Moreover the new `TaskForgetting, TaskBWT` compute the respective value separated for each task label, even if labels are defined for each pattern. However, to do that, the initial values for the metrics (e.g. accuracy) are computed on the training phase and not on the eval phase as for the other Forgetting/BWT metrics. This is due to the fact that task labels may be spread over all eval experiences.

The aforementioned metrics collect task-based values into a dictionary kept by the related standalone metric.

The task label information is taken from `strategy.experience.task_labels`, which is the list of task labels associated to that experience. If it contains more than 1 task labels, no task signature is appended in the metric name. If it contains only one label, that label is used for the metric signature.
To take the task labels for each pattern in metric computation, `strategy.mb_task_id` is used.

The tests are improved by executing 3 configurations with Single Incremental Task, Multi Task (both on MNIST dataset) and on a fake benchmark with task labels defined for each pattern. The obtained results have to match the results saved in three separate files representing the target metric values.

**What is missing**:

1. Update API doc to better describe these changes
2. Extend the task separated computation also for Confusion matrix and the system metrics. These last ones cannot be extended to manage task labels defined for each pattern (since patterns are processed in parallel in minibatches).

Let me know if you have any comments


AndreaCossu:
It seems that by using Pytorch 1.8.1 (torchvision 0.9.1) instead of pytorch 1.9.1, the errors related to py3.9 disappear. Should we limit the pytorch version during installation? I'm not sure what is changing between the two versions though :thinking:

Issue #676, opened 2021-06-20
Add IL2M strategy


Mathieu4141:
Hi!

I'd like to implement IL2M in avalanche, along with the metric that was proposed to show bias toward new classes.
reference paper: https://openaccess.thecvf.com/content_ICCV_2019/papers/Belouadah_IL2M_Class_Incremental_Learning_With_Dual_Memory_ICCV_2019_paper.pdf
code: https://github.com/EdenBelouadah/class-incremental-learning/tree/master/il2m

I don't know if there is already a plan to port this or if it's ok if I start working on that?

vlomonaco:
@EdenBelouadah for sure wanted to port it, but I think she is focusing on #527 right now :-) 
@EdenBelouadah, @AntonioCarta is it ok if @Mathieu4141 starts working on IL2M? @EdenBelouadah maybe you can review the PR!

EdenBelouadah:
Hello,
Yesterday, I started working again on SIW (#527), and the plan was to pass after that to IL2M.
If you want, you can start working on IL2M, and I will join you later. Maybe by the end of this week of the beginning of the other.

AntonioCarta:
Fine by me.

If you have any doubts feel free to ping me on the ContinualAI slack.

Mathieu4141:
Great 😄 

I started to write a metric for the Mean Scores comparison you made in the paper #683 
The `MeanScores` metric can be reused for the plugin :)

PR #683, opened 2021-06-23
[metrics] (MeanScores) Add a mean scores metrics inspired from IL2M

Mathieu4141:
To prepare #676 

This is a metric that was used to show bias toward new classes in the IL2M paper.
The scores of the true class are averaged by past and new classes.

 - For the train plugin I only used the last epoch, the graphs are hard to read if we add more
 - For the eval plugin the full stream needs to be considered, so that we can compare old and new classes

This is what I get for the simple example:
Screen Shot 2021-06-23 at 10 27 40 PM

The `MeanScores` metric can be reused for the `IL2MPlugin`

Edit: Ready for review :)

### DIFF
--- a/avalanche/evaluation/metrics/mean.py
+++ b/avalanche/evaluation/metrics/mean.py
@@ -70,6 +70,17 @@ class Mean(Metric[float]):
         self.summed = 0.0
         self.weight = 0.0
 
+    def __add__(self, other: 'Mean') -> "Mean":
+        """
+        Return a metric representing the weighted mean of the 2 means.
+
+        :param other: the other mean
+        :return: The weighted mean"""
+        res = Mean()
+        res.summed = self.summed + other.summed
+        res.weight = self.weight + other.weight
+        return res
+
 
 class Sum(Metric[float]):
     """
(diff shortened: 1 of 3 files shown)


AndreaCossu:
Thanks @Mathieu4141, seems good! Have you considered building your plugin metric on the `GenericPluginMetric` interface? If it was possible, it would make the metrics more homogenous. Let me know :smile:

Mathieu4141:
> Thanks @Mathieu4141, seems good! Have you considered building your plugin metric on the `GenericPluginMetric` interface? If it was possible, it would make the metrics more homogenous. Let me know 😄

@AndreaCossu I tried at first to do it with `GenericPluginMetric`. However this plugin doesn't quite fit the template so I switch to a custom plugin, and it simplified things a lot
Some of the issues I had:
 - The eval plugin would get reset during eval, so we would loose the new/old classes distinction that can only be taken from the previous training. In a custom plugin I simply get the classes in `before_training_exp`, and reset at `before_training`
 - The train plugin would work on all epochs, but the graphs are really difficult to read and interpret then, doing it on the last epoch seems better

Those issues are really specific to this plugin, so I didn't want to change the generic class only for that

AndreaCossu:
Thank you again @Mathieu4141 , seems all good to me, I will merge this :smile:

AndreaCossu (approved):

PR #691, opened 2021-07-01
Stream completeness check for test stream

AndreaCossu:
This PR extends the `EvaluationPlugin` by making it able to check if the stream provided during `eval` or `periodic_eval` is the complete test stream of the benchmark. This is mainly done to avoid metric inconsistency for stream plugin metrics like stream accuracy. Let me know if you see other use cases.

The benchmark has to be passed as input to the plugin.
The boolean parameter `strict_checks` controls if the plugin will emit only warnings (when `False`) or raise errors (when `True`).

Possible behaviors:
1. Benchmark not provided: warning/errors raised during plugin init
2. Benchmark provided: if the length and the experience number in the input stream are the same as the ones in the `benchmark.test_stream`, no warnings/errors. Otherwise a warning/error is raised on each "wrong" `eval` call.

The check relies on the position of experiences in the original stream through `experience.current_experience` and on the possibility to call `enumerate` on the stream. @lrzpellegrini , could this last requirement be an issue for lazy streams?

With respect to existing implementations, the only change is that the user will see a warning appear during init, but the code will run anyway.

Similar checks could be performed on the `train` calls on the training stream, but since we have no training stream metrics the user should not experience any inconsistency in the metrics.

Currently, this implementation returns a warning/error if user calls `eval` on a stream which is not the test. This can be tackled in future commits (e.g. by specifying the admitted streams). First, I wanted to know if you see any problems in this implementation.


AndreaCossu:
As a minor note, I set `default_logger` to `None` since otherwise it would raise warnings on imports (for example when importing the strategies) even if it is not explicitly created by the user.

PR #695, opened 2021-07-12
Added forward transfer

### LINKED ISSUE #212: Add (Lopez-paz, 2017) Metrics: ACC, FWT & BWT

AndreaCossu:
This builds on #493 to implement the forward transfer.

The problem is mainly about computing the model accuracy at random initialization on all experiences. This is now done by adding to the `BaseStrategy` a `periodic_eval` call before training. The user is expected to: 

1. Enable periodic eval
2. Pass the entire test stream to the `eval_streams` parameter of the `strategy.train()` call. This is the recommended behavior in any case, because otherwise the evaluation plugin will raise a warning/error, depending on its configuration.

This is not the only way to implement forward transfer. For example, the forward transfer metric could automatically call an `eval` on the entire test stream to avoid relying too much on the user calls. We can discuss other ways.

In case you think of a better way to proceed, I will wait some days before moving on to fully test this implementation.

This closes #212


vlomonaco:
Nice @AndreaCossu! Can you add also an example about this? thanks :)

AndreaCossu:
I will add the forward transfer in the example about the evaluation plugin and I will provide a usage example directly in the API doc once we refine it.

AndreaCossu:
@AntonioCarta what do you think about the `periodic_eval` call also before training starts? I think this is the main change related to the training module.

AntonioCarta:
> 
> 
> @AntonioCarta what do you think about the `periodic_eval` call also before training starts? I think this is the main change related to the training module.

I think it's ok for now, but it's an ugly hack. Everyone gets a performance hit because of this. I also don't think it completely solves the problem with metrics. Different metrics will need a different evaluation loop and we will have to adjust it again.

PR #698, opened 2021-07-16
Improved dependency management using Docker containers

### LINKED ISSUE #611: Fix dependencies version
We should fix dep. versions towards a more stable release of the tool.

ggraffieti:
Dependency management was a bit confusing in Avalanche, especially during unit test. The conda environment were created from scratch every time the software was tested. This leads to different unwanted problems, like: i) tests are slow ii) test fails because conda (for any reason) was not able to create the environment iii) a dependency broke the software, so the test failed. 
It was not easy to spot the main reason of a test failure, since we need to control the action console every time. Moreover, we did not have a solid dependency management, where dependencies are refreshed periodically and test ONLY on new environment (and not on new software) are performed. 
This PR tries to solve these issues in the following ways: 
- The test are now performed using static environments and docker containers. The environment is already contained in the container so there is no need to recreate it every time. This allows to testing ONLY the new software, without performing an unwanted integration test at the same time. 
- The dependencies are updated periodically, once a week (midnight of every Sunday). The environment will be recreated from scratch with the latest version of all the dependencies. After that a complete unit test is performed. In this case the software is not changed, but only the new dependencies are tested. 
- If everything is good (environments created correctly and all the test passed) the new environment are pushed to the [avalanche-docker](https://github.com/ContinualAI/avalanche-docker) repository where the corresponding docker containers are created and pushed automatically to docker hub. All the subsequent tests will use the new containers. 
- If something went bad an issue will be automatically created by the @ContinualAI-bot, showing the difference between the last environment and the new environment. This allows to easily spot the changed dependencies and fix the problems more easily. Note that if something goes wrong in the creation of a new environment all the unit test continue to work since the Docker image wont be changed. 

All these phases are independent for each python version. 

The Docker container can be visulized [here](https://hub.docker.com/search?q=continualai&type=image).

This PR resolve #611


ggraffieti:
No comment, I guess it's good for everyone. 
I'll close this PR.

vlomonaco:
Thanks @ggraffieti for this! *-*

PR #706, opened 2021-07-25
[WIP] Fix W&B Artifacts Integration and minor errors

digantamisra98:
Fixing some errors with the current artifacts integration.


digantamisra98:
It seems that the `copy.deepcopy(list(strategy.model.parameters()))` does work but returns the params as a list cloned from the source generator object in the `evaluation/metrics/checkpoint.py` which when used in `logging/wandb_logger.py` doesn't seem to pass through the `if isinstance(value, Tensor):` line since it demands a Tensor while the metric returns a List.

examples/wandb_logger.py
```diff
@@ -21,6 +21,8 @@
 from os.path import expanduser
 
 import argparse
+
+from click import Path
```
AndreaCossu, line 25: is this import used somewhere?

digantamisra98: Oh, my bad, added it by mistake.

AndreaCossu:
@digantamisra98 any news on this?

digantamisra98:
@AndreaCossu I am no longer at W&B. But I am happy to finish this but won't be able to before December. Thanks!

lrzpellegrini:
Added in #1463

PR #707, opened 2021-07-28
FIX JointTraining off-by-one data concatenation

AntonioCarta:
Bug fix.
Also, now it raises an error when `JointTraining` is trained more than once.


vlomonaco:
@AntonioCarta was this the error you were encountering with CORe50?

AntonioCarta:
> 
> 
> @AntonioCarta was this the error you were encountering with CORe50?

Nope, but I also fixed that error (it was on my code).

PR #709, opened 2021-07-29
Docker nightly release released every night

### LINKED ISSUE #708: Nightly Docker release is scheduled for pull requests
The release of the nightly Docker image must be triggered only by commits to the main branch, not by pull requests. 
Problem -> release action triggered by the completion of unit test -> unit test executed in PR -> docker release triggered

No elegant way to pass information to the release action about the event that triggered the unit test. 

Most elegant solution (still not elegant in an absolute way) -> include the docker release in the unit test workflow, in order to have access to the event that triggered the tests. 

Another solution, a real nightly build -> release avalanche every night (maybe we could check the date of the last commit?) 
Problem with this solution: we don't have an updated version of avalanche every time we push something to master.

ggraffieti:
- Nightly Docker release of avalanche build every night (at midnight). If the last commit to master is older than 24h, the build is skipped (nothing new from the last nightly release). 
- Solves the problem of too frequent nightly Docker builds (every time a unit test is performed, both on the main branch and PRs). 

Resolves #708


vlomonaco:
Great job @ggraffieti! can we push it also here? https://docs.github.com/en/actions/guides/publishing-docker-images

ggraffieti:
> Great job @ggraffieti! can we push it also here? https://docs.github.com/en/actions/guides/publishing-docker-images

I check the github packages site and seems that we have a limit to 2GB of storage (pro account)
https://github.com/features/packages#pricing

Our docker image is more than 3GB (mainly due to the python environment). Don't know if we can use this service for free. 
Maybe we could link the docker image on docker hub on the repo.

vlomonaco:
ok I will add it in the getting started guide and in the readme!

ggraffieti:
> ok I will add it in the getting started guide and in the readme!

Actually my bad, these prices only applies to private repositories, we should have infinite space for free!

ggraffieti:
I'll add the docker image to the github packages of the repo and I'll close the PR by myself.

PR #721, opened 2021-08-17
Gss implementation

gab709:
Hi, 
I added the Gss_greedy strategy and a test file in the example folder,  to get the same results as the paper: https://papers.nips.cc/paper/2019/file/e562cd9c0768d5464b64cf61da7fc6bb-Paper.pdf


AntonioCarta:
Thanks @gab709! Can you clean the git history? I see some useless files that are committed.

PR #727, opened 2021-08-27
Updated pastj after experience in CWR* and added support for logging metrics returning dictionary

AndreaCossu:
I noticed that the count of encountered patterns for each class is not updated in CWR*. This PR fixes the problem. 
The problem affected only NIC scenarios, so current examples should be fine.

@lrzpellegrini let me know if I misunderstood the CWR* behavior.


AndreaCossu:
The last commits add support for loggers to manage metrics returning dictionary as value

AndreaCossu:
I'm merging this since it is open for a while. @lrzpellegrini feel free to let me know if the one-line change in CWR* does not convince you.

PR #732, opened 2021-09-02
Added helper method to convert nn.Modules to multi-head models

AlbinSou:
I left both the method and the test in separate files for now, let me know if you want me to integrate them to existing ones.

I think its working quite well, however they are few things to be aware of.

In particular, one thing I could not get around ( but I don't think I should ), is to make pytorch return a state_dict totally ignoring the MultiTaskDecorator class. So the state dict it returns contains the two "modules" from this class which are the "model" object and the "classifier". In any case, if a user wants to load a state dict from a multi head model he won't be able to load it into the non-multi head equivalent so I don't think this is a problem. 

Another related problem will appear when accessing the names of the modules, since now calling something like model.named_modules() will return the names prepended with "model", i.e "model.layer1.weight" instead of just "layer1.weight" . Unless the user is relying on these exact names it won't affect the behavior.

However, accessing and changing the values of model attributes ( be it modules or non modules ) is possible in a transparent manner ( accessing an attribute of the decorator directly gives a reference to the one of the underlying model ).  That way it won't add another layer of complexity in case the user wants to interact with its model attributes during training.

Test wise, I put some very basic tests but I think I will add some more, in particular I would want to test the interaction with plugins that access or change model attributes during training (I'm open to suggestions).

### DIFF
--- a/avalanche/models/__init__.py
+++ b/avalanche/models/__init__.py
@@ -17,3 +17,4 @@ from .slda_resnet import SLDAResNetModel
 from .icarl_resnet import *
 from .ncm_classifier import NCMClassifier
 from .base_model import BaseModel
+from .helper_method import as_multitask
--- a/avalanche/models/helper_method.py
+++ b/avalanche/models/helper_method.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+import torch
+import torch.nn as nn
+
+from avalanche.models.dynamic_modules import MultiTaskModule, MultiHeadClassifier
+
+class MultiTaskDecorator(MultiTaskModule):
+    """ 
+    Encapsulates an existing nn.Module to make it subclass MultiTaskModule,
+    the user should still be able to interact with the encapsulated module as if
+    it was the module itself. 
+    
+    The only things that change are the following, the classifier from the
+    given model will be replaced by a MultiHeadClassifier, and the forward()
+    implementation will be overwritten by one that accepts task labels. 
+    The encapsulated module will then be automatically extended to
+    fit new classes during calls to model.adaptation()
+    """
+    def __init__(self, model: nn.Module, classifier_name: str):
+        """
+        :param model: pytorch nn.Module that does not support multitask
+        :param classifier_name: attribute name of the existing classification
+                                layer inside the module 
+        """
+        self.__dict__['_initialized'] = False
+        super().__init__()
+        self.model = model
+        self.classifier_name = classifier_name
+
+        old_classifier = getattr(model, classifier_name)
+
+        if isinstance(old_classifier, nn.Linear):
+            in_size = old_classifier.in_features 
+            out_size = old_classifier.out_features 
+            # Replace old classifier by empty block
+            setattr(self.model, classifier_name, nn.Sequential())
+        elif isinstance(old_classifier, nn.Sequential):
+            in_size = old_classifier[-1].in_features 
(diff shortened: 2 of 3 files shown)


AntonioCarta:
Thanks @AlbinSou!

> I left both the method and the test in separate files for now, let me know if you want me to integrate them to existing ones.

The file is ok. Can you add the import inside `models/__init__.py`? I would like to be able to import it directly with
```
from models import as_multitask
```

> In particular, one thing I could not get around ( but I don't think I should ), is to make pytorch return a state_dict totally ignoring the MultiTaskDecorator class. So the state dict it returns contains the two "modules" from this class which are the "model" object and the "classifier". In any case, if a user wants to load a state dict from a multi head model he won't be able to load it into the non-multi head equivalent so I don't think this is a problem.

This is ok. The API call needs to be transparent for the user but it's ok if the resulting model changes names or other internal attributes.

Can you do one more change? Instead of using a random initialization for the `MultiHeadClassifier`, you could reuse the parameters of the original classifier. Also, you should add a test checking that the output of the two models is exactly the same (of course only for task ID 0).

AntonioCarta:
Unfortunately, we can't test the GPU during the continuous integration. You should use the flag `USE_GPU` in your tests to remove the CI errors.

AntonioCarta:
@AlbinSou please let me know when you are ready to merge.

AlbinSou:
> @AlbinSou please let me know when you are ready to merge.

I think we can merge.

I just want to make sure of something; when trying to add a test with LwF it raised an error but I think this is due to LwF not working in multitask ( error raised when calling prev_model(x) : task_labels is not given as arguments to forward ). I guess this behavior is being fixed by other contributors since I saw an issue related to that.

AntonioCarta:
> I just want to make sure of something; when trying to add a test with LwF it raised an error but I think this is due to LwF not working in multitask ( error raised when calling prev_model(x) : task_labels is not given as arguments to forward ).

My last PR should have fixed this. Can you try to pull from master and check this again?

AlbinSou:
Hmm, it should have fixed it indeed. However, I now get an error since I have not implemented the forward_all_tasks method in my decorator, which did not exist before the pull request.

However, I'm surprised that this one is not already implemented in the MultiTaskModule ( as it is the case for forward() ). Is this done on purpose ? The documentation indicates that it should return a dictionary that make the correspondence {task_id: output}, which should be possible to do in the MultiTaskModule directly by recursively calling forward_single_task on all known tasks ( a list should be kept inside this same object I guess to keep track of all known tasks ).

I could implement this method in the decorator otherwise but I feel like it makes more sense to implement it directly in the MultiTaskModule class.

AntonioCarta:
You're correct, it should be in the base class. I will add it.

AlbinSou:
> You're correct, it should be in the base class. I will add it.

Advise me when you will have pushed the changes to the master branch, that way I can push the last tests and we can merge then.

AntonioCarta:
Sure. I'm really busy with some experiments right now but I will try to push the changes as soon as I can.

AntonioCarta:
@AlbinSou I pushed the changes you needed. You should be able to finish the PR now.

AlbinSou:
> @AlbinSou I pushed the changes you needed. You should be able to finish the PR now.

Ok, it's ready to merge then !

AntonioCarta:
Thanks again!

Issue #734, opened 2021-09-05
SynapticIntelligence identical to Naive


tachyonicClock:
🐛 **Describe the bug**
I was confused to find SynapticIntelligence performing comparably to naive. To investigate it further I used a manual seed to see if the loss was being correctly used by the strategy. I found that the two methods perform identically when they are given the same seed.

🐜 **To Reproduce**
Run naive with a manual seed
`torch.manual_seed(0)`
Run SynapticIntelligence with a manual seed
`torch.manual_seed(0)`
The results produced will be identical since the loss is not being correctly applied
![image](https://user-images.githubusercontent.com/37557393/132113279-7ca6551b-4e56-455a-bff7-5483a8dbf944.png)

🐝 **Expected behavior**
I expected SynapticIntelligence to perform better.

🐞 **Screenshots**
I think this might be the issue but I'm not familiar enough with the implementation to be 100% certain
![image](https://user-images.githubusercontent.com/37557393/132113263-2b1b656a-c904-4fdf-b3e5-6b2187e391e5.png)

tachyonicClock:
Does this look right?
@AndreaCossu 
@AntonioCarta

AntonioCarta:
Your fix is correct.

@AndreaCossu can you check this problem? I think you did the original implementation.

AndreaCossu:
I can check that, but the original implementation is from @lrzpellegrini . He implemented the SI version they uses for AR1*, which I don't think is exactly the same of the SI paper (the strategy raises a warning for this reason).

AndreaCossu:
@tachyonicClock on which benchmark did you run your experiment?

tachyonicClock:
This is split MNIST. The metric is accuracy on seen tasks.

AndreaCossu:
ok, so I think the performance is fine. To reproduce the experiments of the paper, SI needs to be applied on a multi-headed model.

PR #735, opened 2021-09-05
Calculate loss before backpropogation

tachyonicClock:
Results are no longer identical when the seed is identical implying that the surrogate loss (regularisation) is being used. However since it does not perform as well as I would expect I'm a little concerned that I might not have fixed it.

image
image

#734


AndreaCossu:
Thanks for the fix  @tachyonicClock , on which benchmark did you run your experiments?

AndreaCossu:
The experiments were on Split MNIST so I think the performance is fine. I'm merging this.

PR #750, opened 2021-09-19
Updated Dead Link

DannyFGitHub:
In:
https://avalanche.continualai.org/from-zero-to-hero-tutorial/03_benchmarks#benchmarks-generators

Text Extract:
"Of course also the basic utilities ImageFolder and DatasetFolder can be used. These are two classes that you can use to create a Pytorch Dataset directly from your files (following a particular structure). You can read more about these in the Pytorch official documentation here."

Link Leads to:
https://pytorch.org/docs/stable/torchvision/datasets.html#imagefolder

I believe the link should now be:
https://pytorch.org/vision/stable/datasets.html#torchvision.datasets.ImageFolder


AndreaCossu:
Did in #752

PR #752, opened 2021-09-21
Updated the old Dead Link to ImageFolder docs

DannyFGitHub:
Old Link Leads to a 404:
https://pytorch.org/docs/stable/torchvision/datasets.html#imagefolder
The new link appears to be:
https://pytorch.org/vision/stable/datasets.html#torchvision.datasets.ImageFolder


AndreaCossu:
Thanks!

PR #759, opened 2021-09-23
Fix possible unbounded iterator for classes in experience

AndreaCossu:
The usage of a `defaultdict` for the classes in the experience may generate unbounded iterators (issue #758 ), since an experience id not present in the dictionary simply returns `None` instead of raising an error which would stop the iterator. 
An explicit check solves this problem. I left the case of returning `None` separated, because there may be cases (unsupervised learning?) in which the target is `None` but the experience is present.

@lrzpellegrini does this seem good to you?


AndreaCossu:
This solution seems to break other things, I will have to check it out.

AntonioCarta:
Do we really need the special case with `None`? It doesn't make sense to me. The unsupervised learning setting is a completely different case, where you should have an empty set (or an error). Here, it should never return a value for experiences that do not exist (yet).

Also, shouldn't this be a list like the experience's stream?

AndreaCossu:
The defaultdict behavior seems to be useful only for lazy scenarios, where new experiences come in dynamically. I put a patch for this case but it is not very elegant, since there may be a better way to solve the problem by tweaking the lazy scenario implementation directly. I will leave this PR open for discussion.

AndreaCossu:
I'll merge this upon tests completion. The bug fix for unbounded iterator will require a better handling, but for now is at least solved. I also added the `eps` damping parameter to Synaptic Intelligence strategy.

PR #760, opened 2021-09-24
switch metrics to train_iterations

### LINKED ISSUE #742: allow to change default download location for datasets
Right now the default dataset location is set to `.avalanche/data`. It would be nice to make it customizable, either by a configuration file in the folder `.avalanche` or using an environment variable.

### LINKED ISSUE #756: FIX gdumb model reset
We should fix this problem:

> Retrain from scratch was missing in the test. Up to now it is user responsibility to recreate the model+optimizer in the loop over steps.
I tested and performance is ok, you can merge.

See #136

### LINKED ISSUE #763: Import avalanche as avl
Currently, avalanche does not support import shortcuts like 
```python
import avalanche as avl
ewc = avl.training.strategies.EWC()
```
results in `module avalanche as no attribute training`.

This forces to explicitly import all the needed modules, which I think is often annoying.

Reformatting the `__init__` files should suffice to implement the shortcut.

AntonioCarta:
The PR brings three changes:
- a configuration file to set the default download directory for datasets.
- a new `Clock` plugin, in charge of updating several event counters.
- metrics are changed to use train_iterations, which makes their plot easier to read.

Closes #763.
Closes #756.
Closes #742.


AntonioCarta:
@AndreaCossu can you help me? I updated the pickle files with the target metrics and the tests are correct on my machine.

AndreaCossu:
This is strange, are you sure the pickle files are the same as the ones on your machine? I can't think of anything else right now. I can try to clone your repo and look around.

AntonioCarta:
I tried to update them again but it didn't solve the issue.

AndreaCossu:
I cloned your repo but the tests are failing also there. I updated the metrics files and re-run and they are completing.

PR #772, opened 2021-10-07
Correct training mode error in _periodic_eval

ashok-arjun:
**How I found this bug**:

In my `model_adaptation()` function, I was freezing the batch norms of the model. 
But the batch norm running mean and variances kept changing even then. This caused large changes for high-dimensional datasets.

Then I found that the current `_periodic_eval` function puts everything to `train` mode after evaluating, which is causing this.

Ideally, we would like the params to be restored to the original modes. 

**Fix:**

So, I have deepcopied the previous model's params, and restored them later.

Deepcopy - because the mode will change inside `eval`


AntonioCarta:
Is the deepcopy really necessary? Shouldn't `module.training` be a simple boolean?

PR #781, opened 2021-10-20
NEW notebooks

AntonioCarta:


AntonioCarta:
I'm not sure about the CI, [it looks like](https://github.com/ContinualAI/avalanche/blob/master/.github/workflows/notebook-deploy.yml) it's already deploying the entire `notebooks` folder. We'll see.

PR #798, opened 2021-11-03
FIX config type in WandBLogger by accepting a plain dictionary as input

hamedhemati:
WandBLogger now accepts as input a dictionary which can be helpful for cases where the hyperparameters are loaded as a dictionary from a file.


AndreaCossu:
Thanks! Is this backward compatible with the `args.namespace` previously accepted as input? If so, I'll merge this.

hamedhemati:
> Thanks! Is this backward compatible with the `args.namespace` previously accepted as input? If so, I'll merge this.

As also shown in `examples/wandb_logger.py`, it works with `args.namespace` if you convert it to a dictionary using `vars(.)` when feeding it to the logger. Another solution would be checking the config type inside the logger and converting it to a dict correspondingly when needed. But expecting the user to provide a dictionary shouldn't cause a problem in general.

AndreaCossu:
Agreed

PR #799, opened 2021-11-03
Add support for gym.Env experiences  

NickLucche:
This change is aimed at supporting [Avalanche-RL](https://github.com/ContinualAI/avalanche-rl/) fork by including `gym.Env` as a possible data type to be contained within experiences.
It enables Avalanche RL to utilize Avalanche benchmark module as a direct dependency without touching its source code.

I should also mention that OpenAI `gym` would now become a required dependency for Avalanche as well.

I'm very open to discuss this kind of change and evaluate any alternative options that I may overlooked in order to support benchmarks with `gym.Env` experiences .

### DIFF
--- a/avalanche/benchmarks/scenarios/generic_cl_scenario.py
+++ b/avalanche/benchmarks/scenarios/generic_cl_scenario.py
@@ -6,7 +6,12 @@ from typing import Generic, TypeVar, Union, Sequence, Callable, Optional, \
 
 import warnings
 from torch.utils.data.dataset import Dataset
-from gym import Env
+try:
+    from gym import Env
+except ImportError:
+    # empty class to make sure everything below works without changes
+    class Env:
+        pass
 
 from avalanche.benchmarks.scenarios.generic_definitions import \
     TExperience, ScenarioStream, TScenarioStream, Experience, TScenario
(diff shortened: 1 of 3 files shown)


AndreaCossu:
PR looks good to me. 
We can run again tests after adding the `gym` dependency manually so that the container for the test will have it bundled.

AntonioCarta:
Thanks @NickLucche!

Is it possible to make `gym` an optional dependency for the moment? Something like:

```
try:
    from gym import Env
except ImportError:
    # empty class to make sure everything below works without changes
    class Env:
        pass
```

Probably we need to figure out how to simplify the types. I believe this: 
```
RLStreamDataOrigin = Union[Env, Sequence[Env]]
TStreamDataOrigin = Union[AvalancheDataset, Sequence[AvalancheDataset],
                          Tuple[Iterable[AvalancheDataset], int]]
                          Tuple[Iterable[AvalancheDataset], int], 
                          RLStreamDataOrigin]
```
will necessarily lead to a huge number of corner cases.

However, this is more of a future issue to discuss after the refactoring of `avalanche-rl` will be completed.

NickLucche:
> Is it possible to make gym an optional dependency for the moment? 

Yeah totally, I'll add your lines to import gym optionally.
Thanks for the feedback!

PR #812, opened 2021-11-08
Fix setup.py problems: change outdated url, add modules to install_requires list

trenta3:
The PR fixes a couple of problems in the `setup.py` file.

*First* the url to the repository was outdated, pointing at github.com/vlomonaco/avalanche which is currently extremely behind master branch.

*Second* it changes the `python_requires` to include the newly released [Python 3.10 version](https://www.python.org/downloads/release/python-3100/).

*Third* it adds to the list of `install_requires` many modules that must be installed in order to `import avalanche`, under penalty of ModuleNotFoundError.
Those are `torch`, `torchvision`, `gdown`.

As a sidenote the list of currently installed modules is huge, and I would consider setting those as [extras](http://peak.telecommunity.com/DevCenter/setuptools#declaring-extras-optional-features-with-their-own-dependencies) and import them only in code that actually uses them using a standard try/catch clause to warn the user that a certain feature is only available when installing additional dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setuptools.setup(
                 "Research",
     long_description=long_description,
     long_description_content_type="text/markdown",
-    url="https://github.com/vlomonaco/avalanche",
+    url="https://github.com/ContinualAI/avalanche",
     packages=setuptools.find_packages(),
     classifiers=[
         "Programming Language :: Python :: 3",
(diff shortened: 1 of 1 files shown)


AntonioCarta:
Before adding Python 3.10 we should add the action to run the experiments on Python 3.10. @ggraffieti can you help?

Regarding the optional dependencies, notice that most of the optional dependencies are really small. The large ones (torch, matplotlib, scikit-learn) are probably already installed in any environment.

AndreaCossu:
I will take care of updating the actions to test py3.10

AndreaCossu:
Pytorch stable still does not support python 3.10. We will have to wait for them to provide precompiled packages. We can work with python < 3.10 for now. 
@trenta3 can you please update the setup so that we can merge this PR? :smile:

PR #813, opened 2021-11-08
Fix import error in Python 3.9

trenta3:
Under python 3.9 just importing the module causes an error "Protocols cannot be initialized".
The given error is found inside typing_extensions and warns the user about an invalid use of [protocols](https://www.python.org/dev/peps/pep-0544/): instantiation.

https://github.com/python/typing/blob/68e7cb7c90a1932e04fde6808a3ca9d3818fd64c/typing_extensions/src_py3/typing_extensions.py#L1157-L1159

It is clear that instantiating an object of a Protocol class bears no meaning since the protocol only details abstract specificiations and thus cannot be instantiated and this is enforced by typing_extensions when building a generic protocol class:

https://github.com/python/typing/blob/68e7cb7c90a1932e04fde6808a3ca9d3818fd64c/typing_extensions/src_py3/typing_extensions.py#L1233-L1251

To better understand the issue consider as example the class Accuracy in `avalanche/evaluation/metrics/accuracy.py` which inherits from Metric[float] defined in `avalanche/evaluation/metric_definitions.py`, itself inheriting from Protocol[T] where T is a typevariable.
https://github.com/ContinualAI/avalanche/blob/master/avalanche/evaluation/metrics/accuracy.py#L22-L52
Clearly line 47 `super().__init__()` is uncorrect in this setting, since Metric[float] being a protocol there is no initialization to be done.

This PR solves the issue by removing all such `super().__init__()` lines from classes whose MRO would call a Protocol `__init__` method. Be warned that such lines have been found by grepping sources so that the list may be incomplete, but at least tests do pass and the module can be imported.

### DIFF
--- a/avalanche/benchmarks/utils/avalanche_dataset.py
+++ b/avalanche/benchmarks/utils/avalanche_dataset.py
@@ -157,8 +157,6 @@ class AvalancheDataset(IDatasetWithTargets[T_co, TTargetType], Dataset[T_co]):
             the value of the second element returned by `__getitem__`.
             The adapter is used to adapt the values of the targets field only.
         """
-        super().__init__()
-
         if transform_groups is not None and (
                 transform is not None or target_transform is not None):
             raise ValueError('transform_groups can\'t be used with transform'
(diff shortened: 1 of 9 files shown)


AntonioCarta:
Thanks Dario!

PR #829, opened 2021-11-15
Fux bugs in iCaRL and Joint Training strategies

### LINKED ISSUE #807: iCaRL does not work with a different number of classes in each experience
If the experiences of the benchmark do not contain the same number of classes iCaRL do not work, since the offset used to compute the new classes are wrong. 

I've already fixed this issue in my code, I need to clean something up and I'll open a PR.

### LINKED ISSUE #828: Joint training strategy cannot be evaluated after each epoch
Joint training does not have the parameter `eval_every`, so it is evaluated only after the training, and not after every epoch.

I already fixed this in my local version, I'll open a PR in the next few days alongside the fix to #807

ggraffieti:
- iCaRL now can work with benchmarks with a different number of classes in each (or some) experiences. 
- The Joint Training strategy can now be evaluated every epoch (or ever k epochs with k > 1). 

This PR fixes #807 and fixes #828

### DIFF
--- a/avalanche/training/strategies/icarl.py
+++ b/avalanche/training/strategies/icarl.py
@@ -130,9 +130,11 @@ class _ICaRLPlugin(StrategyPlugin):
         tid = strategy.clock.train_exp_counter
         benchmark = strategy.experience.benchmark
         nb_cl = benchmark.n_classes_per_exp[tid]
+        previous_seen_classes = sum(benchmark.n_classes_per_exp[:tid])
 
         self.observed_classes.extend(
-            benchmark.classes_order[tid * nb_cl:(tid + 1) * nb_cl])
+            benchmark.classes_order[previous_seen_classes:
+                                    previous_seen_classes + nb_cl])
 
     def before_forward(self, strategy: 'BaseStrategy', **kwargs):
         if self.input_size is None:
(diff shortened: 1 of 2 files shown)


avalanche/training/strategies/icarl.py
```diff
@@ -187,19 +189,21 @@ def compute_class_means(self, strategy):
...
 
         if self.fixed_memory:
             nb_protos_cl = int(ceil(
                 self.memory_size / len(self.observed_classes)))
         else:
             nb_protos_cl = self.memory_size
-        new_classes = self.observed_classes[tid * nb_cl[tid]:
-                                            (tid + 1) * nb_cl[tid]]
+        new_classes = self.observed_classes[previous_seen_classes:
+                                            previous_seen_classes + nb_cl]
```
AntonioCarta, line 202: new classes may not be contiguous in Avalanche

avalanche/training/strategies/icarl.py
```diff
@@ -130,9 +130,11 @@ def before_training_exp(self, strategy: 'BaseStrategy', **kwargs):
         tid = strategy.clock.train_exp_counter
         benchmark = strategy.experience.benchmark
         nb_cl = benchmark.n_classes_per_exp[tid]
+        previous_seen_classes = sum(benchmark.n_classes_per_exp[:tid])
```
AntonioCarta, line 133: shouldn't this be a set?

avalanche/training/strategies/icarl.py
AntonioCarta: I think class labels should be treated as sets since Avalanche benchmarks may have unordered labels.

AntonioCarta (changes requested):

avalanche/training/strategies/icarl.py
ggraffieti: What do you mean by "new classes may not be contiguous in avalanche"? The iCaRL plugin in avalanche build the `self.observed_classes` field by inserting the classes from `benchmark.class_order`. If you mean that this does not work in a scenario which is not class incremental I agree, but we should reformat all the strategy, since by now iCaRL only works with class-incremental plugins.

avalanche/training/strategies/icarl.py
ggraffieti: no, it's just a counter that counts how many classes I've already observed. Maybe the name is misleading, I can change it.

avalanche/training/strategies/icarl.py
AntonioCarta: Does the strategy works if you have Exp.1 with classes [2, 3], and Exp.2 with classes [0, 1]? If so, you can ignore my comment.

avalanche/training/strategies/icarl.py
AntonioCarta: Does the strategy works if you have Exp.1 with classes [2, 3], and Exp.2 with classes [0, 1]? If so, you can ignore my comment.

avalanche/training/strategies/icarl.py
ggraffieti: The behavior of the strategy is the same as before, I just "generalize" the behavior for working with benchmarks with different number of classes per experience.

ggraffieti: > Does the strategy works if you have Exp.1 with classes [2, 3], and Exp.2 with classes [0, 1]? If so, you can ignore my comment.

I guess, yes, as long as the benchmark has the class order set to [2,3,1,0]

PR #834, opened 2021-11-19
Add force_data_batch_size to ReplayPlugin for manual assignment of data-mem ratio

hamedhemati:


AntonioCarta:
I see that `GroupBalancedDataLoader` and `ReplayDataLoader` have different behaviors regarding the `batch_size`. The first one creates `num_groups` dataloaders of  `batch_size` samples, obtaining a final mini-batch of size `num_groups * batch_size`. The second one uses `batch_size` as the final size. I think we should make the consistent. Either `GroupBalancedDataLoader` uses `batch_size` as the total batch size, or `ReplayDataLoader` creates a minibatch of size `2 * batch_size`. Maybe the first option is slightly more intuitive.

What do you think?

hamedhemati:
> What do you think?

@AntonioCarta  I agree with you. I think keeping the batch size equal to a specified `batch_size` defined by the user makes it more intuitive because that's what one would normally expect (fixed instead of growing). Therefore, I think it's better to set the batch size in `GroupBalancedDataLoader` equal to `batch_size` / `num_groups` for each group; and obviously it should raise an error when `batch_size` < `num_groups`.

hamedhemati:
Changes are applied to `GroupBalancedDataloader` to have a fixed batch size. It just checks whether `batch_size > len(datasets)` to ensure that each dataset gets at least one slot.

AntonioCarta:
Some tests are failing, probably because of the changed behavior in the batch size.

hamedhemati:
> Some tests are failing, probably because of the changed behavior in the batch size.

@AntonioCarta I can see that there is an `Assertion Error` caused by `assert batch_size >= len(datasets)`. So for cases where `batch_size` is smaller than the number of experience, it fails as expected. One solution would be setting the `batch_size` to `min(len(dataset), batch_size)`. Or we could randomly sample `batch_size` items from all experiences to make a fixed batch size when number of experiences is larger than the batch size. What do you think?

AntonioCarta:
I don't think we need to change the `DataLoader`. The tests are the ones that need to be fixed. They were written with the old behavior in mind.
For example, whenever there is `batch_size=1` (which was ok before, but does not make sense now), it should probably be `batch_size=len(datasets)`.

AntonioCarta:
@hamedhemati I fixed the test that was not passing so that I can merge the PR.

PR #844, opened 2021-12-05
Fixing issue#771: concat_datasets_sequentially discards classes

### LINKED ISSUE #771: concat_datasets_sequentially discards classes
When using `concat_datasets_sequentially` the resulting datasets have less classes than the union of the classes in the input dataset

```python
import torch
from torch.utils.data import TensorDataset
from avalanche.benchmarks.utils import concat_datasets_sequentially, AvalancheDataset

#  4 training datasets, 4 test datasets with targets [0,1], [2,3], [4,5], [0,1] respectively. 
#  The final output should remap classes in [0-7]

train = [AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(2, 4, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(4, 6, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,))))]

test = [AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(2, 4, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(4, 6, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,))))]

final_train, final_test, classes = concat_datasets_sequentially(train, test)

# this is ok -> [[0, 1], [2, 3], [4, 5], [6, 7]]
print(classes) 

# THIS IS WRONG -> {0, 1}, expected {0, 1, 2, 3, 4, 5, 6, 7}
print(set([el[1] for el in final_train]))  
```

ashok-arjun:
Fixes #771 raised by @AndreaCossu 

The problem was that `class_mapping` was not correctly created before to adhere to `new_class_id = class_mapping[original_class_id]`.

So a list of size largest class index is created, only the classes present here are filled with the remapped IDs.

**Test**:

```python
import torch
from torch.utils.data import TensorDataset
from avalanche.benchmarks.utils import concat_datasets_sequentially, AvalancheDataset

#  4 training datasets, 4 test datasets with targets [0,1], [2,3], [4,5], [0,1] respectively. 
#  The final output should remap classes in [0-7]

train = [AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(2, 4, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(4, 6, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,))))]

test = [AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(2, 4, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(4, 6, (20,)))), 
            AvalancheDataset(TensorDataset(torch.randn(20,10), torch.randint(0, 2, (20,))))]

final_train, final_test, classes = concat_datasets_sequentially(train, test)

# this is ok -> [[0, 1], [2, 3], [4, 5], [6, 7]]
print("Classes: ", classes) 

# THIS IS WRONG -> {0, 1}, expected {0, 1, 2, 3, 4, 5, 6, 7}
print("Final Train: ", set([el[1] for el in final_train]))  
```

**Output**:

```
Classes:  [[0, 1], [2, 3], [4, 5], [6, 7]]
Final Train:  {0, 1, 2, 3, 4, 5, 6, 7}
```


AndreaCossu:
@lrzpellegrini can you check? seems ok to me!

lrzpellegrini:
Hi there, line 

`dataset_classes = set([el[1] for el in train_set])`

is a bottleneck and the performance hit may be huge for big datasets. Use the `dataset.targets` field instead.

Also, consider including a regression unit test to check that everything is in order!

ashok-arjun:
> Hi there, line
> 
> `dataset_classes = set([el[1] for el in train_set])`
> 
> is a bottleneck and the performance hit may be huge for big datasets. Use the `dataset.targets` field instead.
> 
> Also, consider including a regression unit test to check that everything is in order!

Thanks for the review!

I'm not sure how to add a regression test - the function is specifically meant for classification (as the docstring also suggests).

AndreaCossu:
I think a simple unit test checking the correct presence of the labels in the final dataset and the total number of patterns may be enough. You can add it inside [this test file](https://github.com/ContinualAI/avalanche/blob/master/tests/test_avalanche_dataset.py). You can create a fake dataset and check that the `concat_datasets_sequentially` returns a correct dataset.

ashok-arjun:
@lrzpellegrini @AndreaCossu  Regarding `dataset.targets`, it gives a list of tensors containing the class of each example in the dataset. Here, we would like the unique classes from the dataset, hence we need to convert it to a set. But converting `dataset.targets` directly to a set is not working.

Illustration from the example used above:

```python
print(type(dataset.targets))
print(dataset.targets)
```

```
<class 'list'>
[tensor(0), tensor(1), tensor(1), tensor(1), tensor(0), tensor(1), tensor(0), tensor(1), tensor(1), tensor(0), tensor(0), tensor(1), tensor(1), tensor(1), tensor(0), tensor(1), tensor(0), tensor(0), tensor(0), tensor(0)]
```

```python
print(type(set(dataset.targets)))
print(set(dataset.targets))
```

```
<class 'set'>
{tensor(0), tensor(1), tensor(0), tensor(0), tensor(0), tensor(0), tensor(0), tensor(1), tensor(1), tensor(0), tensor(0), tensor(1), tensor(0), tensor(1), tensor(0), tensor(1), tensor(0), tensor(0), tensor(0), tensor(1)}
```

Both of the above return wrong results and errors.

**On the other hand, mapping the `targets` list to `.item()` and then performing the same operation returns a perfect result**

```python
classes = set(map(lambda x: x.item(), dataset.targets))
print(type(classes))
print(classes)
```

```
<class 'set'>
{0, 1}
```

lrzpellegrini:
Hi @ashok-arjun. Converting using `.item()` works fine when using PyTorch Tensors, but it doesn't work with Python lists.

I recommend using:

```python
set(map(int, dataset.targets))
```

which should work fine in both cases.

Alas, there is no better way to achieve this.

ashok-arjun:
@lrzpellegrini @AndreaCossu I've added the tests.

ashok-arjun:
@AndreaCossu @lrzpellegrini Can you please tell me how to resolve these two errors?

AndreaCossu:
Thanks @ashok-arjun ! I adjusted the line indentations highlighted in the comment above. It seems good now, we can merge.

ashok-arjun:
@AndreaCossu Thanks a ton for the help! :))

PR #875, opened 2022-01-09
Add Experience-Change Logging to WandBLogger

hamedhemati:
This PR adds experience-change logging to the WandB logger by explicitly setting the step in `wandb.log(.)`.   This can be useful to have more meaningful visualizations in some of the plots.

Example:
Screenshot 2022-01-09 at 18 28 44

is mapped to:
Screenshot 2022-01-09 at 18 28 08
)

### DIFF
--- a/avalanche/logging/wandb_logger.py
+++ b/avalanche/logging/wandb_logger.py
@@ -28,7 +28,6 @@ from matplotlib.pyplot import Figure
 from avalanche.evaluation.metric_results import AlternativeValues, \
     MetricValue, TensorImage
 from avalanche.logging import StrategyLogger
-from avalanche.training.strategies import BaseStrategy
 
 
 class WandBLogger(StrategyLogger):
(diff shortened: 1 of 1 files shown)


AndreaCossu:
Hi @hamedhemati ! I haven't tried your changes, yet. If I got this correctly, the user will be able to modify the x-axis interactively on W&B platform by either leaving it as default (minibatch ID) or by selecting experience ID, right?

hamedhemati:
> Hi @hamedhemati ! I haven't tried your changes, yet. If I got this correctly, the user will be able to modify the x-axis interactively on W&B platform by either leaving it as default (minibatch ID) or by selecting experience ID, right?

Hi @AndreaCossu. Yes exactly. By default, the x-axis in all plots remains as "step" (which is the mini-batch increments) but for plots based on streams (eg. acc/loss for test streams), you can change it to experience ID. I use this very often, so I thought it might be useful to merge.

AndreaCossu:
Great, thanks :smile: I think it can be useful, especially for long streams. Can you just change the name to `TrainingExperience` instead of `Experience`? In this way we avoid confusion with test experiences ID.

hamedhemati:
> 

Yea sure, done :)

PR #880, opened 2022-01-18
Initialize BaseOnlineStrategy

HamedHemati:
This PR initializes BaseOnlineStrategy. As discussed before, this is meant to be considered as a possible solution(and not necessarily the final soliton) for the implementation of online strategies. Three new files are added:

`avalanche/training/strategies/base_online_strategy` -> the base strategy for online strategies
`avalanche/training/strategies/strategy_wrappers_online.py` -> tried to avoid mixing strategy wrappers
`examples/online_naive` -> an example for testing

### DIFF
--- a/avalanche/training/strategies/base_online_strategy.py
+++ b/avalanche/training/strategies/base_online_strategy.py
@@ -110,8 +110,8 @@ class OnlineStrategy(BaseStrategy):
         return res
 
     def train_sub_exp(self, experience: Experience, eval_streams=None,
-                  is_first_sub_exp=False, is_last_sub_exp=False,
-                  **kwargs):
+                      is_first_sub_exp=False, is_last_sub_exp=False,
+                      **kwargs):
         """ Training loop over a single Experience object.
 
         :param experience: CL experience information.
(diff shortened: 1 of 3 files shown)


AntonioCarta:
Thanks @hamedhemati!

As we discussed at the previous meeting, there are many issues with this approach that need to be addressed. I think we can accept this PR to use it as a starting point. I will open a separate issue to provide a roadmap for the next steps.

avalanche/training/strategies/base_online_strategy.py
```diff
@@ -0,0 +1,173 @@
...
+                 train_mb_size: int = 1,
+                 eval_mb_size: int = None, device=None,
+                 plugins: Optional[List[StrategyPlugin]] = None,
+                 evaluator: EvaluationPlugin = default_logger, eval_every=-1):
+        super().__init__(
+            model, optimizer, criterion,
+            train_mb_size=train_mb_size, train_epochs=1,
+            eval_mb_size=eval_mb_size, device=device, plugins=plugins,
+            evaluator=evaluator, eval_every=eval_every)
+
```
AntonioCarta, line 28: can you add a warning inside the constructor?
Something like: "This is an unstable experimental strategy. You can check https://github.com/ContinualAI/avalanche/blob/master/examples/cope.py for an alternative approach to online CL (data incremental scenario)."

PR #902, opened 2022-02-04
New base strategy templates

AntonioCarta:
This is the PR I presented you at the last meeting. It adds:
- the train/eval logic, as well as the callbacks, are now implemented by `Templates`
- we can have multiple templates. This PR provides a preliminary organization
- each templates has a corresponding `Plugin`, which implements its supported callbacks
- There is a smarter support for callbacks, which effectively allows reuse between templates. Basically, as long as a `Plugin` implements only supported callbacks, we assume it's ok to run. This is checked dynamically and the user is warned in case of problems.
- `EvaluationPlugin` and `Logger`s have been updated. Most loggers now work for any `Template`, but they can also be extended for more customized behavior.

It's a big PR but should mostly change the internal. I renamed the old `BaseStrategy` and `StrategyPlugin` into `SupervisedTemplate` and `SupervisedPlugin`.


vlomonaco:
Good job! :) When this is merged let's open an issue to remember to reflect this change in the api-doc and tutorials!

PR #905, opened 2022-02-06
Update ReplayPlugin and ReplayDataLoader

HamedHemati:
This PR adds new options to  ReplayDataLoader and ReplayPlugin that enable custom data-memory ratios.

### DIFF
--- a/avalanche/benchmarks/utils/data_loader.py
+++ b/avalanche/benchmarks/utils/data_loader.py
@@ -244,17 +244,14 @@ class GroupBalancedInfiniteDataLoader:
 class ReplayDataLoader:
     """Custom data loader for rehearsal/replay strategies."""
 
-    def __init__(
-        self,
-        data: AvalancheDataset,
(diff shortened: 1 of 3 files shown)


HamedHemati:
The tests only failed for Python 3.6 this time, with the assertion error below:

```
TRAIN STREAM ACC:  {0: 0.865546218487395, 1: 0.8264462809917356, 2: 0.7815126050420168, 3: 0.8571428571428571, 4: 0.6885245901639344}
Traceback (most recent call last):
  File "/__w/avalanche/avalanche/tests/training/test_strategies_accuracy.py", line 92, in test_multihead_cumulative
    > 0.7
AssertionError
```

That's a bit weird that it only happened in 3.6...

AntonioCarta:
that's a non-deterministic test. It makes sense that it doesn't always fails. I believe this may be a failure of my previous changes on the `BaseStrategy`, which did not fail the first time. I will check in the next few days.

PR #919, opened 2022-02-22
Improve plugin compatibility checks

tachyonicClock:


tachyonicClock:
I ran into an issue which is non-trivial to fix where SupervisedTemplate (and its associated plugin class) wrongly inherits from BaseSGDTemplate. Some strategies such as streaming linear discriminant analysis (SLDA) are supervised but are not based on SGD. Previously all compatibility checks were performed with the same PLUGIN_CLASS resulting in the warning falsely occurring in most cases. Thus my fix to improve plugin compatibility checking resulted in streaming linear discriminant analysis erroneously passing.

AntonioCarta:
Your fix seems correct. The problem here is that SLDA should not inherit from supervised but from `BaseTemplate` since it does not have backpropagation (it was implemented before we had multiple templates). 

If you want, you can refactor SLDA to inherit from the `BaseTemplate`. Otherwise, you can comment the tests and we can open a separate issue to fix SLDA later.

tachyonicClock:
Continual learning strategies are diverse and as such not all plugin are compatible with all strategies. Therefore the `Plugin` class tracks interfaces that a plugin requires. At runtime the plugin checks this feature set to see if any unsupported features exist. When creating a plugin simply inherit from the plugin features you need. 

      class MyPlugin(TrainingEvents, RequiresSGD):
          pass

As an advanced feature you may register a plugin as requiring implementation by a continual learning strategy it should inherit from `Plugin` and be decorated with `@Plugin.requires_compatibility`. 

      @Plugin.requires_compatibility
      class MyPlugin(Plugin):

          def new_callback():
              pass

As an example a class that inherits from `SupervisedPlugin` has the following features
```
{TrainingEvents, EvalEvents, FittingEvents, RequiresSGD, DatasetAdaptationEvents}
```

AntonioCarta:
I think your solution is going in the correct direction but I'm a bit confused by some of the nomenclature.

- why do we need a `Plugin.requires_compatibility`? I think we want to always check the compatibility.
- `RequiresSGD` should be called `SGDEvents`.
- what are features and why do you call them `features`? Are they not events/callbacks? If I understand correctly, what matters here is not the "set of classes" but their interface (the union of the events/callbacks they are aware of). I think we should keep the nomenclature consistent. Calling them events is also fine, but calling them features is confusing.


> As an advanced feature you may register a plugin as requiring 
    implementation by a continual learning strategy it should inherit from
    `Plugin` and be decorated with `@Plugin.requires_compatibility`. 

Can you rephrase this? What do you mean? strategies do not implement plugins.

AntonioCarta:
> Plugin.requires_compatibility is used as a way to note that a Plugin requires a strategy to explicitly support it.
Plugin.requires_compatibility -> Plugin.make_required_feature

> I tried to make what I mean by feature more explicit primarily be defining the type FeatureSet = Set[Type['Plugin']] which is a set of classes (inheriting from Plugin) that are used to communicate the requirements of a plugin to a strategy.

I believe there is an easier way to do this with [generics](https://mypy.readthedocs.io/en/stable/generics.html). There are two things that the plugin requires to be supported by a strategy:
1 - the required event must be triggered
2 - the required state attributes must be present

The master branch currently checks (1) but not (2).
IMO, for (2) the easiest way is make the plugin type a generic type, i.e., `Plugin[Template]`, where `Template` is the type of the `strategy` argument received by the plugin's methods. So, if you want a `BasePlugin` that needs SGD, you would have a type `BasePlugin[SGDTemplate]`. 

Using generics, we don't need `Plugin.requires_compatibility` or marker interfaces such as `RequiresSGD`. The advantage of this solution is that it's very easy to choose the correct type and to understand what kind of interface the type is supporting. Instead, `RequiresSGD` is just a marker, it's not clear to me what is supposed to represent. Do RL strategies implement `RequiresSGD`? Maybe, maybe not, depends on some hidden assumptions. With generics, RL strategies implement the required interface if they subclass `SGDTemplate`.

> I split BaseSGDPlugin into RequireSGD and FittingEvents because some strategies are not SGD but still hook into the FittingEvents.

Can you give me some examples of this?

tachyonicClock:
I closed this. I might revisit it at some point but the approach I tried to sell seems to be a bit too confusing/complex.

PR #920, opened 2022-02-24
Detection template

AntonioCarta:
implementation of the object detection template. This is a refactoring of `train_one_epoch` and `lvis_evaluate` which uses Avalanche templates. Metric/logging logic is still missing (notice that this is implemented is several places in the original functions).

I put it in `detection_challenge.py` for the moment, to keep the old file for reference. We can delete it once the refactor is completed.


lrzpellegrini:
Wow, it's great! Are the changes in `avalanche/training/templates/base.py` and `avalanche/training/templates/base_sgd.py` a porting from the master branch? Or are they just needed for detection?

AntonioCarta:
They are mostly needed for the detection. They are simple additional extension points to allow to customize things like the optimizer step without rewriting the entire iteration code.

PR #921, opened 2022-02-24
Added model output export to JSON in example

lrzpellegrini:
Added, in `examples/detection.py`, a simple mechanism to import/export model outputs from/to a JSON file. Added a CLI option to just compute metrics starting from model outputs loaded from such JSON files.

@AndreaCossu, it's quite simple, but it does the job.

### DIFF
--- a/examples/detection.py
+++ b/examples/detection.py
@@ -12,25 +12,32 @@
 """
 This is a simple example on how to run detection benchmarks.
 """
-
+import json
 import logging
 
 # This sets the root logger to write to stdout (your console).
 # Your script/app needs to call this somewhere at least once.
+from json import JSONEncoder
+
+from torch import Tensor
+
 from avalanche.benchmarks.datasets.lvis.lvis import LvisDataset
+from examples.tvdetection.coco_eval import CocoEvaluator
+from examples.tvdetection.lvis_eval import LvisEvaluator
+from lvis_api import LVISEval, LVIS
 
 logging.basicConfig(level=logging.NOTSET)
 
 import matplotlib
+matplotlib.use('Agg')
 
 from avalanche.benchmarks.utils import AvalancheDataset, AvalancheSubset
-from examples.tvdetection.engine import train_one_epoch, evaluate
-
-matplotlib.use('Agg')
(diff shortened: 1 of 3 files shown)


AndreaCossu:
Thanks! So, to better understand what kind of help you need:
I can implement a metric working at eval time which after each minibatch takes the model output and updates the LVISevaluator object. At the end of the evaluation stream the metric will serialize the model outputs to a JSON file and will return to the loggers the final metric values (the ones you computed in `verify_output_only`).
Otherwise, if you prefer, I can implement only the metric gathering the model outputs and saving them to JSON. A separate script will be in charge of taking the JSON file and printing the final metrics.

What about training time instead? I can do the same, but let me know if there are any differences.

lrzpellegrini:
- For the competition, we'll surely need the metric in charge of exporting the model outputs. I don't think we need to store the model outputs for training experiences. Just beware that we'll need a way to store the model outputs separately for each experience (to compute the mean accuracy/AP/... over time). I think that creating files like output0.json, output1.json, ... in a configurable folder would be the most general solution. Consider that this metric/plugin can be a valuable tool to be used in the context of other competitions as well.
- I think that creating an Avalanche metric in charge of computing detection metrics (using the LVIS API) may be useful as well. Even if the test set labels will not be disclosed, participants may create a validation set to check the model performance using this metric.

Apart from this, do you think that we can make the output export plugin able to automatically detect if a classification or a detection task is being executed? (and so store outputs in the most appropriate way)

AntonioCarta:
> do you think that we can make the output export plugin able to automatically detect if a classification or a detection task is being executed? (and so store outputs in the most appropriate way)

Do we really need this? I think it's easier for the users if we provide separate main scripts for each track. Therefore, we can use different loggers for each.

AndreaCossu:
For avalanche, it may be useful to have a single metric so that the user does not need to change it depending on the task. The metric will try to save the model output based on its type rather than the task. If the output is something very strange we can raise an error/warning. 
Otherwise we may have to build `ClassificationModelOutput`, `DetectionModelOutput`, `SegmentationModelOutput` and so on.

lrzpellegrini:
Closing this, as metrics were implemented in #922

PR #922, opened 2022-02-25
Draft for LVIS detection metrics

AndreaCossu:
The metric implements two functionalities:

1. (optional) Saving model outputs to JSON file. The metric produces one file for each eval experience, where each file stores a list of dictionaries. Each dictionary is a model output.
2. Computation of metrics through LVISEvaluator object starting from model outputs. The metric returns a string. This point does not need 1. to work.

The code contains portions of code from #921 , adapted to fit the metric interface.

I am assuming to be able to access model outputs and targets through the usual `strategy.mb_output` and `strategy.mb_y` fields.

I just implemented the metric but I have not tested it in any ways.

### DIFF
--- a/avalanche/evaluation/metric_results.py
+++ b/avalanche/evaluation/metric_results.py
@@ -29,7 +29,7 @@ class TensorImage:
         return self.image.numpy()
 
 
-MetricType = Union[float, int, Tensor, Image, TensorImage, Figure]
+MetricType = Union[float, int, str, Tensor, Image, TensorImage, Figure]
 
 
 class AlternativeValues:
--- a/avalanche/evaluation/metrics/__init__.py
+++ b/avalanche/evaluation/metrics/__init__.py
@@ -14,3 +14,4 @@ from .timing import *
 from .mean_scores import *
 from .labels_repartition import *
 from .images_samples import *
+from .detection import *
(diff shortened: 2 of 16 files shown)


lrzpellegrini:
I think that the last PR from @AntonioCarta switched the reference LVIS class from lvis_api.LVIS (our copy of the LVIS API) back to lvis.LVIS (as in pip-installed lvis). It seems that we don't need to customize the LVIS API, so we shoulld stick with lvis.LVIS from pip. In any case, removing that assertion seems reasonable.

One thing: looking at the `LvisMetrics` code it seems that `LvisEvaluator` doesn't get reset after/before each experience. I think it's reasonable to reset everything before each eval exp.

AndreaCossu:
Thanks @lrzpellegrini , I now create the `LvisEvaluator` from scratch before each eval experience.
Also, I didn't mean to merge this PR, I don't know exactly what happened (I simply pushed the changes), but since we are in a separate branch I guess it's ok. 

I am still getting type errors on the `evaluate` call after the first experience (the lvis API still expects `lvis.LVIS` type). Since we cannot modify that part of the code inside the lvis library maybe we could revert the change of our class back to `lvis.LVIS`.

lrzpellegrini:
Ok, no problem. It should be easy to move everything back to `lvis.LVIS`. I can take care of this. Thanks!

PR #933, opened 2022-03-10
Consolidating changes from the detection branch

lrzpellegrini:
This PR is a porting of the contributions accumulated in the detection branch.

In particular, the contributions (that in the detection branch were mainly implemented as examples) have now been ported into the avalanche package.

The following have been changed w.r.t. the detection branch:

- The LvisMetric will log the metrics from the Lvis API instead of creating the CodaLab score string
- The detection strategy is now a subclass of the supervised template. Few field names have been changed. The behavior of a couple of methods has been slightly changed.

Those changes are standalone and should not affect the behavior of the rest of Avalanche, with the following exceptions:
- Concerning Avalanche, in the base template, the backward and optimizer step operations are now executed in two methods (`backward` and `optimizer_step`)
- The `lvis` pip package has been added to environment.yml, environment-dev.yml, and setup.py

With this, we add support for object detection with the LVIS dataset. The plan for the next few days is to implement support for COCO+other datasets (by unifying the detection API and evaluator interfaces), and instance segmentation (including model outputs (de)serialization).

**For the CI, we'll need to re-create the docker with the new dependency (`lvis`).** For the moment, the unit tests fail due to the missing package. This is why I'm creating this as a draft.


AntonioCarta:
Thanks, everything looks good. Can we make lvis an optional dependency?

lrzpellegrini:
The last commit greatly improves the support for detection. In fact, this should conclude the minimum set of features needed to support object detection/segmentation in Avalanche. The detection support does no longer need the scripts from pytorch/vision (which I'll leave in the examples folder for the moment).

Currently, only LVIS is supported, but in the next few days I'll add COCO and PennFudan Pedestrians (the one used in the torchvision tutorial).

AndreaCossu:
man you are on fire!

lrzpellegrini:
@ggraffieti To make unit tests work, we might need to add this to the Docker: https://stackoverflow.com/questions/55313610/importerror-libgl-so-1-cannot-open-shared-object-file-no-such-file-or-directo

lrzpellegrini:
As the CLVision challenge is approaching, I'd like to merge this PR asap. Please let me know if there are any issues!

AndreaCossu:
I can't see any specific issues, I guess we will discover them as we go :laughing:

AntonioCarta:
Looks good. You can merge it when you are ready.

PR #941, opened 2022-03-21
Implementation of the top-k accuracy for Avalanche

rudysemola:
This adds all the sets of classes where it is possible to use the top-k accuracy in a similar way to the other metrics in the library.

- **TopkAccuracy** base class,
- **TopkAccuracyPluginMetric** plugin class,
- **MinibatchTopkAccuracy**, **EpochTopkAccuracy**, **RunningEpochTopkAccuracy**, **ExperienceTopkAccuracy**, **StreamTopkAccuracy**, **TrainedExperienceTopkAccuracy** classes
- **topk_acc_metrics** helper function

I used the k=3,5 for my experiments and it works.

### DIFF
--- a/avalanche/evaluation/metrics/__init__.py
+++ b/avalanche/evaluation/metrics/__init__.py
@@ -14,3 +14,4 @@ from .timing import *
 from .mean_scores import *
 from .labels_repartition import *
 from .images_samples import *
+from .topk_acc import *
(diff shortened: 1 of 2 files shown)


AndreaCossu:
Thanks @rudysemola ! The PR looks good, we just need to add `torchmetrics` as avalanche dependency, I'll take car of that. I think we can make it an optional dependency later but we still don't have a proper workflow to do that.

lrzpellegrini:
I just added the new dependencies. The test container should be generated soon (I suppose).

PR #956, opened 2022-03-29
LaMAML strategy

HamedHemati:
This PR includes an implementation of the [LaMAML](https://arxiv.org/abs/2007.13904) strategy in Avalanche. 
It is a supervised meta-learning-based strategy that exploits a memory buffer (res-sampling). I thought having an example of a strategy that has meta-updates can be helpful for others to get started.

** This implementation is different from the official implementation that works with a stateful version of the model to do inner updates (fast updates) in the meta optimization steps. I'm using [Higher](https://github.com/facebookresearch/higher) in this code, which allows you to create a stateless version of your model and keep track of the optimization trajectory in a cleaner and easier to follow way. This would allow running the same code for new models without the need to re-define the forward function. I think adding Higher to the Avalanche installation requirements can be useful for other meta-learning-based strategies as well.

The results look similar (for the original CNN model) in both implementations.


AndreaCossu:
Thanks @HamedHemati this is great! Would you mind to port your Split CIFAR-100 example to our [reproducible continual learning repo](https://github.com/ContinualAI/reproducible-continual-learning)? It should be fairly quick.
Let me know if you need help :smile:

HamedHemati:
> 
At the moment, I only have full results for an older implementation and not exactly this one (with the `MTSimpleCNN` model). This time I just waited for the first few tasks and it generated very similar results to the older one. But I will run new experiments soon (for a baseline comparison that I need) and can share that :)

HamedHemati:
Hey @AntonioCarta, the tests have failed due to the missing dependency (higher). Do we need to install it on the remote machine? 
It's a lightweight library and can be installed with `pip install higher`.

AndreaCossu:
I'll take care of it

AntonioCarta:
Thanks @HamedHemati, I think we can merge this as  soon as you have the script for the reproducibility repo ready.

HamedHemati:
> Thanks @HamedHemati, I think we can merge this as soon as you have the script for the reproducibility repo ready.

I have the results for Avalanche's MTSimpleCNN:

CIFAR-100: ~ 0.56,  in the paper: ~0.70. (20 experiences - 5-way classification)
TinyImageNet: ~0.53, in the paper: ~0.66 (40 experiences - 5-way classification)
 
I guess the difference in the setup (model architecture/seeds/..) can be the reason for the gap. 

@AntonioCarta  To add the script for the reproducibility repo, first, we need to have the strategy in Avalanche, or do you mean something else?

AntonioCarta:
Can you test the strategy using their network and hyperparameters? Otherwise it's difficult to compare the performance and catch possible bugs.

HamedHemati:
> Can you test the strategy using their network and hyperparameters? Otherwise, it's difficult to compare the performance and catch possible bugs.

Ok, here are the results for the same model (just replaced the final head with a MultiTask head instead of output masking):

CIFAR-100 (3 conv + 3 linear / no dropout or bn): 
this implementation: ~0.69,  in the paper: 0.70

TinyImageNet (4 conv + 3 linear / no dropout or bn):
this implementation: ~0.64, in the paper: 0.66

Another result that shows architecture really matters :D

HamedHemati:
I think the whole branch is merged. If the results look ok, you can ignore this PR and I will make another PR with the same two files.

AndreaCossu:
Ok, so we can close this?

HamedHemati:
> Ok, so we can close this?

I had to reset upstream, I'm going to make another PR.

PR #962, opened 2022-03-30
Add RWalk implementation and example

flandolfi:
Hi!

Here's my implementation of the Riemannian Walk (RWalk) strategy ([paper](https://arxiv.org/abs/1801.10112)). This was done as a final project for the CL PhD course at UniPi.

Few comments about the implementation:

 -  This was entirely based on what was written in the paper. I ignored other implementations (e.g., the one contained in [A-GEM](https://github.com/arslan-chaudhry/agem) by the author, or the unofficial one in [FACIL](https://github.com/mmasana/FACIL)).
 - The Fisher information matrix is computed at every iteration (i.e., every *batch*), differently from EWC or from the FACIL implementation, as "[...] is computed efficiently every *t*-step using moving average [...]" (page 8 of the paper).
 - The sum in the score *s_t1^t2* is updated every `delta_t` iteration/batch, specifically the sum from *t1* to *t2* is computed at instants *t1* + *k*`delta_t`. If *t2* != *t1* + *k*`delta_t` for any *k*, then the sum is not updated at *t2*.
 - RWalk (in the paper) makes use of sampling strategies (e.g., Replay or iCarl). I did not add these strategies as should be plugged-in using existing implementations in Avalanche.
 - The default parameters in the methods refer to the ones in the Appendix of the RWalk paper.

Running the example with the `pmnist` scenario you should obtain the following results:
```
        Loss_Stream/eval_phase/test_stream/Task000 = 0.1392
        Loss_Stream/eval_phase/test_stream/Task001 = 0.1355
        Loss_Stream/eval_phase/test_stream/Task002 = 0.1830
        Loss_Stream/eval_phase/test_stream/Task003 = 0.1453
        Loss_Stream/eval_phase/test_stream/Task004 = 0.1534
        StreamBWT/eval_phase/test_stream = -0.0097
        StreamForgetting/eval_phase/test_stream = 0.0097
        Top1_Acc_Stream/eval_phase/test_stream/Task000 = 0.9633
        Top1_Acc_Stream/eval_phase/test_stream/Task001 = 0.9589
        Top1_Acc_Stream/eval_phase/test_stream/Task002 = 0.9474
        Top1_Acc_Stream/eval_phase/test_stream/Task003 = 0.9541
        Top1_Acc_Stream/eval_phase/test_stream/Task004 = 0.9535
```

I'll stay in touch for eventual modifications or corrections!

Have a nice day,
Francesco


avalanche/training/supervised/strategy_wrappers.py
```diff
@@ -693,6 +694,79 @@ def __init__(
         )
 
 
+class RWalk(SupervisedTemplate):
```
AndreaCossu, line 697: to avoid confusion, since RWalk uses also replay, I would either 1) not create this class or 2) introduce also replay in it (it can be provided by the user as input plugin or defined by the strategy if the original paper uses a specific one already in avalanche).

examples/rwalk_mnist.py
```diff
@@ -0,0 +1,149 @@
...
+            minibatch=True, epoch=True, experience=True, stream=True
+        ),
+        loss_metrics(minibatch=True, epoch=True, experience=True, stream=True),
+        forgetting_metrics(experience=True, stream=True),
+        bwt_metrics(experience=True, stream=True),
+        loggers=[interactive_logger, tensorboard_logger],
+    )
+
+    # create strategy
+    strategy = RWalk(
```
AndreaCossu, line 75: if you remove the class, this becomes a Naive with RWalkPlugin

avalanche/training/plugins/rwalk.py
```diff
@@ -0,0 +1,206 @@
...
+
+import torch
+import torch.nn.functional as F
+
+from avalanche.training.plugins.strategy_plugin import SupervisedPlugin
+from avalanche.training.utils import copy_params_dict, zerolike_params_dict
+from avalanche.models.utils import avalanche_forward
+
+
+class RWalkPlugin(SupervisedPlugin):
```
AndreaCossu, line 11: Maybe RWalkImportancePlugin or EWC++Plugin? To clarify what it is doing.

avalanche/training/supervised/strategy_wrappers.py
flandolfi: Ok, I think I'll remove the class. There's no specified default sampling strategy, although I believe MoF (from ICARL) and simple replay are the most effective in the ablation study. I don't know actually if it suffice to use ICARL+RWalk to have the results in the paper, or if we need to separate MoF from ICARL first (I don't know how the latter works).

examples/rwalk_mnist.py
flandolfi: Thanks, maybe I'll add an option to use also a replay plugin.

avalanche/training/plugins/rwalk.py
flandolfi: In the paper they refer to RWalk as the penalty/regularization term (see the title of the third paragraph of sect. 4.1). The sampling strategies are later introduced in sect. 4.2. In any case, I'll add a warning in the docs!

AndreaCossu: thanks, that should be enough

examples/rwalk_mnist.py
AndreaCossu: If you remove the strategy I think you can just use Naive + Rwalk plugin

avalanche/training/supervised/strategy_wrappers.py
AndreaCossu: That would require further examination to understand which kind of replay is best suited to be included as default in the RWalk strategy. You can remove the class for now.

flandolfi:
Hi! I made the following changes:

 - Removed the `RWalk` strategy class;
 - Used `Naive`+`RWalkPlugin` in the example, instead;
 - Added a note in the `RWalkPlugin` docs to warn about the usage and reproducibility issue;
 - Added a Unit Test in `test_strategies.py` (even though it's not a strategy, it's only a plug-in used with `Naive`);
 - Refactored the plugin code (the weights penalties are computed once at the end of the experience and not at every iteration).

AndreaCossu:
Thanks, if @AntonioCarta agrees I think we can merge this.

AntonioCarta:
Thanks @flandolfi!

PR #965, opened 2022-04-01
Adding CLEAR-10 benchmark to Avalanche

linzhiqiu:
#964 PR for this feature request


avalanche/benchmarks/classic/clear.py
```diff
@@ -0,0 +1,234 @@
...
+        test_split = 'all'
+    elif evaluation_protocol == 'iid':
+        assert seed in SEED_LIST, "No seed for train/test split"
+        train_split = 'train'
+        test_split = 'test'
+    else:
+        raise NotImplementedError()
+    
+    if feature_type is None:
+        assert isinstance(train_transform, type(None)), "No image transform"
```
lrzpellegrini, line 141: I haven't tried to run the code, but from the main below it seems that non-None transformations should be allowed when feature_type is None...

lrzpellegrini (changes requested):
Thank you, the PR looks great! Everything seems to be in order apart from a comment I left regarding transformations.

avalanche/benchmarks/classic/clear.py
linzhiqiu: You are right, I should remove the assertion statement (it was a typo). My original purpose is to make sure the user specify a transformation for image, and not inputting a transformation for feature tensors. But on a second thought, I should leave this choice to users!

linzhiqiu:
Made the change by removing assertion statements.

AntonioCarta:
@lrzpellegrini if you are satisfied we can merge the PR.

lrzpellegrini (approved):

AntonioCarta:
Thanks again @linzhiqiu

PR #967, opened 2022-04-01
Update OnlineSupervsiedTemplate and InteractiveLogger

HamedHemati:
This PR updates the `OnlineSupervisedTemplate` and `InteractiveLogger`. As we discussed in our last meeting, all mini-experiences in online strategies now have three attributes assigned to them:
```
experience.is_online_exp
experience.online_exp_id
experience.online_exp_total
```

and these attributes are used to determine when to call model_adaptation, interactive logger updating and resetting, and potentially other conditional statement checking in the future.

I've tested it for NaiveOnline and NaiveOnline + Replay. For NaiveOnline, it works almost as expected; here are the profiler outputs:
ration for 

**Simple Online Naive Strategy (no Avalanche)**:  37.64 sec 

**Without interactive logger:**
OnlineNaiveStrategy:  84.01 sec
NaiveStrategy:  76.25

**With interactive logger:**
OnlineNaiveStrategy+IntLogger:  97.79 sec
NaiveStrategy+IntLogger:  111.2 sec

** the numbers can slightly change for different runs. We can say that the performances of Online and Batch-based strategies in Avalanche are similar.

With these changes, new strategies can be implemented with OnlineSupervisedTemplate, and the interactive logger works fine. However, the ReplayPlugin is slow, so the next step would be focusing on the performance of ReplayPlugin in online setups.


AntonioCarta:
Maybe it's better to split this into two parts. First, we have to agree on what an `OnlineExperience` is, and how to define it. The current proposal:
```
experience.is_online_exp
experience.online_exp_id
experience.online_exp_total
```
breaks or reinvents a lot of features that I think we already have in place. It's ok if you want to change them, but it's not ok to duplicate the same logic with different names.

Some comments:
```
experience.is_online_exp --> instead of an attribute, can we do isinstance(OnlineExperience)? Otherwise, we may end up with several flags in the future just to check the type of an experience.
experience.online_exp_id --> isn't this experience.current_experience (i.e. the experience identifier)
experience.online_exp_total --> shouldn't this be a property of the stream?
```

HamedHemati:
@AntonioCarta 
```
experience.is_online_exp --> instead of an attribute, can we do isinstance(OnlineExperience)? Otherwise, we may end up with several flags in the future just to check the type of an experience.

-> Then we would have to define OnlineExperience, which I think is not necessary. This just extends an experience to distinguish between the type (batch or online). I don't think we'll have other types?! it's just these two types.

experience.online_exp_id --> isn't this experience.current_experience (i.e. the experience identifier)

->not exactly, but a good idea, we can change it to : experience.current_experience 

experience.online_exp_total --> shouldn't this be a property of the stream?: 

->I don't think so, because the stream itself can have multiple experiences which are each split into individual mini-experiences. This total is related to the current actual experience, not the stream.
```

HamedHemati:
@AntonioCarta I think the solution to online CL is very simple if we agree on changing the `ReplayPlugin`. 

By setting `n_epochs=1`, the strategy will become online (even if `n_batch!=1`). However, the biggest issue here is that the buffer gets updated only after each experience. So, if we create a ReplayPlugin that can also get updated after each iteration (can be chosen by the user), and change the `collate_fn` of the main data loader in a way that it combines the current batch with the sample from the buffer, I think we are done. If you think this is not correct, please let me know.

HamedHemati:
> We already have RL, which is yet another scenario. We will add others in the future.

I think here the point is how many times we observe the data: just once or multiple times, not the problem type.

>There is no notion of mini-experiences in the final OCL stream. An experience in OCL is the single minibatch. This are generated from a larger experience, and we can use an ID to track this. It's also not clear to me what online_exp_total really is.

Maybe the naming is not ok, that's something that I also wanted to discuss. mini-experience is the previously called sub-experience that we had. I thought the word mini would be more meaningful as in online CL we have very small batches (either 1 or up to 10). So each mini experience is a part of a bigger experience that the model goes through. The `online_exp_total` I had to add only because of the interactive logger. You need to specify the total number of splits you make for the progress bar.

HamedHemati:
I hope this visualization makes things more clear:
Screenshot 2022-04-01 at 15 47 28

ggraffieti:
Sorry to interrupt the discussion about this PR, but this is quite relevant and in line with the PR itself. 
The concept of splitting an experience into many "mini-experiences" is interesting and I'm currently dealing with it in the field of online CL. However, I don't think the current nomenclature of an experience that is split into many mini-experiences is totally correct since from the Avalanche paper an experience is _a set composed of one or multiple samples that can be used to update the model. This is often referred to in the literature as batch or task._ In the case of mini-experience, the experience cannot be used to update the model, since only a mini-experience is available. 
In our recent work we "flip" the relationship, and we called _experience_ the mini-experiences in the figure posted by @HamedHemati and _macro-experience_ the original big experience. 
IMO this is more coherent with the definition of experience as a chunk of data that can be used together to update the model. In this case, the macro-experience (another possible name is _season_) is just a collection of experiences that share some common properties. 
Maybe I should post this on the discussion panel since is more a nomenclature problem than an implementation one.

vlomonaco:
Thanks @ggraffieti, we indeed discussed about this in the last avalanche meeting. In the end, the agent / strategy only sees a stream of experiences it cannot control, so there's no actual need to have a semantic distinction for the end users which remain the same discussed in the paper (@AntonioCarta correct me if I'm wrong).

PR #968, opened 2022-04-03
LaMAML Strategy

HamedHemati:
Results for the same model (just replaced the final head with a MultiTask head instead of output masking):

CIFAR-100 (3 conv + 3 linear / no dropout or bn):
this implementation: ~0.69, in the paper: 0.70

TinyImageNet (4 conv + 3 linear / no dropout or bn):
this implementation: ~0.64, in the paper: 0.66


AntonioCarta:
Thanks! dont forget to make the  PR to the reproducibility repo.

PR #977, opened 2022-04-11
Proposal: New generic scenarios & ExML scenarios

AntonioCarta:
As I told you some time ago, I did some changes to the scenarios to make them more general.

- the Experience is designed to be the only object received by CL strategies. Streams and benchmarks are accessible from the experience but the are hidden from the strategy.
- the generic scenario don't have mandatory datasets and task labels anymore. This means that scenarios like RL are easier to support.
- The old generic scenarios have become "GenericClassificationScenario". These is mostly a renaming of internal stuff, so it shouldn't break anything.
- The new API is based on generators. By default, it assumes that experiences are generated lazily. This allows for an easier (and possibly more efficient) implementation of long streams. Lists are still supported.
- added a method to hide information. Now experiences have three modalities: "train", "eval", "logging". These allows to add any attribute to the experience without worrying that users may misuse it (e.g. task labels available only at train/test/logging time). At the same time, it allows full access to any info during logging and metric computations.

To experiment with the new API, I implemented two scenarios:
- Online CL, with the attributes that were proposed by @HamedHemati in his last PR. Notice that most attributes are hidden during training and the experiences are lazily generated. Basically, this is a lazy version of the current data-incremental generator. I profiled it, and it is as fast as the naive implementation without Avalanche.
- Ex-Model Continual Learning. This come from our last [paper](https://arxiv.org/abs/2112.06511) (me, @AndreaCossu, @vlomonaco, Davide Bacciu).

I'm leaving this open for comments, let me know what you think.


vlomonaco:
Wow, this is massive! Thanks @AntonioCarta I think these are quite an important changes. It seems this is not impacting much the end users indeed. Still, the concept of experience changes a bit from _a static set of examples that can be processed together_ to _agnostic data sources_. We can discuss this in the next meeting and change the docs / tutorials accordingly.

PR #980, opened 2022-04-13
Deterministic test accuracy

### LINKED ISSUE #974: CI Test failing on python 3.8
`tests.training.test_strategies_accuracy.StrategyTest.test_multihead_cumulative` is failing on python 3.8.
I tested it locally and it is working. Here are the results of two independent runs:

```
TRAIN STREAM ACC:  {0: 0.8823529411764706, 1: 0.864406779661017, 2: 0.859504132231405, 3: 0.859504132231405, 4: 0.8760330578512396}
TRAIN STREAM ACC:  {0: 0.95, 1: 0.95, 2: 1.0, 3: 0.9333333333333333, 4: 0.6974789915966386}
```

the test requires average accuracy above 0.7.

ggraffieti:
Now the accuracy tests inside `tests/training/test_strategies_accuracy.py` are deterministic and should not give any problem when run in the unit testing action. 
This is only a quick fix for the test inside the file reported, a comprehensive revision of all the test is needed, especially is accuracy or the results of experiments are taken into account, in order to make all the test deterministic. 
This PR closes #974

### DIFF
--- a/tests/training/test_strategies_accuracy.py
+++ b/tests/training/test_strategies_accuracy.py
@@ -26,7 +26,8 @@ from avalanche.training.supervised.cumulative import Cumulative
 from avalanche.evaluation.metrics import StreamAccuracy, ExperienceAccuracy
 from avalanche.training.supervised.strategy_wrappers import PNNStrategy
 
-from tests.unit_tests_utils import get_fast_benchmark, get_device
+from tests.unit_tests_utils import get_fast_benchmark, get_device, \
+    set_deterministic_run
 
 
 class TestMLP(nn.Module):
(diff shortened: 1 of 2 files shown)


ggraffieti:
@AntonioCarta let me know if this PR can be closes as is (as a quick fix) or it's better to make all the test deterministic once for all without having a quick fix before.

AntonioCarta:
I think those two tests shoud be the only problematic ones. The others should are mostly integration tests, so they shouldn't break with different seeds.
Can you just make a function `set_deterministic_run` inside the test_utils so that we can easily use it when we need it somewhere else?

PR #984, opened 2022-04-15
Add RLScenario for Avalanche RL

NickLucche:
This PR adds a first implementation of the RLScenario we use in [Avalanche RL](https://github.com/ContinualAI/avalanche-rl). Essentially it adds support for a scenario wrapping streams of `gym.Env` environments.

Due to changes to the old "Strategy" model, I still haven't managed to test this PR within avalanche-rl, as the integration would require a few changes in order to match the new Templates-based structure.
Nonetheless, I would appreciate any feedback on this first implementation and whether it adheres to the new benchmark guidelines.

Also, I would like to raise a few issues regarding typing, as with the current implementation we're not able to implicitly infer the type of the experience (e.g. RLExperience) which is returned when iterating a stream (e.g. https://github.com/NickLucche/avalanche/blob/refactoring-avalanche-rl/tests/benchmarks/scenarios/test_rl_scenario.py#L21).
Perhaps we could borrow some of the Generics definitions from the old implementation of scenarios, or define a few subclasses just for better type hinting in `rl_scenario.py`.

### DIFF
--- a/avalanche/benchmarks/scenarios/rl_scenario.py
+++ b/avalanche/benchmarks/scenarios/rl_scenario.py
@@ -9,14 +9,108 @@
 # Website: avalanche.continualai.org                                           #
 ################################################################################
 """Reinforcement Learning scenario definitions."""
-from typing import Union, Sequence
+from avalanche.benchmarks.scenarios import CLExperience, ExperienceAttribute, CLScenario, EagerCLStream
+from typing import Callable, List, Union, Dict
+import numpy as np
+import torch
+import random
 
 try:
-    from gym import Env
+    from gym import Env, Wrapper
 except ImportError:
-    # empty class to make sure everything below works without changes
+    # empty classes to make sure everything below works without changes
     class Env:
+        pass
+    class Wrapper:
         pass
 
 
-RLStreamDataOrigin = Union[Env, Sequence[Env]]
+class RLExperience(CLExperience):
+
+    def __init__(self, env: Union[Env, Callable[[], Env]], n_envs: int = 1, task_label: int = None, current_experience: int = None, origin_stream=None):
+        # current experience and origin stream are set when iterating a CLStream by default
+        super().__init__(current_experience, origin_stream)
+        self.env = env
+        self.n_envs = n_envs
+        # task label to be (optionally) used for training purposes
+        self.task_label = ExperienceAttribute(task_label, use_in_train=True)
+
+    @property
+    def environment(self) -> Env:
+        # supports dynamic creation of environment, useful to instantiate envs for evaluation
+        if not isinstance(self.env, Env):
+            return self.env()
+        return self.env
+
+
+class RLScenario(CLScenario):
+
+    def __init__(self, envs: List[Env],
+                 n_experiences: int,
+                 n_parallel_envs: Union[int, List[int]],
+                 eval_envs: Union[List[Env], List[Callable[[], Env]]],
+                 wrappers_generators: Dict[str, List[Wrapper]] = None,
+                 task_labels: bool = True,
+                 shuffle: bool = False, 
+                 seed: int = None):
+
+        assert n_experiences > 0, "Number of experiences must be a positive integer"
+        if type(n_parallel_envs) is int:
+            n_parallel_envs = [n_parallel_envs] * n_experiences
+        assert len(n_parallel_envs) == len(envs)
+        assert all([n > 0 for n in n_parallel_envs]
+                   ), "Number of parallel environments must be a positive integer"
+        tr_envs = envs
+        eval_envs = eval_envs or []
+        self._num_original_envs = len(tr_envs)
+        self.n_envs = n_parallel_envs
+        # this shouldn't contain duplicate envs, but it's difficult to ensure if scenario isn't created through a benchmark generator
+        tr_task_labels = list(range(len(envs)))
+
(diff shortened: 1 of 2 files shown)


avalanche/benchmarks/scenarios/rl_scenario.py
```diff
@@ -9,14 +9,108 @@
...
+    # empty classes to make sure everything below works without changes
     class Env:
         pass
+    class Wrapper:
+        pass
+
+
+class RLExperience(CLExperience):
+
+    def __init__(self, env: Union[Env, Callable[[], Env]], n_envs: int = 1, task_label: int = None, current_experience: int = None, origin_stream=None):
```
AntonioCarta, line 30: can you add some docstrings for documentation?

AntonioCarta, line 30: Callable[[], Env] is needed to create environments lazily? Isn't it better to simplify this definition and make the entire stream lazy?

avalanche/benchmarks/scenarios/rl_scenario.py
```diff
@@ -9,14 +9,108 @@
...
+        # current experience and origin stream are set when iterating a CLStream by default
+        super().__init__(current_experience, origin_stream)
+        self.env = env
+        self.n_envs = n_envs
+        # task label to be (optionally) used for training purposes
```
AntonioCarta, line 35: I think we can drop the requirement of having mandatory task labels (eventually set to `None`).

Also, in a multi-task scenario you should also set `use_in_eval=True`

avalanche/benchmarks/scenarios/rl_scenario.py
```diff
@@ -9,14 +9,108 @@
...
+        # supports dynamic creation of environment, useful to instantiate envs for evaluation
+        if not isinstance(self.env, Env):
+            return self.env()
+        return self.env
+
+
+class RLScenario(CLScenario):
+
+    def __init__(self, envs: List[Env],
+                 n_experiences: int,
```
AntonioCarta, line 49: Do we need the `n_experiences` arguments? Since the constructor takes a list of environments I don't see a need for it. The user can manipulate the list easily to modify the stream before calling the constructor.

avalanche/benchmarks/scenarios/rl_scenario.py
```diff
@@ -9,14 +9,108 @@
...
+            tr_task_labels = tr_task_labels[:n_experiences]
+        elif n_experiences > len(tr_envs):
+            # cycle through envs sequentially, referencing same object to create a longer stream
+            for i in range(n_experiences - len(tr_envs)):
+                tr_envs.append(tr_envs[i % len(tr_envs)])
+                tr_task_labels.append(
+                    tr_task_labels[i % len(tr_task_labels)])
+
+        # move to template/strategy?
+        if seed is not None:
```
AntonioCarta, line 84: Setting the seed here is dangerous. Can we pass the seed where it's needed?

tests/benchmarks/scenarios/test_rl_scenario.py
```diff
@@ -0,0 +1,26 @@
+from avalanche.benchmarks.scenarios.rl_scenario import RLScenario, RLExperience
+import pytest
```
AntonioCarta, line 2: we don't use pytest

tests/benchmarks/scenarios/test_rl_scenario.py
```diff
@@ -0,0 +1,26 @@
...
+import numpy as np
+try:
+    import gym
+    skip = False
+except ImportError:
+    skip = True
+
+
+@pytest.mark.skipif(skip, reason="Need gym to run these tests")
```
AntonioCarta, line 11: you can use unittest here:
```
    @unittest.skipIf(...)
```

AntonioCarta (changes requested):

AntonioCarta:
> Due to changes to the old "Strategy" model, I still haven't managed to test this PR within avalanche-rl, as the integration would require a few changes in order to match the new Templates-based structure.

Of course, this would be the next step to figure out.

> Perhaps we could borrow some of the Generics definitions from the old implementation of scenarios

I agree that we need to fix the typing. Also, right now mypy is basically useless because it throws a ton of errors.

avalanche/benchmarks/scenarios/rl_scenario.py
NickLucche: We also need to support environments which handle the change of task internally (e.g. habitat) and whose instance must persist across experiences; I'd rather leave it as is for now

AntonioCarta: makes sense, let's keep it as is.

avalanche/benchmarks/scenarios/rl_scenario.py
AntonioCarta: This is still a problem, it should be `self.task_label = ExperienceAttribute(task_label, use_in_train=True, use_in_eval=True)`. If `task_labels=None` is you can also decide to not having the attribute or having  `self.task_label = ExperienceAttribute(task_label)`

AntonioCarta:
apart from the task labels, everything looks good.

NickLucche:
I've added some code to handle the detection of task labels which should cover the removal of the argument `n_experiences` and added a test showing its behavior

PR #993, opened 2022-04-20
Feat/add margin to early stopping

evertonaleixo:
As discussed in #990, this pull request adds a margin in the early stopping plugin.


AntonioCarta:
Thanks @evertonaleixo!

PR #907, opened 2022-02-07
Distributed support (Work In Progress)

lrzpellegrini:
This draft contains the implementation for supporting distributed (multi-GPU) training.

This is very rough, and many things are still to be completed (mainly checking that plugins work as expected).

The `examples` folder contains a simple example (to run naive and replay) and a bash script to launch it.

More commits to come! In the meantime, what do you think of this?

PR #996, opened 2022-04-22
Distributed support (rework)

lrzpellegrini:
This draft contains the implementation for supporting distributed (multi-GPU) training.

The examples folder contains a simple example (to run naive, replay, replay+scheduler) and a bash script to launch it.

This is an alternative to #907 created with the idea to minimize the changes required to the strategies.


lrzpellegrini:
It seems that unit tests failed due to a temporary error while downloading CIFAR. I triggered a re-run and unit tests are ok!

avalanche/distributed/distributed_batch.py
```diff
@@ -0,0 +1,129 @@
...
+    The current design, in which child classes have to implement
+    `_merge_single_values`, allows for this change to happen without affecting
+    child classes.
+    """
+
+    def __init__(self, name: str, initial_local_value: LocalT):
+        super().__init__(name, initial_local_value)
+        self._value_is_tuple = False
+
+    def _synchronize_distributed_value(self) -> LocalT:
```
AntonioCarta, line 55: can we abbreviate the names? `_set_local`, `_synchronize`, `_merge`

avalanche/distributed/distributed_batch.py
```diff
@@ -0,0 +1,129 @@
...
+    def _merge_objects(self, objects: List[LocalT]) -> DistributedT:
+        pass
+
+
+class DistributedBatch(DistributedObject[LocalT, LocalT], ABC):
+    """
+    An intermediate abstract class in charge of synchronizing data batches.
+
+    This class can handle batches as either tuples of elements (as usual) or
+    even single values.
```
AntonioCarta, line 39: I don't understand this. Why do we need a single class to deal with both tuples and single values? It seems easier to have different classes each with their own merge and sync methods.

avalanche/distributed/distributed_tensor.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Child classes must override `_merge_tensors` to define how those tensors
+    should be merged.
+    """
+    def _synchronize_distributed_value(self) -> Tensor:
+        return self._merge_tensors(
+            DistributedHelper.gather_all(self.local_value))
+
+    @abstractmethod
+    def _merge_tensors(self, tensors: List[Tensor]) -> Tensor:
```
AntonioCarta, line 25: maybe just `_merge`? I would avoid different names for each type since it's always the same operation (just different implementations).

AntonioCarta:
I left some general comments about the API and some things I didn't understand about the implementation. Overall, I think it's now much easier to add distributed support and integrate it. I think the API is a bit more verbose than it needs to be, especially due to the large number of methods that we need to add to a new distributed value. Personally, I would have chosen a more minimal set of methods but the solution as is is already quite good.

One thing that is missing, and it's quite important, is a test that checks which plugins actually work with the distributed training, which may not be obvious right now.

avalanche/training/plugins/evaluation.py
```diff
@@ -228,10 +229,45 @@ def before_eval(self, strategy: "SupervisedTemplate", **kwargs):
                 warnings.warn(msgw)
 
 
+class LazyDefaultLoggersList(Sequence["BaseLogger"]):
```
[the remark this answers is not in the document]
lrzpellegrini: Makes sense, but we would need to pass some kind of special value to describe the default logger. And I think we can't use None there, correct me if I'm wrong!

avalanche/training/templates/supervised.py
```diff
@@ -241,23 +247,50 @@ def make_eval_dataloader(
...
+        """
+        with self.use_local_input_batch():
+            return self._forward()
+
+    def _forward(self):
+        """Implementation of the forward pass."""
         return avalanche_forward(self.model, self.mb_x, self.mb_task_id)
 
+    @final
     def model_adaptation(self, model=None):
```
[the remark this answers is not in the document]
lrzpellegrini: The reason here is that we want the users to override `_model_adaptation` instead of `model_adaptation`. The _model_adaptation is then called from within the `with `self.use_local_model()` context. This is needed to ensure the users are changing the local model instead of the `model` (which may be the wrapped one).

examples/distributed_training.py
```diff
@@ -0,0 +1,229 @@
...
+    # MODEL CREATION
+    model = SimpleMLP(num_classes=scenario.n_classes)
+
+    optimizer = SGD(model.parameters(), lr=0.001, momentum=0.9)
+
+    # CREATE THE STRATEGY INSTANCE (NAIVE)
+
+    loggers = []
+    if DistributedHelper.is_main_process:
+        distr_str = 'single_process'
```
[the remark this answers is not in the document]
lrzpellegrini: Agree! I can add a experiment_name parameter as a cli argument to customize the log name.

avalanche/training/templates/supervised.py
```diff
@@ -241,23 +247,50 @@ def make_eval_dataloader(
...
+    @final
     def forward(self):
-        """Compute the model's output given the current mini-batch."""
+        """
+        Compute the model's output given the current mini-batch.
+
+        This method should not be overridden by child classes.
+        Consider overriding :meth:`_forward` instead.
+        """
+        with self.use_local_input_batch():
```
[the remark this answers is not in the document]
lrzpellegrini: I think I can create such a single context manager. However, using the explicit per-field context managers is more explicit in terms of readability (as you can immediately understand which fields are being affected).

avalanche/distributed/distributed_batch.py
lrzpellegrini: I think that `_synchronize` makes sense.

avalanche/distributed/distributed_batch.py
lrzpellegrini: A class that can handle both situations may be useful for unsupervised strategies that may have a batch made of the X value only. In that case, the batch is not a tuple and things need to be managed in a slightly different way. Switching classes depending on that may be more complicated than doing this...

lrzpellegrini:
@AntonioCarta I fixed the naming of the methods you mentioned and added the `use_local` strategy method. I also added, in the example, an in-code description of the changes to be made in the main script to support distributed training. I'll add more docstrings and fix the remaining things ASAP.

We still have to finalize a choice regarding the `_forward`, `_model_adaptation`, ... methods.

AntonioCarta:
Thanks!

> We still have to finalize a choice regarding the _forward, _model_adaptation, ... methods.

Maybe we could have a `forward` and a `_forward_impl`, with the idea that users call `forward` but override `_forward_impl`? Should fine if documented.

lrzpellegrini:
@AntonioCarta The PR is almost completed and I'd like to add more unit tests to be sure everything works fine.

The only real part that I still need to figure out is the `mb_output` collate. Currently, I slightly reworked the whole collate things to have an abstract class that contains:
- The collate_fn (the usual function passed to DataLoader)
- A collate for single values (that can be used to collate only x or y o t)
- A collate that merges batches (that were already collated) _[needed for distributed training support]_
- A collate that merges single value batches (that were already collated)  _[will be needed in the future to make metrics/logging in distributed training more efficient]_

The good part is that we can use instances of this class to populate the collate_fn field of AvalancheDatasets because they have a __call__ method that calls the collate_fn. For the moment, I have implemented the Collate class for classification and detection. However, this works for the input batch.

For distributed support, we also need to implement the same thing, but on the outputs. Synchronizing the output across processes is important to have the plugins and metrics aligned. I already took care of the loss by averaging it across processes, but mb_output needs a proper collate. The problem is that it has to be different from the input one (as the format may be different for inputs and outputs). We can't leverage the collate_fn from the experience dataset unless we enrich the Collate class to also contain the collate methods needed to manage the output (this is one possible solution). A different solution is to have the strategy itself define the collate for the output.

Do you any suggestion on this?

AntonioCarta:
I need to think about this. Unfortunately the output collate cannot be part of the dataset since it depends on the model's definition and its output.

At this point, I'm wondering whether it's just easier to enforce a more structured and general mini-batch definition. Something like [tensordict](https://github.com/pytorch-labs/tensordict/). It would provide a single and well defined method to collate values (also for scalar and tensor metrics).

AndreaCossu:
I agree that `TensorDict` (the official torch API for that is still experimental) or simply dictionaries are a more flexible choice to manipulate the minibatch information (x, y, task labels etc). However, even with `TensorDict` you need to know the semantics of the tensors to know how to collate them (how to collate multiple x together, multiple y together etc).

I think if we use dictionaries and 1. a collate for single values + 2. a collate that merges single value batches (often the same as 1.) we could recover all the other collate operations. 
Creating the minibatch in the dataloader amounts to use 1 + dictionary. Stacking together multiple mini-batches amounts to use 2 + dictionary.

AntonioCarta:
I was definitely underestimating the complexity of possible collate functions.

> I think if we use dictionaries and 1. a collate for single values + 2. a collate that merges single value batches (often the same as 1.) we could recover all the other collate operations.

There is also an additional problem. In general, it may not be true that the single collate is the same for all the values in the minibatch.

botcs:
Hi,

Thanks for the amazing job!
Any update on this?

lrzpellegrini:
I'm moving the development status to #1315. This PR, in its current state, is very difficult to merge in a single step as it is too big and based on a too old codebase.

PR #1002, opened 2022-04-26
Fix Templates type hints

### LINKED ISSUE #989: Fix type hints to make use of new benchmark structure
I think we should update the typing in function signatures such as https://github.com/ContinualAI/avalanche/blob/master/avalanche/training/templates/base.py#L74 requiring specific experiences to be more general, replacing them with the new [CLExperience](https://github.com/ContinualAI/avalanche/blob/master/avalanche/benchmarks/scenarios/generic_scenario.py#L73), perhaps even swapping the `Sequence` of experiences requirement with a [CLStream](https://github.com/ContinualAI/avalanche/blob/master/avalanche/benchmarks/scenarios/generic_scenario.py#L154).

NickLucche:
First draft attempting to fix #989

### DIFF
--- a/avalanche/training/templates/base.py
+++ b/avalanche/training/templates/base.py
@@ -60,7 +60,7 @@ class BaseTemplate:
         self.is_training: bool = False
         """ True if the strategy is in training mode. """
 
-        self.current_eval_stream: Optional[CLStream] = None
+        self.current_eval_stream: Optional[ExpSequence] = None
         """ Current evaluation stream. """
 
     @property
--- a/avalanche/training/templates/base.py
+++ b/avalanche/training/templates/base.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Sequence, Optional, Union, List
+from typing import Iterable, Sequence, Optional, Union, List
 
 import torch
 from torch.nn import Module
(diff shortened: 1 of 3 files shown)


avalanche/training/templates/base.py
```diff
@@ -52,15 +54,13 @@ def __init__(
...
+        self.experience: Optional[CLExperience] = None
         """ Current experience. """
 
         self.is_training: bool = False
         """ True if the strategy is in training mode. """
 
-        self.current_eval_stream: Optional[
-            Sequence[ClassificationExperience]
-        ] = None
+        self.current_eval_stream: Optional[CLStream] = None
```
AntonioCarta, line 63: this is taken from the train and eval arguments, so it should be an `ExpSequence`

AntonioCarta (changes requested):

AntonioCarta:
Thanks, just a minor comment and then we are ready to merge!

avalanche/training/templates/base.py
NickLucche: nice one, I'll fix it right up

NickLucche:
nice

PR #1008, opened 2022-05-04
Fix bug in cwr star plugin

ggraffieti:
Fixed a bug in CWR star plugin. 
Before the fix the fist time a class in is encountered the number of patterns is not saved into the variable containing the past number of examples. 
Now the number of past examples is saved correctly.


AntonioCarta:
We could also add a test that checks the counters

AndreaCossu:
The test is a good idea. In addition, it would be nice to add CWR and AR1 to the reproducible repository.
Tests are currently failing for the usual HTTP error, it should be sufficient to run them again.

lrzpellegrini (approved):

AndreaCossu (approved):

PR #1011, opened 2022-05-04
Add Generics for new benchmark components

NickLucche:
Proposal/first draft addressing the type hint issue which we currently have with CLStream and custom CLExperience (language server can't see their properties and methods).
An example of this is the [RLScenario](https://github.com/ContinualAI/avalanche/compare/master...NickLucche:refactoring-avalanche-rl#diff-2d765ee33fe8c41d986c0e95bb3cbbb972ad57c894905087fade0bf899636344R45) which now properly displays the return type of its train/eval streams (RLExperience).

I've also added task labels for eval experiences when creating an RLScenario.


AntonioCarta:
Great!

PR #1015, opened 2022-05-09
Refactor the CLEAR training example

linzhiqiu:
1. We add a metric to our benchmark
2. We add an example on how to use the linear features of CLEAR to train and evaluate


AntonioCarta:
Thanks!

AntonioCarta:
This is a problem with some of the tests that download the dataset. I will merge this anyway since it's a momentary error.

PR #1019, opened 2022-05-11
apidoc improvements

AntonioCarta:
- Made some minor changes to docstrings to fix errors
- added a small function that checks for classes missing from the apidoc
- added a readme in `docs` for documentation for developers building the docs.

This was quite painful.

TODO: there are many classes that are not in the apidoc but should be. Let's try to add them before the next release.

@AndreaCossu can you do this check for the evaluation modules? also some of the descriptions may need to be updated.

@lrzpellegrini can you do it for the detection stuff?

read the doc for info on how to compile the doc without seeing tons of errors. Please try to not commit file with syntax errors.

### DIFF
--- a/avalanche/benchmarks/classic/cinaturalist.py
+++ b/avalanche/benchmarks/classic/cinaturalist.py
@@ -55,8 +55,8 @@ def SplitInaturalist(
     eval_transform: Optional[Any] = _default_eval_transform,
     dataset_root: Union[str, Path] = None
 ):
-    """
-    Creates a CL benchmark using the iNaturalist2018 dataset.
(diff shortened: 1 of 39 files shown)


AntonioCarta:
@ggraffieti how can we silence pycodestyle on W605 warnings?

```
avalanche/training/supervised/strategy_wrappers.py:526:18: W605 invalid escape sequence '\*'
```
This is not an error, we need to escape `*` because it's a special character in the ReST markup.

ggraffieti:
You can edit the `setup.cfg' file in the main project directory and add warning W605 into the list of the ignored errors/warnings :+1:

Issue #886, opened 2022-01-20
Examples in the documentation


AntonioCarta:
We should add some examples in the documentation, at least for the high level classes and all the features more frequently used by the end user.

See #885 for an example

niniack:
Is the idea that an Avalanche documentation page such as [this about the SimpleCNN class](https://avalanche-api.continualai.org/en/v0.1.0/generated/avalanche.models.SimpleCNN.html#avalanche.models.SimpleCNN) would include an example [similar to this](https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html#:~:text=%3E%3E%3E%20m%20%3D%20nn.conv1d(16%2C%2033%2C%203%2C%20stride%3D2)) 

If I have got the idea, then I'm interested in contributing! If there is anything I should keep in mind to get started, please let me know.

AntonioCarta:
Yes, that's what we want to achieve. I think you can start immediately, you don't need much background on avalanche to do it. You can make a PR once you have a couple of examples if you want some comments.

PR #1022, opened 2022-05-11
Examples in docs

niniack:
### Adding code blocks as examples in the API Documentation

This PR is a result from discussion in Issue #886 and is inspired by PyTorch documentation ([here's an example](https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html#:~:text=%3E%3E%3E%20m%20%3D%20nn.conv1d(16%2C%2033%2C%203%2C%20stride%3D2)))

#### Changes: 
- Update API docs footer to be 2022 (previously was 2021)
- Add examples to Simple CNN and SimpleMLP classes. It defaults to python syntax highlighting.
- Use the `sphinx_copybutton` extension to allow for users to quickly copy code blocks.

#### Notes:
- The `sphinx_copybutton` extension can be installed in the conda environment with [these instructions](https://pypi.org/project/sphinx-copybutton/)

**Before**:
 
avalanche-old

**After**:

avalanche-new

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ logs/
 *.pyc
 doc_html/*
 .coverage
+tb_data/
+docs/generated/
--- a/avalanche/models/simple_cnn.py
+++ b/avalanche/models/simple_cnn.py
@@ -20,6 +20,13 @@ from avalanche.models.dynamic_modules import (
 class SimpleCNN(nn.Module):
     """
     Convolutional Neural Network
+    
+    **Example**::
+
+        >>> from avalanche.models import SimpleCNN
+        >>> n_classes = 10 # e.g. MNIST
+        >>> model = SimpleCNN(num_classes=n_classes)
+        >>> print(model) # View model details
     """
 
     def __init__(self, num_classes=10):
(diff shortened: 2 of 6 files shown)


AntonioCarta:
Thanks, it looks very nice. Can you add the dependency to `environment-dev.yml`?

niniack:
I'm curious if the extent of those examples is enough? It didn't feel right to throw in a full model-training example

Generally, is that level of "depth" good for examples?

AntonioCarta:
I think examples should be minimal. E.g. see [pytorch modules](https://pytorch.org/docs/stable/generated/torch.nn.Linear.html?highlight=linear#torch.nn.Linear).

AntonioCarta:
If you fix the pep8 I think we can merge this. You can use `pycodestyle` to check errors.

AndreaCossu:
I think the dependency should also be added to `docs/requirements.txt`

niniack:
Sounds good, I'll add the dep to `requirements.txt` and fix the PEP8 errors

niniack:
Looks like format on save was turned off on previous commits which is why that pep8 error slipped by. Thanks for letting me help out, hopefully I can help out some more with examples + other areas

Good luck on the release!

AntonioCarta:
Thanks @niniack, I will merge this. If you add other examples you can make other PRs, they should be much quicker now that we have all the dependencies in place.

PR #1024, opened 2022-05-13
Update `OnlineSupervisedTemplate`

HamedHemati:
This PR includes the following changes:

- `BaseOnlineSGDTemplate` is added: which is different from `BaseSGDTemplate` in `train_exp` and `training_epoch -> training_pass` callbacks and `PeriodicEval`.

- `OnlineSupervisedTemplate` is a sub-class of `BaseOnlineSGD` and implements `model_adaptation` for online strategies.

- Online naive and replay examples and strategy wrappers (online) are updated correspondingly.

- `InteractiveLogger` now works with online strategies: `before/after_training_exp` initialize and update the progress bar for online streams.

- `OnlineCLScenario` now support creating online scenario through a sequence of experiences.

Some notes regarding performance:
- OnlineNaive: works fine as expected.
- OnlineReplay: is slow for large buffers: I did some profiling tests and realized that the `AvalancheConcatDataset` part used in the update functions of ReservoirSampling slows down the buffer update.  I'm going to create an issue for this to see what we can do to speed up buffer updates in online streams.
- GSS: since the plugin is written for the older version, they use an internal buffer and update it after every forward. Therefore, it needs to be adapted to the current version.

I think it's fine to add these updates, and fix the existing issue (like slow buffer updates) in separate PRs.


avalanche/training/templates/online_supervised.py
```diff
@@ -1,209 +1,318 @@
...
+
+    def make_optimizer(self):
+        """Optimizer initialization.
+
+        Called before each training experiene to configure the optimizer.
+        """
+        # we reset the optimizer's state after each experience.
+        # This allows to add new parameters (new heads) and
+        # freezing old units during the model's adaptation phase.
+        reset_optimizer(self.optimizer, self.model)
```
AntonioCarta, line 302: is this correct? This would reset the momentum after each mini-batch. It's ok in CIL, not in OCL.

AntonioCarta:
two comments:
- We probably don't want to reset the optimizer here.
- can you add a test for the Naive OCL strategy?

once these are fixed, I think we can merge the PR.

HamedHemati:
> two comments:
> 
> * We probably don't want to reset the optimizer here.
> * can you add a test for the Naive OCL strategy?
> 
> once these are fixed, I think we can merge the PR.

`make_optimizer` is fixed for online supervised, and the test is also added.

avalanche/training/templates/online_supervised.py
```diff
@@ -1,209 +1,332 @@
...
-                module.adaptation(self.full_experience)
+        # For training:
+        if isinstance(self.experience, OnlineCLExperience):
+            # If the strategy has access to task boundaries, adapt the model
+            # for the whole origin experience to add the
+            if self.experience.access_task_boundaries:
+                avalanche_model_adaptation(model,
+                                           self.experience.origin_experience)
+            else:
+                self.model_params_before_adaptation = list(model.parameters())
```
AntonioCarta, line 275: I don't think this if-then-else is needed. Also notice that it is called at every experience in both cases

avalanche/training/templates/online_supervised.py
```diff
@@ -1,209 +1,332 @@
...
+        """Optimizer initialization.
+
+        Called before each training experience to configure the optimizer.
+        """
+        # We reset the optimizer's state after each experience if task
+        # boundaries are given, otherwise it updates the optimizer only if
+        # new parameters are added to the model after each adaptation step.
+
+        # We assume the current experience is an OnlineCLExperience:
+        if self.experience.access_task_boundaries:
```
AntonioCarta, line 309: This condition is incomplete. It should be something like:
```
self.experience.access_task_boundaries and self.experience.is_first_subexperience
```

Can you add a test to make sure we get this right?

tests/training/test_online_strategies.py
```diff
@@ -0,0 +1,105 @@
...
+        criterion = CrossEntropyLoss()
+        benchmark = self.load_benchmark(use_task_labels=False)
+        benchmark_streams = benchmark.streams.values()
+        ocl_benchmark = OnlineCLScenario(benchmark_streams)
+        return model, optimizer, criterion, ocl_benchmark
+
+    def test_naive(self):
+        # SIT scenario
+        model, optimizer, criterion, my_nc_benchmark = self.init_sit()
+        strategy = OnlineNaive(
```
AntonioCarta, line 45: can you test with and without boundaries? we need both.

Instead, I don't think testing multi-task is useful here. You can remove it if you want

avalanche/training/templates/online_supervised.py
HamedHemati: So there are three possibilities:
During training -> if we have task boundaries, we adapt to the whole original experience at the begggining of the original experience  to avoid model adaptation every sub-experience,
if there we have no task boundaries, we save the model params to call and adapt to `self.experience` to update optimizer later

During evaluation-> since the experience is most probably not an `OnlineCLExperience`, we just adapt the model to `self.experience` without saving current model parameters.

avalanche/training/templates/online_supervised.py
HamedHemati: The `access_task_boundaries` + `is_first_subepxerience` checking is done in `base_online_sgd` in line 156. There we first check both conditions to adapt the model and call `make_optimizer`. 

So here we only check task boundaries to check whether we need to reset (for task boundaries) or update (without task boundaries) the optimizer. Therefore, if the strategy has task boundaries, it means it was already checked in terms of being the first sub-exp inside online_base_sgd.

tests/training/test_online_strategies.py
HamedHemati: Ok, will add w/o task boundaries as well.

What do you mean by multi-task not being useful here?

HamedHemati: > Instead, I don't think testing multi-task is useful here. You can remove it if you want

I see what you mean now, that's a mistake :D. Will remove it!

avalanche/training/templates/online_supervised.py
AntonioCarta: What are you trying to achieve exactly? 

There is no difference between doing the adaptation at each step using `self.experience` or doing it on `self.experience.origin_experience`.

avalanche/training/templates/online_supervised.py
AntonioCarta: ok, it makes sense. I expected the if inside the method but your solution is also correct.

avalanche/training/templates/online_supervised.py
AntonioCarta: Ok, I see it now. You need the if because you check the task boundaries _before_ calling `model_adaptation`. IMO you are needlessly complicating the code here because there is no need to distinguish these two cases. Nonetheless, the code seems correct.

PR #1031, opened 2022-05-19
Evaluation and Logging: Updated FZTH and APIDoc for release 0.2.0

AndreaCossu:
Just one point: CLEARMetric is a class within the `benchmarks.classic.clear`, should I link it in the evaluation instead? It seems better to me since it computes the metrics used in the CLEAR paper.


AntonioCarta:
Not sure about `CLEARMetric`. It seems easier to keep it in the same module of the benchmark definition since it doesn't make sense to use it without it.

AndreaCossu:
Just to be clear, I meant to put it in the evaluation only for the APIDoc. 
Apart from that, there should be no missing classes for evaluation and logging.

PR #1035, opened 2022-05-27
Minor Fix in Examples

JuliousHurtado:
Fix some minor issues if the following examples:

`ewc_mnist.py `
- Decay Factor must be None we using "separate" mode in EWC

`pytorchcv_models.py`
- Device was set to "CPU" manually, change it to use the argument

`simple_ctrl.py`
- Change the name of the metrics
- Add logger to the second strategy (line 115)

`standalone_metric.py`
- Add atribute _task_labels_ to metric update

`synaptic_intelligence.py`
- Remove _tb_log_exp_name_ from logger

### DIFF
--- a/examples/ewc_mnist.py
+++ b/examples/ewc_mnist.py
@@ -80,6 +80,8 @@ def main(args):
         loggers=[interactive_logger, tensorboard_logger],
     )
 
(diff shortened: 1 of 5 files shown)


AntonioCarta:
Thanks @JuliousHurtado

PR #1036, opened 2022-05-27
Fix "training" and "putting it all together" notebooks from the FZTH tutorial

ggraffieti:
- Fix the code to work with the 0.2.0 version of Avalanche.
- Rewrite some text block to adapt them to the new training class hierarchy (+ templates, - BaseStrategy).


ggraffieti:
Please @AntonioCarta let me know if I missed something in the text or if there is something that could be improved!

AntonioCarta:
Thanks @ggraffieti, the notebook looks ok. I will check it again before the release to see if I can add something there, given the several changes that we did to the training module.

PR #1043, opened 2022-06-02
Optional Depencencies

ggraffieti:
- Management of optional dependencies
  - Only the core dependencies are installed when the library is installed
  - Optional dependencies are divided into non-exclusive groups (a dependency can be into more groups). Each group defines the particular set of dependencies required to extend avalanche with certain functionalities. 
    - So far we have 3 groups: detection, extra, rl
  - The optional dependencies are memorized in a txt file, where for each of them one or more groups are indicated. This also allows having an additional group: "all", that install all the optional dependencies at once. 
  - The optional dependencies can be installed by using `pip install avalanche-lib[<group>]`, e.g. `pip install avalanche-lib[detection]`
    - This also works if avalanche is already installed, it simply installs the required optional dependencies
-  Library works without any additional dependency installed
    - Avalanche now works even without any additional dependency installed. All the top-level imports work and no errors are thrown. 
    - The classes/methods that use optional dependencies are not included in the top-level imports but should be imported explicitly.
    - If a class that requires an additional dependency is imported and the additional dependency is not installed, a talking error is thrown, guiding the user to install the required optional dependency group.


ggraffieti:
In the `rl_scenario` (avalanche.benchmarks.scenarios.rl_scenario` the `gym` import is already enclosed inside a try-catch, but instead of throwing and error if gym is not installed, it defines the mock classes `Env` and `Wrapper`. Don't know if this is needed or we can substitute it with an error that indicated to the user to install the rl dependencies like in all the other cases.

AntonioCarta:
Thanks, this looks perfect! The gym class stub was used just to avoid the error. Now that we have optional dependencies, we can remove it and emit the same error of the other packages when gym is not avaiable.

NickLucche:
Yep good work, we can run tests assuming all dependencies are installed right? Currently I'm checking for gym installation before skipping tests when testing rl_scenario, is there a better way to do it taking into account these changes?

NickLucche (approved):

ggraffieti:
> Yep good work, we can run tests assuming all dependencies are installed right? Currently I'm checking for gym installation before skipping tests when testing rl_scenario, is there a better way to do it taking into account these changes?

I don't think we have a better way than trying to import the dependency, catch the error and don't run the tests if the dependency is not present. I'll take a look at the tests but I think your solution is already the best one.

PR #1047, opened 2022-06-03
Fixed ewc plugin for growing multitask head

AlbinSou:
Before, when using the "online" mode for EWC (which is the most useful one since it uses less memory), no regularization was applied on the new classifier heads in the case of a task-incremental scenario. Since the zip(old_importances, importances) was limited by the smallest importance vector, which is the old importance and does not contain new heads parameters. I replaced it with zip_longest so that new importances values are initialized for new heads.

The new behavior is the following: If they don't exist they are initialized directly as the importance value computed at the last task. If they already exist they are updated using the decay factor like other importance values.

I also added an example usage as a part of examples/ewc_mnist.py, the use case that is fixed is the one where the script is used in combination with --ewc_mode online and --scenario multitask_smnist

### DIFF
--- a/avalanche/training/plugins/ewc.py
+++ b/avalanche/training/plugins/ewc.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from typing import Dict, Tuple
 import warnings
+import itertools
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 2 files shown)


AntonioCarta:
Thanks Albin!

PR #1048, opened 2022-06-03
Add Expert Gate to Avalanche

### LINKED ISSUE #1023: Adding ExpertGate
I'm interested in adding ExpertGate to Avalanche. A few questions/thoughts I wanted to clarify before moving forward:

  - While the original ExpertGate model uses a pretrained AlexNet, this limits the ExpertGate to whatever task PyTorch's AlexNet was pretrained for (I think its vision tasks?). Should this implementation allow one to feed whichever model they'd like? Or, am I grossly misunderstanding something here?
  - I imagine ExpertGate would be implemented as a `Model`? I can also view it as a `Strategy` where you feed in your choice of model (see point above). Although, my gut says its should be a `Model`.

niniack:
Hi, this is my initial draft of the ExpertGate Plugin. It resolves #1023

A few points: 
- The ExpertGate model is implemented as a wrapper which holds two dictionaries, one for containing Autoencoders, another for holding "Experts" (Alexnets)
- Using the ExpertGate model requires using the ExpertGateStrategy, which relies on the ExpertGatePlugin.
- Almost all of the action happens in the `before_training_exp` and `before_eval_iteration` methods extended by the plugin
- Before training a new "expert" for each task (except the initial task)
  - An autoencoder is trained
  - A previous, most related expert is selected using the task_relatedness metric
  - The feature module is extracted from that most related expert and plugged into the "new expert" model
  - Based on the task_relatedness result, LWF is used or not for training
- Before each evaluation iteration
  - Each autoencoder reconstructs the input and a reconstruction error is produced
  - The reconstruction errors are fed through a softmax layer to determine the most relevant autoencoder
  - The expert associated with that autoencoder is selected for evaluation

I am still in the process of running the model with real scenarios to squash any silly bugs. I wanted to open this draft in case somebody wanted to check it out and suggest any pointers.


AntonioCarta:
Thanks! Let me know when you are ready for a full code review. From a quick glance it looks quite good.

As a small suggestion, if you move the code from `before_eval_iteration` inside the forward method (only during eval mode) I think it would be better. This way, you can call the model forward outside the avalanche strategy, which is always a good thing.

niniack:
I'm trying to update my expert-gate branch to incorporate changes from avalanche/main, how do you suggest I go about this?

AntonioCarta:
You can do a `git pull upstream master` on your local branch. The recent changes should not break anything in your code.

niniack:
Fair, I was trying to rebase to avoid the unnecessary merge commit but that didn't go very well. Maybe in another lifetime I'll understand how to do it :D

Thanks, I'll move stuff out of `before_eval_iteration` into the forward method (during eval mode). I'll drop another message here when I'm more comfortable with a code review!

niniack:
Hi, 

I have arrived at a point where I think I would greatly benefit from an additional set of eyes. 

My current concerns are: 
1. The loss for the "expert" model doesn't go down. Things I thought were culprits but I have ruled out:
- I thought the optimizer wasn't getting the right parameters. After checking that the model's parameters are the same object as the expert's parameters, I don't think this is the case
- The loss was wrong because the output of the classifier was a vector the size of the number of classes and the target was a single value. But, it seems pytorch's cross entropy loss handles this gracefully. 
- The learning rate was badly picked, but I've experimented with a few.

2. I think I am not using the MultiTaskModule correctly. I was initially using the `forward_single_task` method but it kept giving me errors regarding broadcasting the output of the classifier (e.g. 10 classes for mnist) into the highest target value of that task (maybe task 0 has digit 3, so it wants to broadcast [mb_size, 10] to [mb_size, 3])

3. (Kind of a minor concern) There is a ton of confusing output because so many things are getting trained and evaluated in the strategy. I guess this is a more "how do you guys handle this"-stylistic question

avalanche/models/expert_gate.py
```diff
@@ -0,0 +1,147 @@
...
+        self.expert_dict = nn.ModuleDict()
+
+        self.expert = (
+            models.__dict__[arch](pretrained=pretrained_flag)
+            .to(device)
+            .eval()
+        )
+
+    def forward(self, x, task_labels):
+        return self.expert(x)
```
AntonioCarta, line 147: shouldn't you use the task label here to call the correct expert for the task?

avalanche/training/supervised/expert_gate.py
```diff
@@ -0,0 +1,293 @@
...
+                            *args, 
+                            **kwargs):
+        super().before_training_exp(strategy, *args, **kwargs)
+
+        task_label = strategy.experience.task_label
+
+        # Always remove the LWF plugin before every experience
+        # If needed, it will be appended later
+        if (self.lwf_plugin in strategy.plugins):
+            strategy.plugins.remove(self.lwf_plugin)
```
AntonioCarta, line 110: why do you remove lwf?

niniack: From my understanding, if a plugin is present then it will be used when training. ExpertGate calls for using LwF only when task relatedness meets a threshold. So, before every experience, I remove the LwF plugin, in case the model used doesn't meet the threshold. 

I add the LwF plugin back, if it meets the threshold, in lines 128-132. If there is a better way to "suppress" plugins that I have missed out on, it would make more sense. I personally don't think what I have done is the cleanest code either.

avalanche/training/supervised/expert_gate.py
```diff
@@ -0,0 +1,293 @@
...
+                              strategy: "SupervisedTemplate",
+                              *args, 
+                              **kwargs):
+        super().before_eval_iteration(strategy, *args, **kwargs)
+
+        # Build a probability dictionary
+        probability_dict = OrderedDict()
+
+        # Softmax layer
+        softmax = Softmax(dim=0)
```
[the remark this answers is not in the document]
niniack: On the functional version: noted

Regarding dimensions: I will check this, thanks for pointing it out.

avalanche/models/expert_gate.py
```diff
@@ -0,0 +1,147 @@
...
+from .utils import FeatureExtractorBackbone
+
+from avalanche.models import MultiTaskModule
+from avalanche.models.utils import Flatten
+from avalanche.benchmarks.scenarios.generic_scenario import CLExperience
+
+
+def AE_loss(target, reconstruction):
+    loss_method = MSELoss(reduction="sum")
+    reconstruction_loss = loss_method(reconstruction, target)
```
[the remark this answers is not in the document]
niniack: Noted

niniack:
Thank you so much for taking the time, your code review is really appreciated. I will double check the things you have pointed out and get back.

> Is the problem in the autoencoder or the final model? 

The loss for autoencoder goes down. It is the AlexNet "expert" that doesn't seem to train. Consequentially, the final model performs poorly. Although, I will double check this claim after some of the things you pointed out in your code review.

> This is correct. If you do one-hot encoding it doesn't work.

Yep, got it. I haven't done one-hot encoding here. Thankfully shot that idea down by reading some of the documentation :D

> Are you following the original paper? Do you have the same hyperparameters?

Yes, I'm following the original paper. What I meant to say was: I thought this was an issue, but I've ruled it out by experimenting with a few values. 

> ExpertGate doesn't need task labels so you don't need a MultiTaskModule.

I see. I think I understand and it makes sense with my intuition. I will circle back if I realize I didn't understand.

Thanks again for this! I understand that you must be busy and these code reviews take time, but it was incredibly helpful to have you take a look! Hopefully, I can get this all cleared up soon!

niniack:
Hello, 

I have made several notable changes: 

- I figured out why the expert wasn't training: the optimizer wasn't updating the model's parameters as a new expert was being selected. I make use of the `update_optimizer` method now to make sure this happens. Training happens now.
- As recommended, I moved the expert selection logic from `before_eval_iteration` to the `forward` method when not in training mode so that a typical inference call could make use of that logic as well.
- I used the `functional` methods where I could
- I stopped using the `MultiTaskModule` for the `ExpertGate` class
- I modified how `LwF` is called to make it slightly more readable and also fix an issue where it was reusing the same plugin between experts. It should be using a "fresh" plugin to avoid cross-contamination.
- I added more documentation and cleaned up the example file.

I am a lot more confident about this submission and believe I have largely managed to get ExpertGate working in Avalanche!

Any suggestions are appreciated :)

AntonioCarta:
Thanks, this looks much cleaner. You also fixed all the previous issues.

Did you check the final performance? How does it compare to the original paper?

niniack:
Hello, 

Thank you for your patience. I spent some time to do aggressive testing (and breaking of things) to eventually fix the following concerns:

- Using deep copies to prevent irrelevant weights from changing (major bug)
- Reset optimizer instead of updating it. The latter was adding parameters instead of replacing (major bug)
- Fixing where tensors exist when a GPU is available. It's handled more strictly now with the device argument being passed around to every object.
- Updated how pretrained weights from AlexNet are grabbed, old method is going to be deprecated (minor)
- General code cleanup

Also, as suggested, I set up the scenario from the original paper. The simplest experiment from the paper uses MIT Indoor Scenes 67, Caltech Birds 200, and Oxford Flowers 102, and treats each of these datasets as a separate task. Please find the results below.
 
| Source            | Method                                           | Scenes | Birds | Flowers | avg  |
|-------------------|--------------------------------------------------|--------|-------|---------|------|
| Original Paper    | Fine-tuned AlexNet on Scenes                     | 63.4   | -     | -       | -    |
| Original Paper    | Fine-tuned AlexNet on Scenes -> Birds            | 50.3   | 57.3  | -       | -    |
| Original Paper    | Fine-tuned AlexNet on Scenes -> Birds -> Flowers | 46.0   | 43.9  | 84.9    | 58.2 |
| This Pull Request | Fine-tuned AlexNet on Scenes                     | 56.9   | -     | -       | -    |
| This Pull Request | Fine-tuned AlexNet on Scenes -> Birds            | 36.3   | 45.1  | -       | -    |
| This Pull Request | Fine-tuned AlexNet on Scenes -> Birds -> Flowers | 32.2   | 35.4  | 65.2    | 44.3 |

| Source            | Method     | Scenes | Birds | Flowers | avg  |
|-------------------|------------|--------|-------|---------|------|
| Original Paper    | ExpertGate | 63.5   | 57.6  | 84.8    | 68.6 |
| This Pull Request | ExpertGate | 58.9   | 45.3  | 65.6    | 56.6 |

Notes on the results:

- The original paper does not really describe any hyperparameters aside from the latent dimensions of the Autoencoder (100 neurons). In fact, it even cites the wrong paper for the CUB-200 dataset. 
- For all experiments, I trained with
    - 50 epochs for AlexNets
    - 10 epochs for autoencoders (relevant only in ExpertGate)
    - SGD momentum=0.9 weight_decay=0.0005
    - AlexNet learning rate 1e-3 with StepLR, stepsize 20 and gamma 1e-1
    - AutoEncoder learning rate 5e-4 (relevant only in ExpertGate)

Despite the fact that the accuracies my experiments achieved are lower than that from the paper, I claim that the ExpertGate implementation here works as expected. 

Considering only the first task (scenes), the naive fine-tuning on AlexNet results in catastrophic forgetting of MIT Scenes by the time it fine-tunes on the final task (Oxford Flowers). On the other hand, after training on all three tasks, the ExpertGate model achieves similar accuracy to an AlexNet fine-tuned only on the first task. Simply put, the fine-tuned AlexNet kind of forgets about Scenes by the end, but the ExpertGate doesn't forget at all (unsurprising, because it has a separate "expert" for each task)

I attribute the lower accuracies in my results to my hyperparameter selection. I don't have access to what configuration was used for training. From some experimentation and reading around, I came up with these sensible and typical hyperparameters. 

Hopefully, this is a meaningful contribution to Avalanche. Thank you for reviewing this for me.

AntonioCarta:
Hey @niniack thanks for all you work. I agree with you that the performance gap seems to be due a problem in the base performance of the model. I think the code is ready to be merged in Avalanche in its current state.

To warn the users that there may still be some minor issues, I suggest to add a warning inside the constructor:
```
warnings.warn("This strategy is currently in the alpha stage and we are still working to reproduce the original paper's results. You can find the code to reproduce the experiments at github.com/continualAI/continual-learning-baselines")
```

So, I would ask you to do a few last things before merging the PR:
- adding the warning.
- adding a test under `tests/training/test_strategies.py`. You can copy/paste the one for the other strategies.
- (after the PR) can you push your scripts to the [continual-learning-baselines](github.com/continualAI/continual-learning-baselines) repository? This will help us to reproduce the results and possibly match the paper's performance in the future.

Thanks again for all your work.

niniack:
Hi, thank you for taking the time to review the work. 

I will add the warning message and add a test in `tests/training/test_strategies.py`. 

In regards to the baselines repository: I can publish the script I used but I wrote a few custom dataloading functions for the datasets. I wouldn't be worried if I was using Pytorch datasets, but in this case I had to download these datasets myself since they aren't on Pytorch. So, is it okay for me to include an "auto-download" the way Pytorch does?

AntonioCarta:
All the datasets that we provide in Avalanche have automatic download if possible. You can put them in `avalanche.benchmarks.datasets`. Check Tiny ImageNet for a simple example of the API that we follow. Also, in the script for continual-learning-baselines you can put any custom code that you need to reproduce the experiments, it's not a problem.

AntonioCarta:
Hi @niniack, do you have any updates?

niniack:
> Hi @niniack, do you have any updates?

Hi, thanks for your patience. I had to take a pause and incidentally noticed a not-instantly-fixable bug while preparing the final fixes for this PR. Now that things are picking up, I expect to be able to wrap it up within a week. 

Thanks for checking in

niniack:
Hi @AntonioCarta, 

Updates:
- Significantly cleaned up the example file 
- Bug has been identified and I added a note at the top in the example file as well as fixed the example
- Added alpha warning
- Added test

Once the PR is merged, I will add the benchmark script to the baselines repository

AntonioCarta:
Thanks! everything looks in order now. Can you fix the merge issues? We removed the `suppress_warning` argument from the logger (we removed the warnings).

I will merge the PR as soon as the CI is green.

AntonioCarta:
Hi niniack, I'm still getting a bunch of errors from your strategy:
```
Error
Traceback (most recent call last):
  File "D:\OneDrive - University of Pisa\Uni\code_repo\avalanche\tests\training\test_strategies.py", line 834, in test_expertgate
    model = ExpertGate(shape=(3, 227, 227), device=self.device)
  File "D:\OneDrive - University of Pisa\Uni\code_repo\avalanche\avalanche\models\expert_gate.py", line 181, in __init__
    models.__dict__[arch](
  File "C:\Users\w-32\Anaconda3\envs\avalanche-env\lib\site-packages\torchvision\models\alexnet.py", line 62, in alexnet
    model = AlexNet(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'weights'
```

niniack:
Hi, apologies it has been a very busy few months. I'll take a look at what I need to wrap this PR up soon!

AntonioCarta:
Sure, don't worry. I tried to fix it myself before the release but I got a bit busy with the other release work and in the end I couldn't make.

niniack:
The tests fail on:
`FAIL: test_cumulative_accuracy (test_metrics.GeneralMetricTests)`
which is unrelated, is this a wider issue?

AntonioCarta:
yes, it's unrelated. @AlbinSou please check this

examples/expert_gate_perf.py
```diff
@@ -0,0 +1,466 @@
+import os
+import torch
```
AntonioCarta, line 2: Can you remove this file and add it to [continual-learning-baselines](github.com/continualAI/continual-learning-baselines) instead? we are moving the experiments there.

avalanche/training/supervised/strategy_wrappers.py
```diff
@@ -434,6 +437,70 @@ def __init__(
         )
 
 
+class AETraining(SupervisedTemplate):
```
AntonioCarta, line 440: add docstring

avalanche/training/supervised/expert_gate.py
```diff
@@ -0,0 +1,363 @@
...
+from avalanche.models.dynamic_optimizers import reset_optimizer
+from avalanche.training.supervised import AETraining
+from avalanche.training.templates import SupervisedTemplate
+from avalanche.training.plugins import (SupervisedPlugin,
+                                        EvaluationPlugin,
+                                        LwFPlugin)
+from avalanche.training.plugins.evaluation import default_evaluator
+
+
+class ExpertGateStrategy(SupervisedTemplate):
```
AntonioCarta, line 34: add ref to paper

avalanche/models/expert_gate.py
```diff
@@ -0,0 +1,230 @@
...
+
+        return x.view(-1, *self.shape)
+
+
+class ExpertModel(nn.Module):
+    """The expert classifier behind the autoencoder that is trained for a
+    specific task.
+    """
+
+    def __init__(self,
```
AntonioCarta, line 99: docstring with params

avalanche/models/expert_gate.py
```diff
@@ -0,0 +1,230 @@
...
+        # Encode input
+        x = self.encoder(x)
+
+        # Reconstruction
+        x = self.decoder(x)
+
+        return x.view(-1, *self.shape)
+
+
+class ExpertModel(nn.Module):
```
AntonioCarta, line 94: can you expand this a bit?

avalanche/models/expert_gate.py
```diff
@@ -0,0 +1,230 @@
...
+from .utils import FeatureExtractorBackbone
+
+from avalanche.models import MultiTaskModule
+from avalanche.models.utils import Flatten
+from avalanche.benchmarks.scenarios.generic_scenario import CLExperience
+
+
+def AE_loss(target, reconstruction):
+    """Calculates the MSE loss for the autoencoder by comparing the 
+    reconstruction to the pre-processed input.
```
AntonioCarta, line 22: missing parameter descriptions

niniack:
> Hey @niniack I added some comments in the places where the docstring are missing. I can merge this as soon as you are ready. Can you open a PR at continual-learning-baselines with the script to reproduce the performance?

Sure, I'll add those in, thanks for catching those!

wrt to continual-learning-baselines, right now, I haven't been able reproduce the results from the paper. But, I haven't tried too hard. I also put in a user warning (as you suggested a while ago) about this. Still, would you like me to open up a PR there? Or, should I do this when I get a chance to mess with the hyperparameters? Memory is hazy, but I don't think the paper was very clear with its hyperparameter selection.

AntonioCarta:
> wrt to continual-learning-baselines, right now, I haven't been able reproduce the results from the paper. But, I haven't tried too hard. I also put in a user warning (as you suggested a while ago) about this. Still, would you like me to open up a PR there? Or, should I do this when I get a chance to mess with the hyperparameters? Memory is hazy, but I don't think the paper was very clear with its hyperparameter selection.

yes, it's easier to keep track of its status in the continual-learning-baselines repository.

AntonioCarta:
@niniack Thank you for all the work. Please remember to open the PR at continual-learning-baselines.

PR #1049, opened 2022-06-06
Added learn avalanche in 5 minutes notebook

AndreaCossu:
I converted the tutorial into a notebook and adjusted the code to work with the current avalanche version. Is there something else needed? Is this automatically used in the gitbook in place of the current version?


AndreaCossu:
Yes, it may be better. Let's see how it goes.

PR #1051, opened 2022-06-06
AMCA metric

lrzpellegrini:
This PR adds support for the Class Accuracy and Average Mean Class Accuracy metrics. This extends the #679 PR.

- Class Accuracy computes the per-class accuracy. It will do so separately for each task.
- The AMCA metric will compute its value for each `stream/task` and it will work only on the eval phase. In addition, it will automatically ignore periodic evaluations (this behavior can be customized).

Creating the AMCA metric has been quite rough, as this is the first stateful metric (others are computed and reset within the scope of a phase or less). @AntonioCarta @AndreaCossu can you give this a look? I introduced an extended metric mechanism and generalized the existing "value naming" procedure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ logs/
 doc_html/*
 .coverage
 tb_data/
+csvlogs/
 docs/generated/
--- a/avalanche/evaluation/metric_definitions.py
+++ b/avalanche/evaluation/metric_definitions.py
@@ -390,7 +390,8 @@ class ExtendedPluginMetricValue:
         """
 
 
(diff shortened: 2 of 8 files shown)


AntonioCarta:
I don't understand what is extended here. Metric plugins can emit heterogeneous `MetricValues`, without any constraints on names or anything. Theoretically, you could have a single `MetricPlugin` computing all your desired metrics. Maybe it's easier if we discuss this next wednesday.

It seems that the tests are failing because of a missing dependency.

AndreaCossu:
I'm fine with the implementation, but it seems that `fmean` is not present in the `statistics` package. Maybe you meant `mean`?

PR #1056, opened 2022-06-08
Support py3.10 and warning for cope

AndreaCossu:
Added support for python 3.10. It works in a fresh environment I created locally.

Added warning for COPE since it is not working currently.


AndreaCossu:
Tests are failing because there is no docker container for py3.10 but that should be created automatically after merging the PR (we have an action for that). I'll merge it since the tests are completing locally.

PR #1082, opened 2022-07-01
Update online strategy profiler

HamedHemati:
This PR updates the online strategy profiler:

I removed the obsolete profilers and added `CProfile` to for more detailed profiling.

Current profilers:
`profile_online_naive_no_avl` -> online naive without Avalanche
`profile_online_naive_lazy_stream` -> online naive without Avalanche using lazy streams
`profile_online_avl` -> online (naive/er) using Avalanche

This is an example profiling output for Online ER:
It's interesting to see that almost half of the training time for the Online ER strategy is spent initializing datasets (most probably the `AvalancheConcatDataset` part):

Screenshot 2022-07-01 at 23 04 51

** cumtime refers to the amount of time spent on the function itself + all its internal calls

### DIFF
--- a/profiling/online_strategy.py
+++ b/profiling/online_strategy.py
@@ -14,86 +14,113 @@ This simple profiler measures the amount of time required to finish an
 experience in an online manner using different strategies and options.
 """
 
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-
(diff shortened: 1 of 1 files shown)


AntonioCarta:
Thanks. I opened #1086 to keep track of this issue.

PR #1089, opened 2022-07-07
AvalancheDatasetType removal, dataset collate_fn used in dataloading during training and eval, sequential benchmark 

AndreaCossu:
* AvalancheDatasetType is now removed. This will cause regressions because signature of many functions/methods has changed.
* Dataloading in strategies now checks if the dataset has a "collate_fn" function and uses that unless one is specified through kwargs (which takes precedence).
* I added the Speech Commands dataset for sequence classification, providing an example with a custom collate embedded into the dataset.


AntonioCarta:
Thanks!

PR #1090, opened 2022-07-08
MultiTask Improvements

AntonioCarta:
- loss & accuracy are task-agnostic by default. Improves logging in multitask settings.
- classes ids from zero in each exp has been set to False on all the benchmarks. Previously it was True and it caused a lot of surprise and bugs to end users.
- removed benchmark from EvaluationPlugin arguments. The benchmark is taken directly from the current experience. Removed the corresponding warning.

### DIFF
--- a/avalanche/evaluation/metrics/accuracy.py
+++ b/avalanche/evaluation/metrics/accuracy.py
@@ -20,12 +20,12 @@ from collections import defaultdict
 
 
 class Accuracy(Metric[float]):
-    """
-    The Accuracy metric. This is a standalone metric.
+    """Accuracy metric. This is a standalone metric.
 
-    The metric keeps a dictionary of <task_label, accuracy value> pairs.
-    and update the values through a running average over multiple
-    <prediction, target> pairs of Tensors, provided incrementally.
+    The update method computes the accuracy incrementally
+    by keeping a running average of the <prediction, target> pairs
+    of Tensors provided over time.
+
     The "prediction" and "target" tensors may contain plain labels or
     one-hot/logit vectors.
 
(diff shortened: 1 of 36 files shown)


avalanche/evaluation/metrics/accuracy.py
```diff
@@ -379,6 +445,7 @@ def accuracy_metrics(
 
 __all__ = [
     "Accuracy",
+    "TaskAwareAccuracy",
```
AndreaCossu, line 448: Shouldn't the `*_metrics` functions exposes the `split_by_task` option? This would make it easier to get the corresponding metrics.

avalanche/training/plugins/evaluation.py
```diff
@@ -44,20 +43,16 @@ def __init__(
...
-            If False, warnings and errors will be raised following
-            `benchmark` and `strict_checks` behavior.
         """
         super().__init__()
         self.collect_all = collect_all
-        self.benchmark = benchmark
         self.strict_checks = strict_checks
-        self.suppress_warnings = suppress_warnings
+
+        self.benchmark = benchmark
```
AndreaCossu, line 54: Why the `EvaluationPlugin` does not check benchmark anymore? I know it will cause regressions but in this case I think it would be cleaner to remove it from the code.

avalanche/evaluation/metrics/accuracy.py
AntonioCarta: the `*_metrics` just provide reasonable defaults.

avalanche/training/plugins/evaluation.py
AntonioCarta: you mean removing `self.benchmark`? it's used like before. The only difference is that `self.benchmark` is initialized automatically using the experience. This also explains why we still have `strict_checks`.

AndreaCossu: I meant removing `benchmark` from the `__init__` signature, since I noticed it has been removed from the APIdoc. If the user can still provide it as input it's better to add it again in the APIdoc.

PR #1101, opened 2022-07-28
Fixed EWC bug in importances, converted Permuted and Rotated MNIST targets in 0-9

### LINKED ISSUE #1099: in EWC the penalty is 0
🐛 **Describe the bug**
Currently in [EWC ](https://github.com/ContinualAI/avalanche/blob/master/avalanche/training/plugins/ewc.py)plugin the penalty is always 0 
```
n_units = saved_param.shape[0]
cur_param = saved_param[:n_units]
penalty += (imp * (cur_param - saved_param).pow(2)).sum()
```

because `cur_param` is equal to `saved_param` in the penalty from computation `penalty += (imp * (cur_param - saved_param).pow(2)).sum()`

AndreaCossu:
Fixed #1099 bug.
Now Permuted and Rotated MNIST show targets in 0-9 for each experience.


AndreaCossu:
@AntonioCarta is it ok for you using these benchmarks this way? 
For EWC I have checked and penalties are now >0. 
The check on the performance on PMNIST is on-going.

AntonioCarta:
I'm merging this because we need the EWC fix soon. For the benchmarks, I think both are used frequently, we just disagree on which one is more frequent. We should add an option to recover the old behavior too.

AndreaCossu:
Ok, also consider that the previous implementation broke a single head with 10 units, while the current one is compatible with both single and multi-head.

PR #1118, opened 2022-09-05
WIP: optional task and class labels in AvalancheDataset

AntonioCarta:
*Still a work in progress*

AvalancheDatasets are doing many different things together:
- advanced indexing
- management of class/task labels
- management of transformations
- efficient concat/subset operations needed for replay

This PR separates all these features into different components. The advantage is that now we can do naturally some things that were a bit clunky such as:
- having datasets without class/task labels
- easily add new dataset types without having to worry about compatibility with task labels/collate_fn and other details of supervised datasets.


avalanche/__init__.py
```diff
@@ -11,10 +11,10 @@
 
 
 def _avdataset_radd(self, other, *args, **kwargs):
-    from avalanche.benchmarks.utils import AvalancheDataset
+    from avalanche.benchmarks.utils import AvalancheClassificationDataset
```
lrzpellegrini, line 14: This should be valid for all AvalancheDataset(s), not only for classification datasets.

avalanche/benchmarks/utils/data.py
```diff
@@ -0,0 +1,714 @@
...
+        ####################################
+        # Init data attributes
+        ####################################
+        if isinstance(dataset, AvalancheDataset):
+            # inherit data attributes from original dataset
+            self._data_attributes = {**dataset._data_attributes}
+
+        if data_attributes is not None:
+            da_dict = {da.name: da for da in data_attributes}
+            self._data_attributes.update(da_dict)
```
lrzpellegrini, line 120: I think that a way to add dataset fields as data attributes by specifying their name (using the attribute with the same name of the dataset object) may be useful. Is it doable?

avalanche/benchmarks/utils/data.py
```diff
@@ -0,0 +1,714 @@
...
+
+        for el in self._data_attributes.values():
+            setattr(self, el.name, el)
```
lrzpellegrini, line 123: Maybe I'm paranoid, but we should check for disallowed names (like the ones starting with `_`, `collate_fn`, etc.)

avalanche/benchmarks/utils/data.py
```diff
@@ -0,0 +1,714 @@
...
+                print("PARENT TRANSFORMS:\n")
+                dd.print_nonfrozen_transforms()
+
+    def print_transforms(self):
+        """Prints the current transformations."""
+        self.print_frozen_transforms()
+        self.print_nonfrozen_transforms()
+
+    @property
+    def transform(self):
```
lrzpellegrini, line 172: I think that we can make it read-only (by not declaring a setter)

avalanche/benchmarks/utils/transform_groups.py
```diff
@@ -0,0 +1,128 @@
...
+        if "train" not in transform_groups:
+            transform_groups["train"] = None
+
+        if "eval" not in transform_groups:
+            transform_groups["eval"] = None
+
+    def __getitem__(self, item):
+        return self.transform_groups[item]
+
+    def __setitem__(self, key, value):
```
lrzpellegrini, line 46: In general I don't love the idea of having mutable objects in the dataset, same for the with_transform method. The issue is that we may end up having datasets in strategy fields, plugins (especially the replay plugin), etc. that change behaviour in time.

avalanche/benchmarks/utils/transforms.py
```diff
@@ -1,13 +1,26 @@
+"""
+    Avalanche transformations are multi-argument.
+"""
 import warnings
 from typing import Callable, Sequence
 from inspect import signature, Parameter
 
 
-class Compose:
+class MultiParamTranform:
```
lrzpellegrini, line 9: Typo

avalanche/benchmarks/utils/data.py
AntonioCarta: Datasets are designed to be immutable objects. Changing attributes should be done by wrapping existing datasets e.g. `new_data=AvalancheDataset(old_data, data_attributes=[...])`. If we allow mutability indexing breaks.

avalanche/benchmarks/utils/transform_groups.py
AntonioCarta: notice that right now the `__getitem__` pass the current group explicitly to the datasets in the hierarchy. So, the current transformation group can be safely changed without problems. However, I agree that it may be better to make transformation groups as an immutable object. We could even remove `self.current_group` from transformation groups.

avalanche/benchmarks/utils/data.py
lrzpellegrini: Sorry, I put this in the wrong way. What I was trying to say is that, when wrapping a non-Avalanche dataset that contains a useful field, that field could be specified like this:

```
plain_dataset = ...
avl_dataset = AvalancheDataset(plain_dataset, data_attributes=['some_field', ...])
```

instead of:
```
plain_dataset = ...
attributes = [DataAttribute(plain_dataset.some_field, name='some_field'), ...]
avl_dataset = AvalancheDataset(plain_dataset, data_attributes=attributes)
```

just to improve usability a little bit.

AntonioCarta:
New PR at #1142

PR #1122, opened 2022-09-08
Checkpointing support

lrzpellegrini:
This PR adds support for checkpointing.

Checkpointing is implemented as a strategy plugin that will store a checkpoint after each evaluation phase.

The entire strategy object is stored, as well as the state of random number generators.

Particular custom procedures have been implemented to manage loggers and a couple of torchvision datasets.

An example is provided which includes a description of the steps required to add checkpointing in the main script.

This PR also introduces a fix in `data_loader.py` when `persistent_workers` is used and adds the download progress bar for inaturalist.


vlomonaco:
This is super needed, thanks @lrzpellegrini !

examples/task_incremental_with_checkpointing.py
```diff
@@ -0,0 +1,235 @@
...
+def main(args):
+    # FIRST CHANGE: SET THE RANDOM SEEDS
+    # In fact, you should to this no matter the checkpointing functionality.
+    # Using `use_deterministic_algorithms` may require setting the
+    # CUBLAS_WORKSPACE_CONFIG=:4096:8 environment variable.
+
+    # Remember to load checkpoints by setting the same random seed used when
+    # creating them...
+    RNGManager.set_random_seeds(1234)
+    torch.use_deterministic_algorithms(True)
```
AntonioCarta, line 51: do we have to use deterministic algorithms? Because those are often slower. From pytorch docs: "Deterministic operations are often slower than nondeterministic operations, so single-run performance may decrease for your model. "

tests/test_checkpointing_plugins_all.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Script used to manually test various combinations of plugins when used with
+# the checkpointing functionality
+set -euo pipefail
+./test_checkpointing_plugins.sh
```
AntonioCarta, line 5: can we add this to the unittests so that they are run by the CI? If too difficult because of multiprocessing, we could just add this script to the CI pipeline?

AntonioCarta: Also, shouldn't we check that the results of the checkpointed strategy are the same of the non-checkpointed one?

examples/task_incremental_with_checkpointing.py
lrzpellegrini: They are not mandatory. Determinism is needed in the example to check if the checkpointing plugin can correctly resume the training. I'll add an explanation to the examples.

avalanche/logging/tensorboard_logger.py
```diff
@@ -92,6 +86,22 @@ def log_single_metric(self, name, value, x_plot):
         elif isinstance(value, TensorImage):
             self.writer.add_image(name, value.image, global_step=x_plot)
 
+    def _make_writer(self, tb_log_dir, filename_suffix):
+        writer = SummaryWriter(tb_log_dir, filename_suffix=filename_suffix)
+
+        # Shuts down the writer gracefully on process exit
+        # or when this logger gets GCed. Fixes issue #864.
+        # For more info see:
+        # https://docs.python.org/3/library/weakref.html#comparing-finalizers-with-del-methods
```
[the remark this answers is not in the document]
lrzpellegrini: weakref should handle both graceful and unexpected exit. A more precise test may be needed, but this should be the safer way to close the writer.

avalanche/training/determinism/cuda_rng.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Seeds the state of the cpu RNG alone.
+
+    This is different from `torch.random.manual_seed`, which also initializes
+    the cuda RNGs.
+    """
+    default_generator.manual_seed(seed)
+
+
+def cuda_rng_save_state():
```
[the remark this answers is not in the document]
lrzpellegrini: We can make this optional, but it is a good addiction and doesn't "cost" much.

avalanche/training/plugins/checkpoint.py
```diff
@@ -0,0 +1,347 @@
...
+                raise RuntimeError(
+                    'Cannot update the strategy CheckpointPlugin: more than '
+                    'one found.')
+            elif len(checkpoint_plugin_indices) == 1:
+                to_replace_idx = list(checkpoint_plugin_indices)[0]
+                strategy.plugins[to_replace_idx] = self
+
+        return strategy, exp_counter
+
+    def after_eval(self, strategy: BaseSGDTemplate, *args, **kwargs):
```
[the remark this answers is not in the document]
lrzpellegrini: I'm not sure, but it may be better if checkpointing is skipped on periodic eval. I'll try to implement  this check.

tests/test_checkpointing_plugins_all.sh
lrzpellegrini: We can add the script to the pipeline, but I'm not sure how to check for consistency of results... maybe using the metrics dictionary returned from "eval"?

AntonioCarta: yes. Using the dictionary is enough. Please test also a couple of different configurations for PeriodicEval to check that serialization does not happen in the middle of an experience training.

avalanche/training/plugins/checkpoint.py
```diff
@@ -212,6 +215,10 @@ def load_checkpoint_if_exists(
         return strategy, exp_counter
 
     def after_eval(self, strategy: BaseSGDTemplate, *args, **kwargs):
+        if self._training:
```
AntonioCarta, line 218: isn't it easier to checkpoint inside `after_training_exp`?

examples/task_incremental_with_checkpointing.py
```diff
@@ -42,12 +42,15 @@
...
     # Remember to load checkpoints by setting the same random seed used when
     # creating them...
     RNGManager.set_random_seeds(1234)
+
+    # Determinism is not strictly required, but it is here added to show that
+    # checkpointing doesn't alter the results of an experiment.
+    # If you don't want to use deterministic algorithms, you can remove this.
+    # Using `use_deterministic_algorithms` may require setting the
+    # CUBLAS_WORKSPACE_CONFIG=:4096:8 environment variable.
     torch.use_deterministic_algorithms(True)
```
AntonioCarta, line 56: I think this is a bad practice in general. Setting the seeds is useful, but setting `use_deterministic_algorithms` is harmful, apart from the debugging purposes. Can we split the test script from the example? So that we can have an example without the determinism?

tests/test_checkpointing_plugins_all.sh
lrzpellegrini: I added a test script to check if metrics are exactly the same. I tested periodic eval "by hand" 😅

avalanche/training/plugins/checkpoint.py
lrzpellegrini: It is doable. In fact, we may checkpoint both after train and eval. In that case, we also need add to a flag to tell if the loaded checkpoint was created after train or eval (and add a couple of 'if's to the main script).

examples/task_incremental_with_checkpointing.py
lrzpellegrini: Agreed, I'll split the scripts!

avalanche/training/plugins/checkpoint.py
AntonioCarta: is there any difference? no plugin should ever modify things inside the eval loop. I'm suggesting `after_training_exp` only because it seems the logical place to put it. If this is not true, maybe there is some complicated interaction that I'm ignoring.

lrzpellegrini: It's mostly related to how things are handled in the main script.

By checkpointing after eval, one can resume a checkpoint by just changing the initial experience like this:

```
for exp in train_steam[resumed_from:]:
  train()
  eval()
```

by checkpointing after train, you will end up with something like this:

```
resumed_from_checkpoint = True/False

for exp in train_steam[resumed_from-1:]:
  if not resumed_from_chechpoint:
    train()
  eval()
  resumed_from_checkpoint = False
```
(actually, we can simplify `resumed_from-1` by re-loading an already `-1`ed value.)

By checkpoint on both, things are actually not bad:

```
resume_from = 'eval/train' 
# resumed_from_checkpoint is not needed as we can just set resume_from='train' 
# same for the value of resumed_from, which would be adjusted accordingly

for exp in train_steam[resumed_from:]:
  if resume_from == 'train':
    train()
  eval()
```

AntonioCarta: The assumption that you have an `eval` call is not necessarily true. If I call `strategy.train(exp, eval_streams=[valid_stream])` I may not need an additional eval since it's already inside the training loop. In fact, I could even have a single train call `strategy.train(train_stream, eval_streams=[train_stream, valid_stream])`.

However, we can safely do multiple `eval` calls. So, I would do something like this:

```
if resume_from_checkpoint:
    # load checkpoint here
    # also do eval, because we don't know if it was computed
    # this may not needed if the periodiceval does the final evaluation
    strategy.eval(valid_stream)

# we may or may not have a loop or an eval call here 
strategy.train(train_stream[resumed_from:], eval_streams=[train_stream, valid_stream])
```

lrzpellegrini:
Apart from minor git merge issues, the checkpointing functionality seems to be working with the new datasets!

- I also tested other plugins such as LwF, GDumb, EWC, CWR (only replay is tested by the CI)
- Simplified the test procedure to run on the usual fake benchmark.

AntonioCarta:
@lrzpellegrini can I merge the PR?

lrzpellegrini:
@AntonioCarta yes, it's ready!

PR #1142, opened 2022-10-06
 Optional task and class labels in AvalancheDataset #1118 

### LINKED ISSUE #422: Fallback mechanism when "targets" is not found
At the moment creating a custom benchmark using Avalanche (for instance when using `nc_scenario` or `ni_scenario`) means passing one or more "supported dataset"s to these generators.

A supported dataset is just a PyTorch `Dataset` in which a `targets` field is set. In turn, the `targets` field should be a list of integers containing the class label for each pattern. We need this field because splitting the original dataset in a sequence of experiences usually requires considering the class label for all patterns. In the near future this field will also be used to contain "y" values in general (also for regression, segmentation, detection, ...).

The `targets` field is usually available in all torchvision datasets and is kind of an unofficial standard.

However, there are situations in which having this field is not very beneficial. For instance, if a dataset is already loaded in memory, using the `__getitem__` method, which also loads the "x" value, is sufficiently efficient.

Implementing a mechanism that, after unsuccessfully looking for a `targets` field, falls back to using `__getitem__`, may make things simpler to the user. Of course a proper warning should be shown when Avalanche doesn't find the `targets` field.

### LINKED ISSUE #464: Remove task labels for SIT
Can we set task labels to `None` or some other specific value instead of 0 for single task scenarios? Setting to None has the advantage that the evaluation and loggers can recognize the scenario and adapt the names and output by removing the task label, making the scenario more explicit and the output format easier to read.

Of course we should confirm that Avalanche components work in both scenarios. I think task labels are only used as indexes, therefore having `None` instead of `0` should not make a big difference.

AntonioCarta:
AvalancheDatasets are doing many different things together:
- advanced indexing
- management of class/task labels
- management of transformations
- efficient concat/subset operations needed for replay

This PR separates all these features into different components. The advantage is that now we can do naturally some things that were a bit clunky such as:
- having datasets without class/task labels
- easily add new dataset types without having to worry about compatibility with task labels/collate_fn and other details of supervised datasets.

## New API
transformations:
- all transformations managed by TransformationGroups must be MultiParameterTransformations

MultiParameterTransformations
- are Callable
- number of arguments determine input args

## Deprecated

**AvalancheDataset**:
- dataset types removed.
- `AvalancheDataset` is now a generic class. The old `AvalancheDataset` has been renamed `SimpleClassificationDataset`.
- Similarly, `AvalancheSubset`, `AvalancheConcat` have been renamed `` and `concat_classification_datasets`. These will be removed in the future. Instead, you should use the new `subset` and `concat` operations of the `AvalancheDataset`.

**Transformation Groups**:
- `add_transform`  has been removed.
- `add_transforms_group` removed.
- `add_transforms_to_group` removed.
- `greeze_group_transforms` removed.
- `replace_transforms` renamed to `replace_current_transform_group`.
- `freeze_group_transforms` removed.

**New Behavior**:
- `freeze_transforms` freezes *all* the transformation groups
- added `remove_current_transform_group`

## Performance
performance in general is improved. @AndreaCossu notice that `online er strategy ` has a different number of iterations in the master branch compared to 0.0.2. This probably affects OCL baselines such as CoPE

- NEW `python .\profiling\replay_buffers.py` (ParametricBuffer): 
	- NEW `ParametricBuffer (random sampling) Duration:  10.91 11693it [00:10, 1071.18it/s]`
	- OLD `ParametricBuffer (random sampling) Duration:  94.7800874710083`
- `python .\profiling\online_strategy.py`
	- NEW 
		- `online naive strategy without Avalanche in 41.387 seconds 13007/13007 [00:41<00:00, 310.13it/s]`
		- `online naive strategy  using lazy streams (no AVL) in 36.821 seconds 13007it [00:37, 348.78it/s]`
		- `online naive strategy using Avalanche in 49.155 seconds 13007/13010 [00:49<00:00, 264.65it/s]`
		- `online er strategy using Avalanche in 525.604 seconds, 129560it [08:45, 246.50it/s]`
	- OLD 
		- `online naive strategy without Avalanche in 36.789 seconds [00:37<00:00, 316.17it/s]`
		- `online naive strategy  using lazy streams (no AVL) in 36.147 seconds`
		- `online naive strategy using Avalanche in 49.385 seconds [00:49<00:00, 238.35it/s]`
		- `online er strategy using Avalanche in 86.653 seconds 11769 it. [01:26<00:00, 135.83it/s]`

Closes #422 
Closes #464


HamedHemati:
Is the duration for online ER (New) 525.604 or 52.5604? That seems to be a typo considering the number of iterations.
It seems like the DS concat operation is faster now because as far as I can remember, the ds concatenation part was kind of a bottleneck in online ER.

AntonioCarta:
> Is the duration for online ER (New) 525.604 or 52.5604? That seems to be a typo considering the number of iterations.

Unfortunately the numbers are correct. The higher cost is due to the larger number of iterations compared to 0.0.2, which is unrelated to this PR.

> It seems like the DS concat operation is faster now because as far as I can remember, the ds concatenation part was kind of a bottleneck in online ER.

Yes, concat is not a bottleneck anymore and OCL is so fast to be negligible compared to forward/backward steps. There may be some edge cases where merging (the internal dataset optimization) fails and operations become much slower but I didn't find any for the moment.

PR #1161, opened 2022-10-21
Add ParamData to manage expansion of data associated to model parameters

AndreaCossu:
This PR proposes `ParamData` to manage data like importances (in EWC, SI etc) or any metadata associated to the model parameters (like in SI). 
The main feature is the expansion of the parameter, currently supported only on existing dimensions and only one dimension at a time. 
As a proof of concept, I adjusted SI (which did not work with layer shapes expanding over time) by using ParamData.

Any feedback is appreciated. I can try to adapt the other strategies as well or just fix SI with an ad-hoc change.


AntonioCarta:
Thanks!
- can you fix the other importance strategies? EWC, RWalk (maybe others)?
- this PR adds multi-head support for these strategies. Can you add the MT scenario to the tests for those strategies? We removed it because it was still unsupported up to now.

AndreaCossu:
@AntonioCarta tests should be ready and I adapted all the strategies that crashed in the continual-learning-baselines repo. We can merge this.

PR #1196, opened 2022-11-14
Make AR1 closer to the original implementation

BenCrulis:
This PR fixes AR1 freezing the whole model by default and allows to choose the incremental learning rate, making the implementation closer to the original at https://github.com/vlomonaco/ar1-pytorch

With these fixes I am able to nearly replicate the results of the original github on CORe50.

### DIFF
--- a/avalanche/training/supervised/ar1.py
+++ b/avalanche/training/supervised/ar1.py
@@ -53,7 +53,7 @@ class AR1(SupervisedTemplate):
         max_d_max=0.5,
         inc_step=4.1e-05,
         rm_sz: int = 1500,
-        freeze_below_layer: str = "lat_features.19.bn.beta",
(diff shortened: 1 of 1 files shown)


AntonioCarta:
@lrzpellegrini it seems that checkpointing tests are failing here. Any idea why?

BenCrulis:
@AntonioCarta probably because of the added parameter (`inc_lr`) that isn't in the pickled objects I would say, though I'm not sure.

AntonioCarta:
After discussing this with @lrzpellegrini, it appears that some changes in Avalanche dependencies broke the checkpointing. Now, since the errors are unrelated to this PR (you can see python 3.10 is working) I'm going to merge this anyway. @lrzpellegrini is working to fix checkpointing.

PR #1198, opened 2022-11-14
Add "interactive" mode to ReplayPlugin

HamedHemati:
This PR adds an additional option to ReplayPlugin that allows sample concatenation from the buffer before every training iteration. It can be used for implementing some online strategies.

*As for the online replay strategy, the issue of "significant slowdown" is still not solved. I noticed that the bottleneck is caused by buffer updates after every experience, and it only happens from the second experience. All other steps seem to work as they should. I also inspected the buffer update steps (for reservoir sampling),  and everything looked normal.

### DIFF
--- a/avalanche/training/plugins/replay.py
+++ b/avalanche/training/plugins/replay.py
@@ -1,6 +1,7 @@
 from typing import Optional, TYPE_CHECKING
+import torch
+from torch.utils.data import DataLoader
 
-from avalanche.benchmarks.utils import concat_classification_datasets
 from avalanche.benchmarks.utils.data_loader import ReplayDataLoader
 from avalanche.training.plugins.strategy_plugin import SupervisedPlugin
 from avalanche.training.storage_policy import (
@@ -42,6 +43,8 @@ class ReplayPlugin(SupervisedPlugin):
             buffer samples.
     :param storage_policy: The policy that controls how to add new exemplars
                            in memory
+    :param interactive: if True, buffer samples are retrieved and concatenated
+                        per min-batch before training iteration.
     """
 
     def __init__(
@@ -51,12 +54,14 @@ class ReplayPlugin(SupervisedPlugin):
         batch_size_mem: int = None,
         task_balanced_dataloader: bool = False,
         storage_policy: Optional["ExemplarsBuffer"] = None,
+        interactive: bool = False,
     ):
         super().__init__()
         self.mem_size = mem_size
         self.batch_size = batch_size
         self.batch_size_mem = batch_size_mem
         self.task_balanced_dataloader = task_balanced_dataloader
+        self.interactive = interactive
 
         if storage_policy is not None:  # Use other storage policy
             self.storage_policy = storage_policy
@@ -66,9 +71,34 @@ class ReplayPlugin(SupervisedPlugin):
                 max_size=self.mem_size, adaptive_size=True
             )
 
+        if self.interactive:
+            assert batch_size_mem is not None
+
     @property
     def ext_mem(self):
         return self.storage_policy.buffer_groups  # a Dict<task_id, Dataset>
+
+    def before_training_iteration(
+        self, strategy: "SupervisedTemplate", *args, **kwargs
+    ):
+        if not self.interactive:
+            return
+
+        if len(self.storage_policy.buffer) == 0:
+            return
+
+        selected_items = next(self.itr_replay)
+
+        # Buffer mask
+        buffer_mask = [0] * strategy.mbatch[0].shape[0] + \
+                      [1] * len(selected_items[0])
+        strategy.buffer_mask = torch.LongTensor(buffer_mask)
+
+        # Combine buffer and main loader items
+        for i in range(len(strategy.mbatch)):
+            strategy.mbatch[i] = torch.cat(
+                (strategy.mbatch[i], selected_items[i].to(strategy.device)),
+                dim=0)
 
     def before_training_exp(
         self,
@@ -84,6 +114,19 @@ class ReplayPlugin(SupervisedPlugin):
         if len(self.storage_policy.buffer) == 0:
             # first experience. We don't use the buffer, no need to change
             # the dataloader.
+            return
+
+        # If buffer mode is interactive, initialize replay iterator
+        if self.interactive:
+            rep_loader = DataLoader(
+                self.storage_policy.buffer,
+                batch_size=self.batch_size_mem,
+                shuffle=True,
+                drop_last=True,
+                num_workers=num_workers
+            )
+
+            self.itr_replay = iter(rep_loader)
             return
 
         batch_size = self.batch_size
--- a/examples/online_replay.py
+++ b/examples/online_replay.py
@@ -100,7 +100,8 @@ def main(args):
     # CREATE THE STRATEGY INSTANCE (ONLINE-REPLAY)
     storage_policy = ReservoirSamplingBuffer(max_size=100)
     replay_plugin = ReplayPlugin(
-        mem_size=100, batch_size=1, storage_policy=storage_policy
+        mem_size=100, batch_size_mem=1, storage_policy=storage_policy,
+        interactive=True
     )
 
     cl_strategy = OnlineNaive(
@@ -108,7 +109,7 @@ def main(args):
         torch.optim.Adam(model.parameters(), lr=0.1),
         CrossEntropyLoss(),
         train_passes=1,
-        train_mb_size=1,
+        train_mb_size=10,
         eval_mb_size=32,
         device=device,
         evaluator=eval_plugin,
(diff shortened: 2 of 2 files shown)


AntonioCarta:
Maybe I'm missing something here. I thought that in the interactive mode you wanted to avoid the creation of the `DataLoader`. What is the different between interactive and non-interactive?

AntonioCarta:
`self.buffer([data[idx] for idx in idxs])`

AlbinSou:
@HamedHemati If I may add some suggestions about the desired behavior of an interactive buffer plugin

- I am not sure that the important thing is to make it able to add samples before each training iteration for the following reasons. Generally, it is desired to add samples before an iteration when performing experiments in the Online Setting. However, in avalanche the online scenarios are implemented such that before_training_exp and after_training_exp are called whenever a new batch of data is received. Of course before_training_iteration will also be called, so in theory it should work regardless. But it could be problematic when performing several iterations on the same batch, in what case the same image could be added to the buffer several time.

- I think overall if the user uses a proper OnlineScenario, the classical replay buffer handles well the fact of adding new samples from each batch of data. What is problematic as you and antonio also noticed is the reinstantiation of the DataLoader. So I guess we need to find a solution to this problem in particular.

AlbinSou:
I looked a bit into this today and so far these are my conclusions. It is possible to change the content of the dataset used by a loader as long as it's done inplace (i.e loader.dataset = new_dataset raises an error anyway). However, even when doing so, it works only for num_workers = 0, otherwise we will have to wait for the next epoch to have the updates available.

Another problem I foresee is that, even if that fixes the problem of resetting the replay loader, it does not fix the one of resetting the Strategy Loader (which is also reset at every iteration). So I don't think the performance gains would be amazing.

```
  def _before_training_exp(self, **kwargs):
        """Setup to train on a single experience."""
        # Data Adaptation (e.g. add new samples/data augmentation)
        self._before_train_dataset_adaptation(**kwargs)
        self.train_dataset_adaptation(**kwargs)
        self._after_train_dataset_adaptation(**kwargs)

        self.make_train_dataloader(**kwargs) <-- We recreate the dataloader at each batch

        # Model Adaptation (e.g. freeze/add new units)
        # self.model = self.model_adaptation()
        # self.make_optimizer()
        self.check_model_and_optimizer()

        super()._before_training_exp(**kwargs) <-- Replay dataloader creation is done
```

So I guess to solve this slowness problem, an interactive buffer plugin will not be sufficient, we also need to change the way the loader is handled in the Online scenario ... Which can be problematic. Honestly I like the way it's implemented right now with one batch being one experiment. Maybe we can try doing something like having one dataloader that is initialized at the beginning of the training and changing its dataset in place instead of resetting it. But then we run into the num_workers > 0 problem. There is also the risk that the replacing of the dataset takes more or less the same time than the resetting of the dataloader, I don't really know what is the part that takes time in the dataloader reset.

 @AntonioCarta do you think this reseting of the loader is the biggest part of the slow down of these online scenarios or not ? Maybe it's worth checking before making any changes.

AntonioCarta:
Ok, I was finally able to isolate the slowdown. Here is a script to reproduce it:
```
from unittest.mock import Mock

import time

from tqdm import tqdm

from avalanche.benchmarks import fixed_size_experience_split, SplitMNIST, split_online_stream
from avalanche.benchmarks.utils.flat_data import _flatdata_depth
from avalanche.training import ReservoirSamplingBuffer
from avalanche.training import ParametricBuffer

benchmark = SplitMNIST(
    n_experiences=5,
    dataset_root=expanduser("~") + "/.avalanche/data/mnist/",
)

experience = benchmark.train_stream[0]
print("len experience: ", len(experience.dataset))

start = time.time()
buffer = ReservoirSamplingBuffer(100)
for exp in tqdm(split_online_stream(benchmark.train_stream[:2], 1)):
    buffer.update_from_dataset(exp.dataset)

end = time.time()
duration = end - start
print("ReservoirSampling Duration: ", duration)

```

The first experience is very first but the second one slows down (a lot). I need some time to investigate this because it's probably an issue with the concat/sub operations of the dataset.

AntonioCarta:
@HamedHemati @AlbinSou take a look at #1217 
It's a bit speedup but still slower than it should be.

AntonioCarta:
Update: I fixed the performance issue. Now it is faster. Once it's merged, can you check it to see if it's good enough for your use cases?

```
(split MNIST OCL with exp-size=5 samples, 6000 iterations total)
6003it [00:15, 389.10it/s]
ReservoirSampling Duration:  15.429586410522461
6003it [00:29, 205.35it/s]
ParametricBuffer (random sampling) Duration:  29.233972787857056
```

AlbinSou:
> Update: I fixed the performance issue. Now it is faster. Once it's merged, can you check it to see if it's good enough for your use cases?
> 
> ```
> (split MNIST OCL with exp-size=5 samples, 6000 iterations total)
> 6003it [00:15, 389.10it/s]
> ReservoirSampling Duration:  15.429586410522461
> 6003it [00:29, 205.35it/s]
> ParametricBuffer (random sampling) Duration:  29.233972787857056
> ```

Hey Antonio, thanks for the changes. I will check once it's merged and update you.

AntonioCarta:
I merged the changes. Assuming they do work and we have the expected speedup, do we still need this PR? what are the features that distinguish "interactive" mode from the non-interactive?

HamedHemati:
> @HamedHemati If I may add some suggestions about the desired behavior of an interactive buffer plugin
> 
> * I am not sure that the important thing is to make it able to add samples before each training iteration for the following reasons. Generally, it is desired to add samples before an iteration when performing experiments in the Online Setting. However, in avalanche the online scenarios are implemented such that before_training_exp and after_training_exp are called whenever a new batch of data is received. Of course before_training_iteration will also be called, so in theory it should work regardless. But it could be problematic when performing several iterations on the same batch, in what case the same image could be added to the buffer several time.

Thanks @AlbinSou, for the suggestions. The idea is to add modes, even "before_forward"  because there are some strategies that require a different batch of buffer samples for every sample in a mini-experience (an OCL experience). One example is MER (image below), where new buffer samples are retrieved for each sample in the experience:

Screenshot 2022-11-24 at 15 12 21

Also, we need a clear way to distinguish between experience samples and buffer samples, something like a mask, in case we need to treat the samples differently.
Even if some online strategies can be implemented by the current version of ReplayPlugin, I think making it more flexible will make it easier to implement any arbitrary ER-based online strategy.

AntonioCarta:
>  Even if some online strategies can be implemented by the current version of ReplayPlugin, I think making it more flexible will make it easier to implement any arbitrary ER-based online strategy.

Can you give an example of a strategy that **cannot** be implemented with the current plugin? I can't imagine any right now.

> Also, we need a clear way to distinguish between experience samples and buffer samples, something like a mask, in case we need to treat the samples differently.

I agree with this. I'm not sure what is the best way to do it but it's something necessary for some methods. I'm not sure what is the relationship between this point and the interactive mode?

HamedHemati:
> Can you give an example of a strategy that **cannot** be implemented with the current plugin? I can't imagine any right now.

The pseudocode that I attached is an example, but let me give a concrete example:
Let's assume the OCLExperience size is 10, and then you iterate through each sample one by one. For each sample, you need to retrieve 10 random buffer samples (for inner updates in the meta-learning step). To implement this with the ReplayPlugin, you need to set the buffer batch size to `10x10=100`, to reserve all needed buffer samples for all sample iterations before the experience starts. Even if this is fine, the implementation would be more difficult because you need to get access to the index of each sample to slice the buffer samples for each experience sample. Another issue would be the buffer size. For example, a buffer size of 50 would not be possible in this case because you need to get at retrieve 100 samples for each (OCL) experience. Such a limitation should not be a part of the approach afaik.
 

> I agree with this. I'm not sure what is the best way to do it but it's something necessary for some methods. I'm not sure what is the relationship between this point and the interactive mode?

Well, it is related. The current ReplayBuffer does not guarantee that the batch size for the experience and the buffer samples will always be the same because one of the loaders may finish earlier with having fewer samples in the final batch of the dataloader.  Here I'm also talking about batch-based strategies. If you retrieve and concatenate samples before each iteration, then it will be just easier to manage both experience and buffer samples. 

Nevertheless, if you think we can implement this in a different way that fits better, feel free to close this PR so that we can discuss it elsewhere. This was only a suggestion to extend the ReplayBuffer.

AntonioCarta:
Ok, se let me clarify my position. The ReplayPlugin is not supposed to implement all the existing rehearsal strategies. For example, iCaRL and GEM also use a rehearsal buffer but their implementation is not based on the ReplayPlugin because its usage in those strategies is very different. In the case of MER, we have a similar problem because of the interaction between the buffer and the internal loop of the meta-learning.

I suggest to continue to implement these custom rehearsal methods separately from the ReplayPlugin. If we find a strong need for unification we can provide a uniform API once it's clear what we need. Of course, while MER will not be based on the ReplayPlugin it can still reuse Avalanche storage policies.

For the moment, I'm closing this.

PR #1201, opened 2022-11-16
Maximally Interfered Retrieval plugin

AlbinSou:
I tried to implement the ER-MIR strategy described in [Online Continual Learning with Maximally Interfered Retrieval](https://arxiv.org/abs/1908.04742) as a plugin.

The implementation for now is quite slow because of the re-instantiation of the DataLoader at every iteration. I will try to come up with a different solution but right now this is the best-working one that I have so far, so I'll leave this as a basis to improve on.

I also include a replay plugin that works similarly to the MIR one (sample two batches and add the loss coming out). It's just here so that I can compare the two methods fairly. 

I thought it would be hard to implement MIR as classical avalanche replay (concatenating replay batch to current batch), since in MIR you have to perform a temporary update with the current batch before knowing what samples to use replay on. So that's why I have to compare to different replay plugin.

I also added a test script with which I try to get as close as the MIR setting described for CIFAR10, that we can use to check the results. Right now I ran only for one seed and had +4% for MIR. But I know a few details are not present in the example script, for instance I don't think they use input transformations on CIFAR10

Feel free to comment and propose improvements. I think the main idea of the method is here but few optimizations can be done still.


avalanche/training/plugins/mir.py
```diff
@@ -0,0 +1,190 @@
...
+                    buffer, batch_size=self.batch_size_mem, shuffle=True
+                )
+            )
+        else:
+            self.replay_loader = None
+
+
+class ReplayLikeMIR(SupervisedPlugin):
+    """
+    Replay plugin but working like MIR for fair comparison
```
AntonioCarta, line 135: Can you describe between the default Avalanche replay and this one here in the docstring?

AntonioCarta (changes requested):
Hi @AlbinSou, thanks for the PR. I need to ask you for two things:
- you have to add a test in `tests/training/test_strategies`. Just copy-paste the one of the other strategies and adapt it for MIR.
- the example in `examples/mir_cifar10` will go in the [continual-learning-baselines](github.com/continualAI/continual-learning-baselines) repository. Can you open a PR there once this is merged?

The implementation looks good to me. Regarding the slow OCL scenario, we are investigating it with @HamedHemati. Hopefully, we will have a solution soon enough.

avalanche/training/plugins/mir.py
AntonioCarta: @HamedHemati good point. A good tradeoff to avoid refactoring the implementation could be to move the "MIRLikeReplay" to the continual-learning-baselines repository as an OCL baseline, while avalanche will keep the new ReplayPlugin. What do you think?

AlbinSou: Hey Antonio, yes sure this MIRLikeReplay is just here for the purpose of the comparison so it has nothing to to in avalanche. I will move it to the PR in continual-baseline.

AlbinSou: I added the test for now. Just to warn you, I tried to stick closer to the setting by suppressing input rotation and cropping, but this only made things worst .. Now I got 34% for MIR and 37% for replay baselines. I don't know what is wrong. I think for the most part it behaves right (after task 1 2 3 it's above normal replay but last task it drops). Maybe it's just some training artifact that I have to fix. Anyway, I guess later discussions about getting the exact same results than in the paper will happen in the continual-learning-baseline PR.

AntonioCarta: Great! as soon as you fix the pep8 errors we can merge the PR. You can check them locally with the command `pycodestyle avalanche/ examples/ tests/ --exclude examples/tvdetection`

HamedHemati: @AntonioCarta Sorry, I missed the reply to my comment. My point was to implement both `MIRPlugin` and `ReplayLikeMIR`  using the `ReplayPlugin` from Avalanche. As for `MIRLikeReplay`, it should be just a script to replicate results for that strategy by setting the correct arguments, and for `MIRPlugin`, `self.replay_loader` can be replaced with Avalanche's `ReplayPlugin`.

AlbinSou: @HamedHemati The thing that I had in mind was to implement it like that for now and later change it once the interactive buffer system is on the main branch. I don't know if this answers your question. As for ReplayLikeMIR I think it makes more sense to put it only on the continual-baseline repo since it is not very different from normal replay, I use it more to try and get close to the ER baseline the authors use in their code.

AntonioCarta:
@AlbinSou tests are still failing with this error:
```
ImportError: cannot import name 'ReplayLikeMIR' from 'avalanche.training.plugins.mir' (/__w/avalanche/avalanche/avalanche/training/plugins/mir.py)
```
probably because you removed it.

Can you also pull the master and fix the conflicts so that we can merge the PR?

AntonioCarta:
Thanks @AlbinSou! Please remember to open the PR for the continual-learning-baselines repo.

PR #1217, opened 2022-11-23
speed up dataset concat in OCL scenarios

AntonioCarta:
improve the flattening process by recognizing when the same dataset occurs multiple times in the dataset list:
- added `_flatten_and_reindex` to improve flattening
- added a `__hash__` method for `AvalancheDataset` to recognize multiple dataset occurrences

@lrzpellegrini can you check this?

**Update**: fixed it. Now it's fast for the entire stream:
```
(split MNIST OCL with exp-size=5 samples, 6000 iterations total)
6003it [00:15, 389.10it/s]
ReservoirSampling Duration:  15.429586410522461
6003it [00:29, 205.35it/s]
ParametricBuffer (random sampling) Duration:  29.233972787857056
```

### DIFF
--- a/avalanche/benchmarks/utils/data.py
+++ b/avalanche/benchmarks/utils/data.py
@@ -125,6 +125,13 @@ class AvalancheDataset(FlatData):
             self._data_attributes = {}
         else:
             self._data_attributes = {da.name: da for da in data_attributes}
+            for da in data_attributes:
+                ld = sum(len(d) for d in self._datasets)
+                if len(da) != ld:
+                    raise ValueError(
+                        "Data attribute {} has length {} but the dataset "
+                        "has length {}".format(da.name, len(da), ld)
+                    )
         self._transform_groups = transform_groups
         self._frozen_transform_groups = frozen_transform_groups
         self.collate_fn = collate_fn
--- a/avalanche/benchmarks/utils/data.py
+++ b/avalanche/benchmarks/utils/data.py
@@ -194,13 +194,16 @@ class AvalancheDataset(FlatData):
             for da in self._data_attributes.values():
                 # TODO: this was the old behavior. How do we know what to do if
                 # we permute the entire dataset?
-                if len(da) != sum([len(d) for d in datasets]):
-                    self._data_attributes[da.name] = da
-                else:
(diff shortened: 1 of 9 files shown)


tests/benchmarks/test_flat_data.py
```diff
@@ -88,3 +96,127 @@ def test_merging(self):
...
+        print(f"DATA depth={_flatdata_depth(B)}, dsets={len(B._datasets)}")
+        assert _flatdata_depth(B) == 2
+        assert len(B._datasets) == 1
+
+        B = D.concat(B)
+        print(f"DATA depth={_flatdata_depth(B)}, dsets={len(B._datasets)}")
+        assert _flatdata_depth(B) == 2
+        assert len(B._datasets) == 1
+
+    def test_concat_flattens_nc_scenario_dataset(self):
```
lrzpellegrini, line 150: There are no asserts here

avalanche/benchmarks/utils/data.py
```diff
@@ -194,13 +194,17 @@ def __init__(
...
                 # TODO: this was the old behavior. How do we know what to do if
                 # we permute the entire dataset?
-                if len(da) != sum([len(d) for d in datasets]):
-                    self._data_attributes[da.name] = da
-                else:
-                    self._data_attributes[da.name] = da.subset(self._indices)
-
-                    dasub = da.subset(indices)
-                    self._data_attributes[da.name] = dasub
+                # DEPRECATED! always subset attributes
```
lrzpellegrini, line 197: It may make sense to add a unit test to cover this change.

lrzpellegrini:
The PR makes sense to me. Correct me if I'm wrong, but will it be able to flatten most replay buffers into a single dataset (apart from benchmarks created with dataset_benchmarks)? 

I only left a couple of comments, everything else seems ok to me!

AntonioCarta:
> Correct me if I'm wrong, but will it be able to flatten most replay buffers into a single dataset (apart from benchmarks created with dataset_benchmarks)?

yes! The details depend on the benchmark creation. For example, Permuted MNIST experiences are considered separate datasets because of the transformations.

AntonioCarta:
Thanks @lrzpellegrini I added the tests and an error to catch the backward incompatible behavior.

PR #1225, opened 2022-11-25
Issue Fix for OCL

HamedHemati:
- FIX an issue with eval stream initialization
- UPDATE interactive logger for online CL strategies

### DIFF
--- a/avalanche/training/templates/base.py
+++ b/avalanche/training/templates/base.py
@@ -106,7 +106,11 @@ class BaseTemplate:
             experiences = [experiences]
         if eval_streams is None:
             eval_streams = [experiences]
-        self._eval_streams = _group_experiences_by_stream(eval_streams)
+
+        if len(eval_streams) == 1:
+            self._eval_streams = eval_streams
+        else:
+            self._eval_streams = _group_experiences_by_stream(eval_streams)
 
         self._before_training(**kwargs)
 
--- a/avalanche/training/templates/base.py
+++ b/avalanche/training/templates/base.py
@@ -107,10 +107,7 @@ class BaseTemplate:
         if eval_streams is None:
             eval_streams = [experiences]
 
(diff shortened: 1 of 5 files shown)


avalanche/training/templates/base.py
```diff
@@ -106,7 +106,11 @@ def train(
             experiences = [experiences]
         if eval_streams is None:
             eval_streams = [experiences]
-        self._eval_streams = _group_experiences_by_stream(eval_streams)
+
+        if len(eval_streams) == 1:
+            self._eval_streams = eval_streams
+        else:
```
AntonioCarta, line 112: can you put this inside`_group_experiences_by_stream`?

AntonioCarta:
can you add some tests to catch the error?

avalanche/training/templates/base.py
HamedHemati: yea, sure.

HamedHemati:
> can you add some tests to catch the error?

What do you mean by adding tests? is there something that needs to be caught?

AntonioCarta:
> What do you mean by adding tests? is there something that needs to be caught?

We need a unit test that fails if the bug with the OCL stream occurs again because of some changes.

HamedHemati:
> We need a unit test that fails if the bug with the OCL stream occurs again because of some changes.

The online strategy test is updated. Now, it checks `assert cl_strategy.clock.train_exp_counter > 0` after training on the whole OCL stream.

AntonioCarta:
Sorry, I didn't explain myself. This PR is fixing a bug:

> FIX an issue with eval stream initialization

If we want to be sure the bug doesn't happen again, we need a unit test that checks this problem, i.e. passes an OCL eval stream and checks that the periodic eval groups the streams correctly in all the possible cases.

PR #1227, opened 2022-11-26
MER Strategy + OnlineSupervisedMetaLearning Template

HamedHemati:
This PR adds:
- `OnlineSupervisedMetaLearningTemplate`: added to the list of common templates
- MER Strategy: an online meta-learning-based strategy (https://arxiv.org/abs/1810.11910)

The algorithm is adapted from version 1 of the algorithm in the official repository:
https://github.com/mattriemer/MER/blob/master/model/meralg1.py

Unfortunately, the results are only available for MNIST and Omniglot in the original paper, and the rest are RL experiments. In the first run, I tested it `SimpleMLP` from Avalanche on PMNIST, and the results were a little bit higher compared to what is reported in the paper (~0.76 vs. ~0.73), which I guess should be related to the higher capacity of the default MLP model in Avalanche. I can also run it for the original architecture, but since the algorithm is super slow due to the multiple levels of meta-updates in each iteration, I am going to add them later.


avalanche/training/templates/common_templates.py
```diff
@@ -338,3 +338,116 @@ def __init__(
...
+                eval_dataset_adaptation
+                make_eval_dataloader
+                eval_epoch  # for each epoch
+                    # forward
+                    # backward
+                    # model update
+
+    """
+
+    PLUGIN_CLASS = BaseSGDPlugin
```
AntonioCarta, line 391: Should be `SupervisedPlugin`

AntonioCarta:
For the reproducibility we can discuss this on the continual-learning-baselines repo. If we can have the original architecture it's better.

avalanche/training/templates/common_templates.py
HamedHemati: Fixed for all!

PR #1230, opened 2022-11-30
Adding ER-ACE strategy

AlbinSou:
Started to implement the ER-ACE strategy from the "New Insights on Reducing Abrupt Representation Change in Online Continual Learning" paper [openreview link](https://openreview.net/forum?id=N8MaByOzUfb).

The implementation is still not 100% ready but I need your input on a few things. 

First of all, I have created two strategies. One for the online setting (using Online Scenario), and one for the non online setting. It felt a bit bad to do this since they are basically doing nearly the same thing (the only difference is how the replay buffer is updated), but I don't really know how to do things differently. Do you think it's mandatory to have two different strategies for the two different settings (as it is done for Naive and OnlineNaive) ? If you have a better Idea of how to split these, let me know.

Secondly, the test I created is not passing because the scenario that it uses does not set classes_start_from_zero_from_first_exp=True. I think the only place I need to modify is in the er_acl_criterion function, 
```{python}
def er_acl_criterion(
    out_in, target_in, out_buffer, target_buffer, all_classes, new_classes
):
    loss_buffer = F.cross_entropy(out_buffer, target_buffer)
    loss_current = F.cross_entropy(
        out_in[:, len(all_classes) - len(new_classes) :],
        target_in - (len(all_classes) - len(new_classes)),
    )
    return loss_buffer + loss_current
```
where right now I retire an offset (len(all_classes) - len(new_classes) to the target, and it fails when the target is not bigger than the value of len(classes_seen_so_far). I'm still thinking about a fix for this but again, if you have an idea let me know.

### DIFF
--- a/avalanche/training/regularization.py
+++ b/avalanche/training/regularization.py
@@ -189,7 +189,9 @@ class ACECriterion(RegularizationMethod):
         loss_buffer = F.cross_entropy(out_buffer, target_buffer)
         oh_target_in = F.one_hot(target_in, num_classes=out_in.shape[1])
         oh_target_in = oh_target_in[:, list(self.new_classes)]
-        loss_current = cross_entropy_with_oh_targets(out_in[:, list(self.new_classes)], oh_target_in)
+        loss_current = cross_entropy_with_oh_targets(
+                out_in[:, list(self.new_classes)], oh_target_in
+        )
         return (loss_buffer + loss_current) / 2
 
     @property
--- a/avalanche/training/regularization.py
+++ b/avalanche/training/regularization.py
@@ -9,6 +9,15 @@ import torch.nn.functional as F
 from avalanche.models import MultiTaskModule, avalanche_forward
 
 
+def cross_entropy_with_oh_targets(outputs, targets, eps=1e-5):
+    """ Calculates cross-entropy with temperature scaling, 
+    targets can also be soft targets but they must sum to 1 """
+    outputs = torch.nn.functional.softmax(outputs, dim=1)
+    ce = -(targets * outputs.log()).sum(1)
+    ce = ce.mean()
+    return ce
+
+
 class RegularizationMethod:
     """RegularizationMethod implement regularization strategies.
     RegularizationMethod is a callable.
@@ -178,11 +187,9 @@ class ACECriterion(RegularizationMethod):
 
     def __call__(self, out_in, target_in, out_buffer, target_buffer):
         loss_buffer = F.cross_entropy(out_buffer, target_buffer)
-        loss_current = F.cross_entropy(
-            out_in[:, len(self.all_classes) - len(self.new_classes) :],
-            target_in - (len(self.all_classes) - len(self.new_classes)),
-        )
-        # Divide by two for learning rate coherency
+        oh_target_in = F.one_hot(target_in, num_classes=out_in.shape[1])
+        oh_target_in = oh_target_in[:, list(self.new_classes)]
+        loss_current = cross_entropy_with_oh_targets(out_in[:, list(self.new_classes)], oh_target_in)
         return (loss_buffer + loss_current) / 2
 
     @property
--- a/avalanche/training/regularization.py
+++ b/avalanche/training/regularization.py
@@ -1,10 +1,12 @@
 """Regularization methods."""
 import copy
+from collections import defaultdict
+from typing import List
 
 import torch
+import torch.nn.functional as F
 
 from avalanche.models import MultiTaskModule, avalanche_forward
-from collections import defaultdict
 
 
 class RegularizationMethod:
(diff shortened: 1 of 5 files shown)


AntonioCarta:
Can you implement the loss using the [RegularizationMethod](https://github.com/ContinualAI/avalanche/blob/master/avalanche/training/regularization.py) API? This would make it possible to use the ACE loss outside of avalanche training loops. We are trying to make avalanche training modules more usable even for those people that don't use avalanche training strategies.

> Do you think it's mandatory to have two different strategies for the two different settings (as it is done for Naive and OnlineNaive) ? If you have a better Idea of how to split these, let me know.

This is a limitation of the current Avalanche organization of strategies. Unfortunately, there is not much that you can do here, this is a general problem of the Avalanche API.

> where right now I retire an offset (len(all_classes) - len(new_classes) to the target, and it fails when the target is not bigger than the value of len(classes_seen_so_far). I'm still thinking about a fix for this but again, if you have an idea let me know.

we usually use masking to keep track of active units. You can see LwF or the masking in the IncrementalClassifier for an example.

AlbinSou:
> Can you implement the loss using the [RegularizationMethod](https://github.com/ContinualAI/avalanche/blob/master/avalanche/training/regularization.py) API? This would make it possible to use the ACE loss outside of avalanche training loops. We are trying to make avalanche training modules more usable even for those people that don't use avalanche training strategies.
> 
> > Do you think it's mandatory to have two different strategies for the two different settings (as it is done for Naive and OnlineNaive) ? If you have a better Idea of how to split these, let me know.
> 
> This is a limitation of the current Avalanche organization of strategies. Unfortunately, there is not much that you can do here, this is a general problem of the Avalanche API.
> 
> > where right now I retire an offset (len(all_classes) - len(new_classes) to the target, and it fails when the target is not bigger than the value of len(classes_seen_so_far). I'm still thinking about a fix for this but again, if you have an idea let me know.
> 
> we usually use masking to keep track of active units. You can see LwF or the masking in the IncrementalClassifier for an example.

Ok I will look into all those things. So, if I implement it as part of the RegularizationAPI I would just need to use this as a loss in the strategy argument instead of CrossEntropyLoss ? I will try to do it but I hope I can still access the info about what the previous classes and current classes are through this API

AntonioCarta:
> So, if I implement it as part of the RegularizationAPI I would just need to use this as a loss in the strategy argument instead of CrossEntropyLoss ? 

Not really, but this is a good point. The idea is to allow usage of loss functions outside Avalanche. As a result, the `__call__` and `__update__` of  each `RegularizationMethods` have different arguments depending on their needs (instead of relying on the `strategy` state). The disadvantage is that the function calls cannot be updated because the strategy loop doesn't know which argument it needs to use.

> I will try to do it but I hope I can still access the info about what the previous classes and current classes are through this API

In this case, `ACELoss` will need to take `experience` as argument in the `update` method to compute current and previous classes.

AlbinSou:
> > So, if I implement it as part of the RegularizationAPI I would just need to use this as a loss in the strategy argument instead of CrossEntropyLoss ?
> 
> Not really, but this is a good point. The idea is to allow usage of loss functions outside Avalanche. As a result, the `__call__` and `__update__` of each `RegularizationMethods` have different arguments depending on their needs (instead of relying on the `strategy` state). The disadvantage is that the function calls cannot be updated because the strategy loop doesn't know which argument it needs to use. The advantage is that if you don't use avalanche strategies now you know exactly what the loss needs as arguments instead of having to wrap the arguments in a mock `strategy` argument.
> 
> > I will try to do it but I hope I can still access the info about what the previous classes and current classes are through this API
> 
> In this case, `ACELoss` will need to take `experience` as argument in the `update` method to compute current and previous classes.

Hmm, unfortunately, I cannot use active_units for two reasons, first of all (minor reason), right now IncrementalClassifier fails in the incremental scenario because of this classes_in_experience issue. Secondly, if I implement it using the RegularizationMethod interface, then it would require users to enter these active units in the update or __call__ call, which is not necessarly a problem but can get complicated if they are not using an IncrementalClassifier and have to figure out what the structure of this active_units attribute is.

avalanche/training/regularization.py
```diff
@@ -126,4 +128,66 @@ def update(self, experience, model):
...
+        self.old_classes = (
+            set(initial_old_classes) if 
+            initial_old_classes is not None else set()
+        )
+        self.new_classes = (
+            set(initial_new_classes) if 
+            initial_new_classes is not None else set()
+        )
+
+    def update(self, batch_y):
```
AntonioCarta, line 158: update will be called `after_train_exp`, so you can use the `experience` or `dataset` to extract the active units.

AlbinSou:
Ok I think everything is fine. I also have the continual-baselines tests for Split-Cifar10 ready and they work very well. The only thing that I did not do is add a test for the online strategy, do you know where I should put it ? I didn't put it initially because the util for creating a scenario in test_strategies.py does not handle online scenarios.

AntonioCarta:
ACE is still calling update inside `_before_training_iteration` instead of `_before_training_exp`. Can you change it?

AntonioCarta:
> The only thing that I did not do is add a test for the online strategy, do you know where I should put it ?

We don't have separate tests for OCL right now. If there OCL-specific issues in a strategy we can add a test for it but I don't think it's necessary in this case.

AlbinSou:
> > The only thing that I did not do is add a test for the online strategy, do you know where I should put it ?
> 
> We don't have separate tests for OCL right now. If there OCL-specific issues in a strategy we can add a test for it but I don't think it's necessary in this case.

Well, there could be errors due for instance to the fact that for online experience do not have classes_in_experience attribute so it would fail if I try to access it. But I have a test on my side and I made sure it worked in the online scenario.

AlbinSou:
> ACE is still calling update inside `_before_training_iteration` instead of `_before_training_exp`. Can you change it?

I don't see the problem here, would you prefer that I use self.experience.dataset.targets for the update of the class-related knowledge instead of self.mb_y ? I think I switched to using that because I thought I needed self.mb_y. But I can do the same with the self.experience.dataset I guess. The thing is that I cannot use experience.classes_in_this_experience because this one is not available in online scenarios.

AntonioCarta:
Ok, we definitely have to improve the online/batch experiences to make them more consistent.

Thanks @AlbinSou, don't forget to push the reproducibility script to continual-learning-baselines

PR #1251, opened 2022-12-13
Fix MIR potential infinite loop

AlbinSou:
Fixed the infinite loop encountered in some scenarios.

That said, the behaviour of ER-MIR in domain-incremental is a bit weird since the way I implement it (initially thought for class incremental and also done by the authors), is that I remove current task classes from the replay. 

But in domain incremental, all the classes are present all the time so replay should not happen if we follow this logic.

In definitive, the current version is not really adapted to domain incremental. The tests are running for now since the batch size is small enough so that all the classes do not appear in one batch, but it's just a fortunate event, should not be taken for granted for bigger batch sizes.

I also removed a useless function from ER_ACE PR


AntonioCarta:
Thanks.

In general, users should be aware of each algorithm limitations. In the future, we can help them by triggering a warning in case of incompatible benchmark, like in this case.

PR #1253, opened 2022-12-13
Checkpointing notebook and better InteractiveLogger resume

lrzpellegrini:
This adds a notebook guide for the checkpointing functionality and fixes a minor issue in resuming InteractiveLogger when running in notebooks.


lrzpellegrini:
@AntonioCarta I created the notebook as-is, but it did not change anything GitBook side. Do we need to change something on GitBook to make it visible?

AntonioCarta:
You need to create the page on gitbook and add it to the menu. Then, it should be uploaded via the github action. You can check [here](https://github.com/ContinualAI/avalanche/tree/master/docs/gitbook/how-tos) to see if the .md file is there.

PR #1256, opened 2022-12-14
allow multiple eval when using checkpointing

AntonioCarta:
The checkpointing plugins checks that the checkpoint does not exist already. The problem is that this does not allow multiple eval calls.

@lrzpellegrini is there any reason to raise an error in this scenario?

### DIFF
--- a/avalanche/training/plugins/checkpoint.py
+++ b/avalanche/training/plugins/checkpoint.py
@@ -303,10 +303,6 @@ class FileSystemCheckpointStorage(CheckpointStorage):
             checkpoint_name: str,
             checkpoint_writer: Callable[[IO[bytes]], None]):
         checkpoint_file = self._make_checkpoint_file_path(checkpoint_name)
-        if checkpoint_file.exists():
-            raise RuntimeError(
-                f'Checkpoint file {str(checkpoint_file)} already exists.')
-
         checkpoint_file.parent.mkdir(exist_ok=True, parents=True)
 
         try:


lrzpellegrini:
This is mostly a safety measure to prevent users from mistakenly overwriting checkpoints they may not want to actually override (because they forgot to change the checkpoints path, due to carelessness or path configuration errors).

To allow for multiple evals, one can add a flag or similar inside the plugin that can be used to skip that check if the checkpoint file was written in the current session.

AntonioCarta:
> This is mostly a safety measure to prevent users from mistakenly overwriting checkpoints they may not want to actually override (because they forgot to change the checkpoints path, due to carelessness or path configuration errors).

Then isn't it better to check that the folder doesn't exists already when it is created the first time? I think having multiple evals is a basic feature, not something that is optional or that should be dangerous to use.

lrzpellegrini:
Yes, we can check if a folder for the next experience already exists when starting the training phase.

AntonioCarta:
@lrzpellegrini I noticed another problems. dill serializes lambda functions but it seems to fail in some parallel settings. The problem is that I'm not sure how to easily reproduce this issue in a test. Have you ever encountered this problem? I already fixed it (it was the lambda in the `EvaluationPlugin.__init__`) but I would like to add some test to check this hoping we may catch this issue in advance.

lrzpellegrini:
Currently, we have a test that runs a training loop with a bunch of metrics and the replay plugin: https://github.com/ContinualAI/avalanche/blob/master/tests/checkpointing/task_incremental_with_checkpointing.py

The test is called by this bash script: https://github.com/ContinualAI/avalanche/blob/master/tests/checkpointing/test_checkpointing.sh

You should consider adding metrics or other elements that may fail there.

AntonioCarta:
The `EvaluationPlugin` itself is failing, but only in some cases:
- single process execution: everything works correctly.
- parallel execution: serialization fails.

The problem is that the "parallel execution" is done by hydra sweepers, so I'm not sure what it is using, why it is breaking, and how to reproduce it in a minimal example.

lrzpellegrini:
It seems that hydra uses Python multiprocessing. In turn, multiprocessing is probably passing data around using pickle in a quite hard-wired way. I don't know if there is something we can do about it... There is an answer [here](https://stackoverflow.com/questions/19984152/what-can-multiprocessing-and-dill-do-together) of someone that tried to monkey patch the multiprocessing module, but I think it is quite difficult to achieve.

Regarding tests, this seems very multiprocessing-related. Maybe you can extract a minimal training loop and run it through multiprocessing. I don't think it will work using `unittest`, but you can add it to the unit-test action as a separate script.

AntonioCarta:
> I don't know if there is something we can do about it

At the very least we can:
- test whether strategies and plugins work with multiprocessing
- fix them when easily doable

AntonioCarta:
New PR at #1338

PR #1258, opened 2022-12-14
Added optional dependencies test in unit test action

lrzpellegrini:
This adds an additional check in the unit test action to check if optional dependencies are actually not needed to run parts that do not need them.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -55,7 +55,7 @@ jobs:
           python -m unittest discover tests &&
           echo "Checking that optional dependencies are not needed" &&
           pip uninstall -y higher ctrl-benchmark torchaudio gym pycocotools lvis &&
-          python examples/eval_plugin.py &&
+          PYTHONPATH=. python examples/eval_plugin.py &&
           echo "Running checkpointing tests..." &&
           bash ./tests/checkpointing/test_checkpointing.sh &&
           echo "While running unit tests, the following datasets were downloaded:" &&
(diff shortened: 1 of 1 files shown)


lrzpellegrini:
The errors from the CI do not make any sense. It seems that avalanche is not on the PYTHONPATH, but unit tests are executed correctly. I'll try to figure this out...

.github/workflows/unit-test.yml
```diff
@@ -53,8 +53,10 @@ jobs:
           USE_GPU: "False"
         run: |
           python -m unittest discover tests &&
+          echo "Checking that optional dependencies are not needed" &&
```
AntonioCarta, line 56: is it possible to start from a clean environment? otherwise we need to update this test for each new extra dependency

lrzpellegrini:
It seems that "python -m unittest" will set the PYTHONPATH for the tests. This is why setting a PYTHONPATH is not needed usually. Everything seems fine now.

.github/workflows/unit-test.yml
lrzpellegrini: It's quite difficult, as the image used for unit tests already contains all the required dependencies. Una method is to parse [extra_dependencies.txt](https://github.com/ContinualAI/avalanche/blob/master/extra_dependencies.txt), but this would require an ad-hoc script to do it.

AntonioCarta: is it possible to use a different image?

lrzpellegrini: Yes, but that would also require changing the docker image creation actions. It may be a bit overkill.

PR #1285, opened 2023-01-24
Fixed cope plugin

AlbinSou:
Playing around with this plugin to see if a change to ReplayDataLoader would fix it, I noticed the problem was coming from a different cause. self.replay_mem: dict was used in parallel to a normal storage policy but was never updated, I replaced that by using the examplars from the storage policy instead. Running it on the continual-learning-baselines repo for split mnist gives the expected result written in the table of 93%.

Otherwise I think the plugin is not affected by the ReplayDataLoader bug to the exception that the interactive logger shows growing number of iteration for one epoch. But the plugin has a maximum number of iteration parameter.


AntonioCarta:
Thanks, this is great news!

Issue #1274, opened 2023-01-09
ReplayPlugin behaves badly when buffer_size > experience size


AlbinSou:
🐛 **Describe the bug**
This does not make the code fail with an error, but it renders it very slow, and results in a bad behavior that is not the one expected by the user.

When using ReplayPlugin together with an online scenario, replay is activated starting from the second iteration, and thus the training batch will be twice as big starting from this iteration. This is the expected behavior. However, what happens currently is that the strategy trains on the whole buffer additionally, which I don't think is what is suppose to happen.

More precisely, here is the training loop inside SGDUpdate training_epoch:

```{python}
       for self.mbatch in self.dataloader:
           [...]
```

Imagine we set a batch size of 10 and an experience dataset size of 10, with a memory of 1000. We should have len(self.dataloader) = 1 for every seen batch (i.e experience). However, the current behavior is that we have len(self.dataloader) = min(num_batches_seen_so_far, buffer_size/batch_size). So the number of iteration per batch increases, first 1, then 2, 3, 4, 5 .... limited by the size of the buffer. But it basically iterates over the whole buffer every time before moving on to the next batch

As I initially encountered this bug in the online setting (where this effect is highlighted), I reported it using an online scenario (see below). However, I also noticed that in the non-online setting, when the buffer size is bigger than the current experience dataset size, the number of iterations is increased (this time it results in a visible increase of the number of iterations in the interactivelogger)

🐜 **To Reproduce**

```{python}
import numpy as np
import torch
import torchvision.transforms as transforms
from torch.optim import SGD

from avalanche.benchmarks.classic import SplitCIFAR10
from avalanche.benchmarks.generators import benchmark_with_validation_stream
from avalanche.benchmarks.scenarios.online_scenario import OnlineCLScenario
from avalanche.evaluation.metrics import accuracy_metrics, loss_metrics
from avalanche.logging import InteractiveLogger
from avalanche.models import SlimResNet18
from avalanche.models.dynamic_modules import IncrementalClassifier
from avalanche.training.plugins import EvaluationPlugin
from avalanche.training.supervised import OnlineNaive
from experiments.utils import create_default_args, set_seed

def replay_scifar10(override_args=None):
    args = create_default_args(
        {
            "cuda": 0,
            "mem_size": 1000,
            "lr": 0.1,
            "train_mb_size": 10,
            "seed": 0,
            "batch_size_mem": 10,
        },
        override_args
    )
    set_seed(args.seed)
[...]

PR #1287, opened 2023-01-26
replay fix proposition

AlbinSou:
Linked to issue #1274. I think this should  be enough for the fix, however I would prefer to wait for confirmation that this does not break anything. As far as I understood, loader_data is a dictionnary containing data for the potential multiple tasks present in the current experience. So, taking the maximum length over just this dictionnary should ensure we see all the data from current task, and loop on the other task dataloader if they are of unequal size. However, we don't want to take the maximum lenght also from replay loader since it can lead to the bug described in the above issue.

I also removed the part where the length of the memory buffers were added to the loaders used for length estimation.

### DIFF
--- a/avalanche/benchmarks/utils/data_loader.py
+++ b/avalanche/benchmarks/utils/data_loader.py
@@ -334,6 +334,10 @@ class ReplayDataLoader:
         balanced using the task label (i.e. each mini-batch contains a balanced
         number of examples from all the tasks in the `data` and `memory`).
 
+        The length of the loader is determined only by the current 
+        task data and is the same than what it would be when creating a 
+        data loader for this dataset.
+
         If `oversample_small_tasks == True` smaller tasks are oversampled to
         match the largest task.
 
--- a/avalanche/benchmarks/utils/data_loader.py
+++ b/avalanche/benchmarks/utils/data_loader.py
@@ -433,31 +433,6 @@ class ReplayDataLoader:
                     )[0]
                 )
 
-        if isinstance(self.memory_batch_sizes, int):
-            loaders_for_len_estimation.append(
-                _make_data_loader(
-                    memory,
-                    distributed_sampling,
-                    kwargs,
-                    self.memory_batch_sizes,
-                    force_no_workers=True,
-                )[0]
-            )
-        else:
-            for task_id in memory.task_set:
-                dataset = memory.task_set[task_id]
-                mb_sz = self.memory_batch_sizes[task_id]
-
-                loaders_for_len_estimation.append(
-                    _make_data_loader(
-                        dataset,
-                        distributed_sampling,
-                        kwargs,
-                        mb_sz,
-                        force_no_workers=True,
-                    )[0]
-                )
-
         self.max_len = max([len(d) for d in loaders_for_len_estimation])
 
     def __iter__(self):
@@ -477,15 +452,7 @@ class ReplayDataLoader:
         for t in loader_memory.keys():
             iter_buffer_dataloaders[t] = iter(loader_memory[t])
 
-        max_len = max(
-            [
-                len(d)
-                for d in chain(
-                    loader_data.values(),
-                    loader_memory.values(),
-                )
-            ]
-        )
+        max_len = max([len(d) for d in loader_data.values()])
 
         try:
             for it in range(max_len):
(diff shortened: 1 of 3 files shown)


HamedHemati:
> Linked to issue #1274. I think this should be enough for the fix, however I would prefer to wait for confirmation that this does not break anything. As far as I understood, loader_data is a dictionnary containing data for the potential multiple tasks present in the current experience. So, taking the maximum length over just this dictionnary should ensure we see all the data from current task, and loop on the other task dataloader if they are of unequal size. However, we don't want to take the maximum lenght also from replay loader since it can lead to the bug described in the above issue.
> 
> I also removed the part where the length of the memory buffers were added to the loaders used for length estimation.

Thanks @AlbinSou for the PR. I also tested similar changes in a local branch, and it worked fine. Just to be on the safe side, did you test `examples/replay.py` with very small and very large buffer sizes?

AlbinSou:
For now I tried the online replay case. I will try some things as you suggested @HamedHemati with big buffer sizes and non-online setting. @AntonioCarta in the continual learning baselines, I think the only one using ReplayDataLoader is Cope, which automatically cuts training after a set number of iterations. But you can see that with the old replay plugin the number of iteration given by the logger (before getting cut) gets bigger and bigger. With the fix it is not the case anymore.

AlbinSou:
For instance, when I try the old version with Split CIFAR100, 20 splits,10000 memory, bs 10, this is the number of iteration per experience that I get with the old plugin:

1st exp: 237 (task set size: 2375, memory size: 0)
2nd exp: 238 (task set size: 2375, memory size: 2375) -> so far no problem, just one more iteration
3rd exp: 475 (task set size: 2375, memory size: 4750)
4rth exp: 713 ...
5th exp: 950
6th exp: 1000

[...]

This keeps going on until memory is filled fully and then sticks to 1000 iteration (full pass on the full memory)

With the fix, I get 237 iterations in the first experience, 238 in the following ones. Maybe there is still a problem in the sense that this does not drop the last batch of current data. I traced it back to the ReplayPlugin in itself which is not passing the drop_last argument to the loader, I will change that too.

AlbinSou:
> Thanks @AlbinSou, the change looks ok. Can you also update the docstring to describe the new behavior? [This](https://github.com/ContinualAI/avalanche/pull/1287/files#diff-791d75e9921e2a6f9380555947aafadca9a575c54bbdda8cea3102aac93153e6L337) in particular seems outdated now.
> 
> We should definitely test this thoroughly, maybe you can try to run some of the continual-learning-baselines scripts that use the ReplayDataLoader?

About the doc string, I looked at it but I don't really know what is to change. I guess I should test the oversample_small_tasks and task_balanced_loader options which only occurs when task labels are provided. But the rest of the docstring is pretty clear and already explains the current (fixed behavior). Maybe I should precise that the loader will be of the same length than the one from current task to make it clear. Also the distributed sampling option is not explained, but I honestly don't know what it is used for and I don't know how to test it either.

AntonioCarta:
> Maybe I should precise that the loader will be of the same length than the one from current task to make it clear

Yes, you are right. Just mention that the loader uses the experience data to set the number of mini-batches.

Also, can you add a test to check that the number of iterations is correct? To avoid breaking it in the future.

AlbinSou:
I added a test and it seems like it was well needed, because one of them fails. When I check the batch size, it should be always equal to 2 times the current task batch size (if I set batch_size=current_task_bs and batch_size_mem=current_task_bs). However, it's not the case in some cases. I will investigate that but if you have an idea in the meantime it would be welcomed. I suspect the looping onto memory loader might not be effective if done at all, since this bug happens only for 3 tests, where the memory is smaller than the current task set.

AntonioCarta:
Maybe it's happening only for the last minibatch if you don't have a dataset divisible by `batch_size`?

AlbinSou:
> Maybe it's happening only for the last minibatch if you don't have a dataset divisible by `batch_size`?

No because I am using drop_last=True. And also everytime the batch is not a random size, it's either batch_size + batch_size_mem or just batch_size

AlbinSou:
Ok I fixed this it's just the normal behavior when the option oversample_small_task is not used, but it is used by default inside of the ReplayPlugin. Activating it makes the test pass.

AntonioCarta:
Great! Can I merge this or are you planning any other changes?

AlbinSou:
> Great! Can I merge this or are you planning any other changes?

You can merge, no other changes planned for now

PR #1288, opened 2023-01-26
Add `FromScratchTrainingPlugin`

HamedHemati:
This PR adds a new plugin called FromScratchTraining that resets the model's weights and optimizer state after each training experience. It is used as a baseline for comparison with sequential fine-tuning in some papers. 

Minor: we can change the plugin's name if it's wordy. Another potential name could be: "StrategyResetter"

### DIFF
--- a/avalanche/training/plugins/from_scratch_training.py
+++ b/avalanche/training/plugins/from_scratch_training.py
@@ -1,0 +1,54 @@
+from copy import deepcopy
+
+from avalanche.training.plugins.strategy_plugin import SupervisedPlugin
+from avalanche.training.templates import SupervisedTemplate
+from avalanche.models.dynamic_optimizers import reset_optimizer
+
+
+class FromScratchTrainingPlugin(SupervisedPlugin):
+    """ From Scratch Training Plugin.
+
+    This plugin resets the strategy's model weights and optimizer state after
+    each experience. It expects the strategy to have a single model and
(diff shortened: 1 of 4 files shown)


avalanche/training/plugins/from_scratch_training.py
```diff
@@ -0,0 +1,54 @@
+from copy import deepcopy
+
+from avalanche.training.plugins.strategy_plugin import SupervisedPlugin
+from avalanche.training.templates import SupervisedTemplate
+from avalanche.models.dynamic_optimizers import reset_optimizer
+
+
+class FromScratchTrainingPlugin(SupervisedPlugin):
```
AntonioCarta, line 8: I think it's better to use the `SGDTemplate` here to support more use cases.

AntonioCarta:
Should we have a supervised strategy too? Since we have them for the other baseline methods.

avalanche/training/plugins/from_scratch_training.py
HamedHemati: Agreed, will change it!

HamedHemati:
> Should we have a supervised strategy too? Since we have them for the other baseline methods.

Updates:
- Updated the class definition
- Added strategy wrapper
- Added example

AntonioCarta:
Thanks!

PR #1293, opened 2023-02-01
Fixed TensorImage being treated as an unsupported logging type

BenCrulis:
Values of type TensorImage were being treated as unsupported despite having their own code branch line 186.


AntonioCarta:
Thank you, LGTM

PR #1302, opened 2023-02-20
L2p

JuliousHurtado:
Implementation of L2P methods
"Wang, Zifeng, et al. "Learning to prompt for continual learning."  Proceedings of the IEEE/CVF Conference on Computer Vision and Pattern Recognition. 2022."
The contribution consists in:
- Adding a file with the transformers architecture to the models (based on timm library)
- Adding a file with the prompts classes
- Supervised Strategy that implement the training pipeline and loss proposed by the authors
- Add test for L2P strategy, base on a simple MLP
- Remove extra  `_train_exp` in base_sgd


avalanche/models/prompt.py
```diff
@@ -0,0 +1,168 @@
+# ------------------------------------------
+# Based on code added by:
+# -- Jaeho Lee, dlwogh9344@khu.ac.kr
+# https://github.com/JH-LEE-KR/l2p-pytorch
+# ------------------------------------------
+import torch
+import torch.nn as nn
+
+
+class Prompt(nn.Module):
```
AntonioCarta, line 10: can you add docstrings here?

avalanche/models/prompt.py
```diff
@@ -0,0 +1,168 @@
...
+        length=5,
+        embed_dim=768,
+        embedding_key="mean",
+        prompt_init="uniform",
+        prompt_pool=False,
+        prompt_key=False,
+        pool_size=None,
+        top_k=None,
+        batchwise_prompt=False,
+        prompt_key_init="uniform",
```
AntonioCarta, line 22: doc with all arguments

avalanche/models/prompt.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def l2_normalize(self, x, dim=None, epsilon=1e-12):
+        """Normalizes a given vector or matrix."""
+        square_sum = torch.sum(x**2, dim=dim, keepdim=True)
+        x_inv_norm = torch.rsqrt(torch.maximum(
+                                    square_sum, 
+                                    torch.tensor(epsilon, device=x.device)))
+        return x * x_inv_norm
+
+    def forward(self, x_embed, prompt_mask=None, cls_features=None):
```
AntonioCarta, line 66: doc with all arguments

avalanche/models/timm_vit.py
```diff
@@ -0,0 +1,1118 @@
...
+        "fixed_input_size": True,
+        "mean": IMAGENET_INCEPTION_MEAN,
+        "std": IMAGENET_INCEPTION_STD,
+        "first_conv": "patch_embed.proj",
+        "classifier": "head",
+        **kwargs,
+    }
+
+
+default_cfgs = {
```
AntonioCarta, line 85: do we need these configs in avalanche? aren't they already available on timm? I would avoid copy-pasting things that people can take from somewhere else.

avalanche/training/supervised/l2p.py
```diff
@@ -0,0 +1,175 @@
...
+from avalanche.training.templates import SupervisedTemplate
+from avalanche.models.vit import create_model
+
+
+MODEL = "vit_base_patch16_224"
+
+
+class L2PTemplate(SupervisedTemplate):
+    """
+    Learning to Prompt (L2P) plugin.
```
AntonioCarta, line 18: it's not a plugin, it's a strategy (I know that avalanche is a bit confusing on these differences).

AntonioCarta (changes requested):

AntonioCarta:
Hey @JuliousHurtado it's still not completely clear to me the relationship between Avalanche and timm. Ideally, we should use timm for vision transformers. I don't understand why you added `vit.py` and `timm_vit.py`.
If we reuse timm we don't have to maintain vision transformers ourselves, which is much better for us. Of course it's fine to add them if we need anything that is CL-specific.

JuliousHurtado:
Hey @AntonioCarta, I just finished reviewing your comments. I focus on distinguishing what remains in timm from what belongs to CL. Now, the vit.py models are imported from timm_vit.py, which is the VisualTransformer from timm, but we the modification needed to work in CL with Prompts.
Let me know what you think.

PR #1310, opened 2023-03-03
support num_workers in EWC

WenjinW:


AntonioCarta:
Thanks. If you fix the pep8 we can merge this.

lrzpellegrini:
Obtaining the number of workers using

```python
num_workers=kwargs.get('num_workers', 0)
```

may be more general, as we are not sure that `num_workers` will exist in `kwargs`. Apart from that, it seems ok!

@WenjinW Can you update the PR? Otherwise, I can create a new PR for this!

PR #1311, opened 2023-03-04
replace lambda functions with pickle compatible equivalent.

DanielAnthes:
Hello, In the current version of avalanche (0.3.1) setting num_workers > 0 in the train function of a strategy causes a crash.
In my case the reason for this was a number of lambda functions in dataset definitions, which cannot be pickled. Since PyTorch relies on pickle for multiprocessing this prohibits parallelisation in data loading.

This pull request includes a simple fix, although admittedly not a very beautiful one. Unfortunately I do not know the library well enough yet to propose a better solution. Nonetheless I hope that this is helpful to you.

Thank you for all the work you do!

Best wishes,
Daniel

### DIFF
--- a/avalanche/benchmarks/utils/classification_dataset.py
+++ b/avalanche/benchmarks/utils/classification_dataset.py
@@ -476,8 +476,7 @@ def classification_subset(
 
     if class_mapping is not None:
         frozen_transform_groups = DefaultTransformGroups(
-            # (None, lambda x: class_mapping[x])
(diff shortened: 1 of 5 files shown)


AntonioCarta:
Thank you for the fix @DanielAnthes. The PR looks ok. Can you add a test to catch to avoid a regression?

DanielAnthes:
hi @AntonioCarta, I added a test case!

AntonioCarta:
Thanks!

PR #1313, opened 2023-03-06
ER-ACE criterion simplification

AlbinSou:
As mentioned in continual learning baselines repo, I simplified the ER-ACE criterion so that it does not track the old and current classes anymore but only considers classes in the current batch as current classes. That way the method does not have to be aware of the stream structure, as it is done in the paper. The results are not changed drastically, sometimes better sometimes worst but not a big difference overall.


AntonioCarta:
Thanks

PR #1314, opened 2023-03-06
Added DER strategy

AlbinSou:
I added the DER strategy here. Below I list some implementations details and choices that we can change potentially.

- About the use of transformations: For now, I do not use transformations by default on the current data, because in avalanche this is not suppose to be a choice from the strategy but rather one from the benchmark. However, the use of input transformations is a part of the DER method, so it should be either triggered when using it in what case the actual DER will happen, or used without input transformations in what case transformations will be applied only on the buffer data.
- To store the data, I use a derived version of ClassBalancedBuffer which additionally stores the logits of the memory samples so that they can later be replayed. To do this, I use a classification tensor dataset that I create by iterating over the experience dataset, and I collect both the input data and the logits that I pad to a maximum dimension (set to 300 for now), I will explain below why. I also take care of using experience.dataset.eval() so that stored data is not already augmented data, I think this should be sufficient precaution but let me know if you think there might be a problem there.
- Another point that I am not sure how to handle is the logits size. I think the classical behavior that you would expect for the model used inside an avalanche strategy would be to have potentially varying number of logits, which could make the classification tensor dataset fail to be built since it cannot store tensors of different size in the same dataset. This is why I pad the logits to a given shape.
- Again, about the logits size, the classical DER strategy do not use varying logits size and thus store the value of the full (untrained included) logits from the beginning of training. This behavior I think was not initially desired by the authors but they later said in a follow up paper that they realized it was beneficial for the performance to do it like that. In avalanche, the strategy is not suppose to chose the architecture of the model, so the implementation I provide deals with both cases, when using a model with growing logits and also when using a models with fixed number of logits.

TLDR; While this strategy is implementing the DER and DER++ strategy, they are two details from the paper that are classically handled by avalanche (use of input transformations and model head size) that can significantly change the behavior of the method. Setting them correctly will lead to use the actual DER, otherwise it will be a modified version that tries to stick as close as possible to the original one.

### DIFF
--- a/avalanche/training/supervised/der.py
+++ b/avalanche/training/supervised/der.py
@@ -232,12 +232,12 @@ class DER(SupervisedTemplate):
         if self.replay_loader is None:
             return None
 
-        batch_x, batch_y, batch_logits, batch_tid = next(self.replay_loader)
-        batch_x, batch_y, batch_logits, batch_tid = (
+        batch_x, batch_y, batch_tid, batch_logits = next(self.replay_loader)
+        batch_x, batch_y, batch_tid, batch_logits = (
             batch_x.to(self.device),
             batch_y.to(self.device),
-            batch_logits.to(self.device),
             batch_tid.to(self.device),
+            batch_logits.to(self.device),
         )
         self.mbatch[0] = torch.cat((batch_x, self.mbatch[0]))
         self.mbatch[1] = torch.cat((batch_y, self.mbatch[1]))
@@ -278,7 +278,6 @@ class DER(SupervisedTemplate):
                     self.mb_output[: self.batch_size_mem],
                     self.mb_y[: self.batch_size_mem],
                 )
-                self.loss = self.loss / 2
             else:
                 self.loss += self.criterion()
 
--- a/avalanche/training/supervised/der.py
+++ b/avalanche/training/supervised/der.py
@@ -10,8 +10,10 @@ from torch.utils.data import Dataset
 from torchvision.transforms import Compose, RandomCrop, RandomHorizontalFlip
 
 from avalanche.benchmarks.utils import (classification_subset,
+                                        make_avalanche_dataset,
                                         make_tensor_classification_dataset)
 from avalanche.benchmarks.utils.data import AvalancheDataset
+from avalanche.benchmarks.utils.data_attribute import TensorDataAttribute
 from avalanche.benchmarks.utils.transform_groups import (EmptyTransformGroups,
                                                          TransformGroups)
 from avalanche.core import SupervisedPlugin
@@ -21,44 +23,28 @@ from avalanche.training.storage_policy import (BalancedExemplarsBuffer,
                                                ReservoirSamplingBuffer)
 from avalanche.training.templates import SupervisedTemplate
 
-class DatasetWithLogits(AvalancheDataset):
-    def __init__(self, dataset, model, batch_size, device):
-        super().__init__([dataset])
-        self.logits = self.compute_logits(dataset, model, batch_size, device)
 
-    @torch.no_grad()
-    def compute_logits(
-        self,
+@torch.no_grad()
+def compute_dataset_logits(dataset, model, batch_size, device):
+    logits = []
+    loader = torch.utils.data.DataLoader(
         dataset,
-        model=None,
-        batch_size=128,
-        device="cuda",
-    ):
-        logits = []
-        loader = torch.utils.data.DataLoader(
-            dataset,
-            batch_size=batch_size,
-            shuffle=False,
-        )
-        model.eval()
-        for x, _, _ in loader:
-            x = x.to(device)
-            out = model(x)
-            logits.append(out.cpu())
-
-        logits = torch.cat(logits)
-        return logits
-
-    def __getitem__(self, idx):
-        mb = super().__getitem__(idx)
-        import pdb;pdb.set_trace()
-        return *mb, self.logits[idx]
+        batch_size=batch_size,
+        shuffle=False,
+    )
+    model.eval()
+    for x, _, _ in loader:
+        x = x.to(device)
+        out = model(x)
+        logits.extend(list(out.cpu()))
+    return logits
 
 
 def cycle(loader):
     while True:
         for batch in loader:
             yield batch
+
 
 class ClassBalancedBufferWithLogits(BalancedExemplarsBuffer):
     """
@@ -92,13 +78,15 @@ class ClassBalancedBufferWithLogits(BalancedExemplarsBuffer):
 
     def update(self, strategy: "SupervisedTemplate", **kwargs):
         new_data = strategy.experience.dataset
-        new_data_with_logits = DatasetWithLogits(
-            new_data,
-            strategy.model,
-            strategy.train_mb_size,
-            strategy.device,
+        logits = compute_dataset_logits(
+            new_data, strategy.model, strategy.train_mb_size, strategy.device
         )
-
+        new_data_with_logits = make_avalanche_dataset(
+            new_data,
+            data_attributes=[
+                TensorDataAttribute(logits, name="logits", use_in_getitem=True)
+            ],
+        )
         # Get sample idxs per class
         cl_idxs = {}
         for idx, target in enumerate(new_data.targets):
@@ -109,7 +97,7 @@ class ClassBalancedBufferWithLogits(BalancedExemplarsBuffer):
         # Make AvalancheSubset per class
         cl_datasets = {}
         for c, c_idxs in cl_idxs.items():
-            subset = classification_subset(new_data_with_logits, indices=c_idxs)
+            subset = new_data_with_logits.subset(c_idxs)
             cl_datasets[c] = subset
         # Update seen classes
         self.seen_classes.update(cl_datasets.keys())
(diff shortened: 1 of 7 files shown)


AntonioCarta:
> For now, I do not use transformations by default on the current data

Can we reuse the original dataset transformations? This would be the best solution. Alternative: provide an argument where the user must specify the transformations.

> Another point that I am not sure how to handle is the logits size. 

I'm not sure if there is any correct choice here. We can either follow the paper (only curr+past logits) or the implementation (everything). Either way, I would add a comment clarifying this implementation detail since it's quite critical.

> implementation I provide deals with both cases, when using a model with growing logits and also when using a models with fixed number of logits.

I think this is ok.

avalanche/training/supervised/der.py
```diff
@@ -0,0 +1,263 @@
...
+        data.append(x)
+
+    logits = torch.cat(logits)
+    data = torch.cat(data)
+
+    logits = F.pad(logits, (0, _max_size - logits.shape[1]), value=0)
+
+    if len(data.shape) == 4:
+        transform = Compose(
+            [RandomCrop(data[0].shape[2], padding=4), RandomHorizontalFlip()]
```
AntonioCarta, line 52: instead of hardcoding transforms, can we have an argument that can be set by the user?

avalanche/training/supervised/der.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+
+    def __init__(
+        self,
+        model: Module,
+        optimizer: Optimizer,
+        criterion=CrossEntropyLoss(),
+        mem_size: int = 200,
+        batch_size_mem: int = None,
+        derpp: bool = False,
```
AntonioCarta, line 158: I would remove this argument and suggest the user to set `beta=0`.
Otherwise, if a user sets `beta` but forgets `derpp` it will silently ignore the argument.

AntonioCarta (changes requested):

AntonioCarta:
I talked with the authors and they confirmed me that we should keep both the transformations and the classifier size as part of the strategy. I suggest to:
- use the transformations, either taken from the dataset or passed as a strategy argument (if it's not easy to take them from the dataset)
- Return an error if the output size grows over time.

Please mention both of this in the docstring so that users are fully aware of it.

AlbinSou:
> I talked with the authors and they confirmed me that we should keep both the transformations and the classifier size as part of the strategy. I suggest to:
> 
> * use the transformations, either taken from the dataset or passed as a strategy argument (if it's not easy to take them from the dataset)
> * Return an error if the output size grows over time.
> 
> Please mention both of this in the docstring so that users are fully aware of it.

Ok, I think returning an error is a good way to handle that problem. About the transformations. Do you know if there is a simple way that I apply them to an already existing scenario (potentially without transformations), without turning the whole dataset into a tensor dataset. So far only the replay buffer is turned into a tensor dataset.

AlbinSou:
Basically I made all the changes except one so far. I don't known how to apply the transformations to the dataset if they are not here. Since I don't know how to robustly check for the presence of transformations (i.e in the online setting), and if they are not here I'm not sure what is the correct way to add them to the dataset in avalanche. So so far, the transformations passed as an argument are only applied to the replay buffer, letting the responsibility to the user to set the transformations from the scenario accordingly.

One way of handling things could be to create a tensor dataset not only for the replay buffer but also for the current dataset, and adding the transformations to this dataset instead. But I think it's a bit too much, if there is another way it would be nice.

AntonioCarta:
Maybe we can wrap the original dataset and add the logits. Untested example:
```
class DatasetWithLogits(AvalancheDataset):
    def __init__(self, data, model, other_optional_arguments_for_logits_computations,...):
        super().__init__([data])
        
        # TODO: compute logits here using model
        self.logits = ...

    def __getitem(self, idx):
        mb = super().__getitem__(idx)
        return *mb, self.logits[idx]
```

AntonioCarta:
Ok, I think I completely misunderstood the problem initially. What we want to have is:
1 - The user must be aware that augmentations are a fundamental component of DER -> mention this in the doc and ask to set them in the benchmark properly.
2 - Ensure that the replay buffer and current dataset are using the same augmentations -> inherit transformations from the original dataset.

At first, I thought your problem was that you didn't know how to reuse the augmentations from the original dataset.

The transformations of the current dataset **should not be changed** because the users can set them explicitly and we don't want to override this choice. We need to do the opposite: reuse the transformations of the original dataset for the replay buffer (which doesn't happen if we just store the tensors but not the augmentations).

The problem with the current solution is that it makes the usage more complex without solving the issue:
- If the user is aware of transformations, now he needs to set them on the plugin instead of the benchmark, which is potentially more confusing.
- If the user set bad transformations, the problem remains, whether the argument is set via the benchmarks or via the plugins.
- We may create hidden bugs where the transformations are overridden.

Sorry for the misunderstanding.

AlbinSou:
For now I came up with something like that, but I think it's a bit hacky still ... Let me know if you think there is a better way.

AlbinSou:
> > Let me know if you think there is a better way.
> 
> I think reusing transformations from the original dataset is easier if you don't convert to tensors. Basically:
> 
> * Subsample the data
> * create the dataset with logits like [here](https://github.com/ContinualAI/avalanche/pull/1314#issuecomment-1470348290)
> * That's it. The new dataset will use the original trasnforms

I agree and this is how I approached the problem the first time but then I had problems because of the way AvalancheDatasets are made. I can try again this way and tell you what was the problem.

AntonioCarta:
> @AntonioCarta Ok, here is the problem, when I do this, the getitem method that I want to define in the wrapper DatasetWithLogits is not called. I think it's because this dataset is never used as-is but rather some subsets are extracted from it and used to update the ReservoirBuffer .. This is why I changed to the tensor dataset solution initially

Can you share the code?

AlbinSou:
It looks like it works fine now but I am still not happy with the results I get with it ... Something must be wrong because I obtain results close to the ones of normal replay without augmentation but I use both augmentations and the DER strategy. I will keep investigating.

Also a quick remark, I had to fix a bug that was not spotted by the strategy test but only by my test on my continual-learning-baselines fork. This is why I changed classification_subset() call in ReservoirSamplingBuffer resize method to data.subset() call. I hope this does not hurt any other method but I think it's fine.  The basic test was not going through this method, but the integration test was going through and this was creating a bug.

When using the data attribute logits, I thought this would be happened after all the other fields, but actually it's still appened before the task_id, I don't know exactly why. The bug I encountered was that the classification_subset method was reordering this attributes so I had one part of the buffer (the one that had been resized) that was outputing x, y, t, logits, whereas the most recent, not resized one, was outputing x, y, logits, t.

AlbinSou:
> > Something must be wrong because I obtain results close to the ones of normal replay without augmentation but I use both augmentations and the DER strategy. I will keep investigating.
> 
> It's below the original paper results?
> 
> > Also a quick remark, I had to fix a bug that was not spotted by the strategy test but only by my test on my continual-learning-baselines fork. This is why I changed classification_subset() call in ReservoirSamplingBuffer resize method to data.subset() call. I hope this does not hurt any other method but I think it's fine. The basic test was not going through this method, but the integration test was going through and this was creating a bug.
> 
> Can you add a unit test for this bug?
> 
> > When using the data attribute logits, I thought this would be happened after all the other fields, but actually it's still appened before the task_id, I don't know exactly why. The bug I encountered was that the classification_subset method was reordering this attributes so I had one part of the buffer (the one that had been resized) that was outputing x, y, t, logits, whereas the most recent, not resized one, was outputing x, y, logits, t.
> 
> Definitely a bug. I didn't think about this, but probably we are not enforcing that the ordering of attributes is stable. Both in `AvalancheDataset.__init__` and in `AvalancheDataset.__getitem__` we use `for da in self._data_attributes.values()`. `self.self._data_attributes` is a dictionary.

Ok, I will try to add a unit test for this.

Yes, for now the results are lower than the original paper. Though I have not tried to use the exact settings used in the paper, it often gives results inferior to replay, and I found some inconsistencies when I set alpha=0 (which should behave similarly to replay). This last one is hinting me that something fails.

AlbinSou:
I added a test but I don't know  what is the exact behavior you wanted to test here. I checked the order of the attributes if you add a dataset_attribute to an existing dataset with other dataset attributes.

I think the expected behavior should be data, label, att1, att2 for such an operation, but right now if att2 is added after att1, it can happen (I'm not sure if this is all the time), that you have data, label, att2, att1. I think this has to do with what we discussed last time, do we want to output data, y, task_id, the rest, or data, y, the rest, task_id which is the current behavior?

I think in order to change that we have to change the inheritance of data_attributes in benchmarks/utils/data.py

AlbinSou:
Yes, I know this is because of BN but what I don't understand is that it works well without transform and not well with transforms. Anyway, as it is it works so, I will investigate that later.

I fixed the data attribute issue.

AntonioCarta:
Ok, I think we can keep the best version. Add some comments somewhere (even hidden in the implementation) to document these issues.

Can I merge it?

AlbinSou:
The test that is failing is working when I try it on my own (training/test_regularization.py). I don't understand, I just added comments and now the repo is complaining that it fails when it was not failing before. I guess all tests are not run every time ?

AntonioCarta:
All the tests are run every time. I'm not sure why it's not working here...

AntonioCarta:
Ok I tested this locally and it's working (Windows, py3.9). Maybe it has to do with the pytorch version. I'm going to merge this because if there is a problem it's probably unrelated to your changes and we should investigate it separately.

PR #1320, opened 2023-03-14
Allow setting prefetch_factor for multiprocessing

DanielAnthes:
Hello,
setting prefetch_factor to anything else than 2 caused the program to crash during length estimation of the dataset.

### DIFF
--- a/avalanche/benchmarks/utils/data_loader.py
+++ b/avalanche/benchmarks/utils/data_loader.py
@@ -590,6 +590,8 @@ def _make_data_loader(
         data_loader_args['num_workers'] = 0
         if 'persistent_workers' in data_loader_args:
             data_loader_args['persistent_workers'] = False
+        if 'prefetch_factor' in data_loader_args:
+            data_loader_args['prefetch_factor'] = 2
 
     if _DistributedHelper.is_distributed and distributed_sampling:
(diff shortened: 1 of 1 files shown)


AntonioCarta:
What's the problem with `prefetch_factor`? Maybe we should fix that instead of forcing `prefetch_factor == 2`.

DanielAnthes:
Apologies, I should have been more explicit. The fix pertains to a specific case in which `_make_data_loader` is called with `force_no_workers=True`. Specifically this is always the case when the function is called in the context of estimating the number of batches in the dataloader, in which case the no_workers option is set to obtain a single iterator to calculate the 'iterator length'.
In this context, avalanche checks for dataloader kwargs that are incompatible with `num_workers=0`, but currently this only checks for the `persistent` option for multiprocessing. The proposed fix adds an additional test for the `prefetch_factor` option. 
In the remainder of the training loop, the dataloader is recreated with `force_no_workers=False`, so multiprocessing works as intended (including setting the prefetch factor). The issue only exists in this specific 'preprocessing step' that needs a single dataloader process.

The relevant context from which the function call is made is in line 184 in the same data_loader.py file.

AntonioCarta:
Thanks, I looked at the diff but I was missing all the context.

Issue #1318, opened 2023-03-12
Replace `higher` with `functorch`


HamedHemati:
[functorch](https://pytorch.org/functorch/stable/) is a good alternative for `higher` (which is not maintained anymore) in meta-learning-based strategies and, in general, any type of strategy that requires dynamic model generation or efficient computation of Hessian. Currently, LaMAML uses `higher`. I'm creating this issue to (1) remove the requirement for the installation of `higher`  (2) provide an example of how it can be used for fast updates in the inner loop of meta-learning-based strategies in Avalanche.

AntonioCarta:
From functorch doc:
> This library is currently in [beta](https://pytorch.org/blog/pytorch-feature-classification-changes/#beta). What this means is that the features generally work (unless otherwise documented) and we (the PyTorch team) are committed to bringing this library forward. However, the APIs may change under user feedback and we don’t have full coverage over PyTorch operations.

Given the situation, I think we can keep both higher and functorch for now and remove higher only once functorch is beta (to avoid more work).

HamedHemati:
@AntonioCarta Yea will keep the higher version for now and add the `functorch`-based one as the default strategy.

HamedHemati:
Done.

PR #1326, opened 2023-03-23
Remove `higher` dependency

HamedHemati:
This PR update the LaMAML strategy and removes the `higher`  package dependency (related to #1318 ). The new version performs fast-updates using `torch.func` package which requires `torch >= 2.0.0`. In the new version of PyTorch, `functorch` which is the core library behind stateless model execution is stable:
https://pytorch.org/functorch/stable/functorch.html
that's why I removed the old version.

This strategy provides a working example of a meta-learning based strategy in avalanche that doesn't require any external dependencies. 

The accuracy of the model for split-CIFAR-100 is ~70.2. There is a slight improvement (0.2%) in the accuracy which could be related to the internal buffer used in the new version.


AntonioCarta:
Code looks good and functorch looks like an improvement over higher. The only problem is that it requires PyTorch 2.0.
Is there any clear advantage to using this new version?

Otherwise I suggest to:
- keep both versions, the old one and the new one
- keep the new version private for now (e.g. rename as `_la_maml.py`) so that users use the old one by default
- once we decide to drop pytorch < 2.0 or deprecate higher, switch to the new version

The switch to 2.0 doesn't look problematic but it seems a bit early to enforce it.

HamedHemati:
> * keep the new version private for now (e.g. rename as `_la_maml.py`) so that users use the old one by default

I re-added the `higher` version and renamed the `functorch` version to `lamaml_v2.py`.

AntonioCarta:
Great! can you fix the tests so that I can merge it?

HamedHemati:
The error is caused by package import:
```
ModuleNotFoundError: No module named 'higher'
```

It requires `avalanche[extra]`.

HamedHemati:
> Great! can you fix the tests so that I can merge it?

The error is caused by package import:
```
ModuleNotFoundError: No module named 'higher'
```

It requires `avalanche[extra]`.

avalanche/training/supervised/__init__.py
```diff
@@ -12,3 +12,4 @@
 from .er_ace import ER_ACE, OnlineER_ACE
 from .der import DER
 from .l2p import LearningToPrompt
+from .lamaml import LaMAML
```
AntonioCarta, line 15: Remove to fix the error?

AntonioCarta:
I know, but it was introduced in this PR, I believe because of the import in the `__init__`

HamedHemati:
> I know, but it was introduced in this PR, I believe because of the import in the `__init__`

Ah, right. I added it in this PR. Will remove it, thanks.

Issue #1306, opened 2023-02-26
feature request: support gradually changing distributions


murphyk:
It would be very helpful to support smoothly changing distributions, as illustrated below
(from [FOO-VB ](https://arxiv.org/pdf/2010.00373.pdf)paper, which uses permuted mnist).

<img width="648" alt="Screenshot 2023-02-24 at 5 47 30 PM" src="https://user-images.githubusercontent.com/4632336/221430103-855ae091-57d0-480b-a2e0-d889e329fabc.png">

AntonioCarta:
Avalanche benchmarks already support this use case. We are missing some examples of benchmarks with blurry tasks and a high-level generator. We will definitely add them in the future since this setting is becoming quite popular.

murphyk:
Can you point me to any examples of your existing functionality?
We are hoping to submit a paper to COLLAS (https://lifelong-ml.cc/Conferences/2023/call) on March 6th,
so would like to try this ASAP. Thx.

HamedHemati:
Hi @murphyk ,

AFAIK we currently don't have an example that creates streams with smooth transitions, but one way to do this is to use `OnlineCLScenario` from `avalanche.benchmarks.scenarios.online_scenario` to convert the list of experiences in a continual learning stream to a sequence of sub-experiences.

For example, if we have a continual learning stream with five experiences $S = [e_1, e_2, ..., e_5]$, we can convert it to a stream of smaller experiences i.e. an online stream as below:

``` python
ocl_benchmark = OnlineCLScenario(
        original_streams=benchmark.streams.values(),
        experiences=benchmark.train_stream,
        experience_size=10,
        access_task_boundaries=True
    )
```

by setting the size of each "sub-experience".

By default, the transition between the original experiences is "abrupt" since `OnlineCLScenario` creates generators for each experience in the original stream in a sequential order and goes through them one by one and outputs sub-experiences, so the final stream will look like this:
$S = [e_{1}^{(1)}, e_{1}^{(2)}, ..., e_{1}^{(L)}, e_{2}^{(1)},  ..., e_{2}^{(L)}, e_{e}^{(1)}, ..., e_{5}^{(L)}]$

For smooth transitions, the `exps_iter()` function in `avalanche.benchmarks.scenarios.online_scenario` needs to be changed. This function defines how and in which order the sub-experience generators should be iterated through  :
```python
    def exps_iter():
        for exp in original_stream:
            for sub_exp in split_foo(exp, experience_size):
[...]

Issue #1319, opened 2023-03-12
Meta-BGD strategy


HamedHemati:
I'm creating this issue to implement the Meta-BGD strategy from this paper:

It's a commonly used baseline for meta-learning-based strategies:
[Task Agnostic Continual Learning via Meta Learning](https://arxiv.org/pdf/1906.05201.pdf)

PR #1336, opened 2023-04-04
Continuous Task-Agnostic Scenario

HamedHemati:
This PR adds the gradual transition feature (also discussed in  #1306). Similar to the online scenario it takes as input a stream of experiences and generates a lazy stream of sub-experiences. The main difference in the parameters  is a parameter called alpha: $0 < \alpha \leq 0.5$ (I chose $\alpha$ for simplicity), which sets the amount of overlap between two consecutive tasks(or experiences):

transition

The gradual transition is defined as a function over the mixture of overlaps between experiences. The default transition is "linear". I will add "random" transition later. Other types of transition should also be implementable using the same approach. I've added an example and test it successfully with naive online.

Regarding the name, let me know if you have any suggestions.

Here are some examples for different values of $\alpha$:
gradual_transition_alpha_0 05

gradual_transition_alpha_0 1
gradual_transition_alpha_0 2
gradual_transition_alpha_0 3
gradual_transition_alpha_0 5


AntonioCarta:
Hi @HamedHemati, thanks for the PR. I don't understand the `GradualOnlineCLExperience`. In the smooth boundaries scenario we don't have task boundaries. Yet the definition provides them:
```
class GradualOnlineCLExperience(CLExperience):
    """Gradual Online CL (OCL) Experience.

    GOCL experiences are created by splitting a larger experience. Therefore,
    they keep track of the original experience for logging purposes.
    """

    def __init__(
        self,
        current_experience: int = None,
        origin_stream=None,
        origin_experience=None,
        subexp_size: int = 1,
        is_first_subexp: bool = False,
        is_last_subexp: bool = False,
        sub_stream_length: int = None
    ):
        """Init.

        :param current_experience: experience identifier.
        :param origin_stream: origin stream.
        :param origin_experience: origin experience used to create self.
        :param is_first_subexp: whether self is the first in the sub-experiences
            stream.
        :param sub_stream_length: the sub-stream length.
        """
        super().__init__(current_experience, origin_stream)
        self.access_task_boundaries = False

        self.origin_experience = ExperienceAttribute(
            origin_experience, use_in_train=False
        )
        self.subexp_size = ExperienceAttribute(
            subexp_size, use_in_train=False
        )
        self.is_first_subexp = ExperienceAttribute(
            is_first_subexp, use_in_train=False
        )
        self.is_last_subexp = ExperienceAttribute(
            is_last_subexp, use_in_train=False
        )
        self.sub_stream_length = ExperienceAttribute(
            sub_stream_length, use_in_train=False
        )
```
I would have expected something like:
```
class GradualOnlineCLExperience(CLExperience):
    """Gradual Online CL (OCL) Experience.

    GOCL experiences are created by splitting a larger experience. Therefore,
    they keep track of the original experience for logging purposes.
    """

    def __init__(
        self,
        current_experience: int = None,
        origin_stream=None,
        origin_experiences=None, # this becomes a list
        subexp_size: int = 1,
        # is_first_subexp: bool = False, <-- we don't have boundaries
        # is_last_subexp: bool = False, <-- we don't have boundaries
        sub_stream_length: int = None
    ):
        """Init.

        :param current_experience: experience identifier.
        :param origin_stream: origin stream.
        :param origin_experience: origin experience used to create self.
        :param is_first_subexp: whether self is the first in the sub-experiences
            stream.
        :param sub_stream_length: the sub-stream length.
        """
        super().__init__(current_experience, origin_stream)
        # self.access_task_boundaries = False   <-- we don't have boundaries
       
        # SHOULD BE A LIST
        self.origin_experience = ExperienceAttribute(
            origin_experience, use_in_train=False
        )
        self.subexp_size = ExperienceAttribute(
            subexp_size, use_in_train=False
        )

        # self.is_first_subexp = ExperienceAttribute(  <-- we don't have boundaries
        #     is_first_subexp, use_in_train=False
        # )
        # self.is_last_subexp = ExperienceAttribute(
        #    is_last_subexp, use_in_train=False
        # )
        self.sub_stream_length = ExperienceAttribute(
            sub_stream_length, use_in_train=False
        )
```

Also, are you following a definition from same paper? We should not introduce new benchmarks in Avalanche, we should just take them from the literature.

HamedHemati:
I kept the boundaries because of the interactive logger. The logger uses it to initialize the progress bar. The attributes are always disabled for training.

I couldn't find a consistent definition among existing papers. One option would be to use a (multinomial) sampler to sample from different tasks, but I don't think that would be general enough for every possible stream with blurry boundaries. Also, not all papers necessarily use a distribution-based transitions. I thought of setting some kind of overlap factor and define the transitions there. This way you can define the transition as you want using a function. For example you can set the transitions like switching back and forth between two experiences in the overlapping areas.

AntonioCarta:
> @AntonioCarta I kept the boundaries because of the interactive logger. The logger uses it to initialize the progress bar. The attributes are always disabled for training.

I think it's better to have a single progress bar for this scenario. After all, there is no reasonable point where we can reset the pbar since we don't have hard boundaries. We can add options to control this behavior later.

> I couldn't find a consistent definition among existing papers

We don't need a super general benchmark generator. What we need is:
- a definition of the experience (i.e. what attributes it should have) that is reasonably general. It's better to start with a minimal definition and add attributes later if needed (removing useless and confusing attributes is a backward-incompatible change)
- we need an example of a generator that replicates the setup of a paper. Other generators can be defined by copy/pasting that generator. Later, we can provide a general method if possible

HamedHemati:
> I think it's better to have a single progress bar for this scenario. After all, there is no reasonable point where we can reset the pbar since we don't have hard boundaries. We can add options to control this behavior later.

Currently it uses one progress bar for the entire stream, it's not restarting after each task because there are no  task boundaries. We set the `is_first_exp` only for the very first  sub-experience. We need a way to tell the progress bar not to refresh after each sub-exp. 

> we need an example of a generator that replicates the setup of a paper. Other generators can be defined by copy/pasting that generator. Later, we can provide a general method if possible

This is the part that I'm not sure about. The current generator can create streams similar to the one in figure 4 in https://arxiv.org/pdf/2010.00373.pdf.  I also looked at their code (in https://github.com/chenzeno/FOO-VB/blob/main/datasets.py), they define a linear decay in the number of samples between experiences. Similar smooth transitions can be generated with linear transition function in the current version of this PR. 

If the goal is to implement that paper, we can implement a strategy with a dataloader a dataloader that uses a customized sampler similar to the one in their code. But if we want to define the gradual transition as a feature for AVL streams, we need to properly formulate the problem first. This part that is missing. We need a reference paper.

AntonioCarta:
> Currently it uses one progress bar for the entire stream, it's not restarting after each task because there are no task boundaries. We set the is_first_exp only for the very first sub-experience. We need a way to tell the progress bar not to refresh after each sub-exp.

Oh, right. We should detect if it's an online experience (with or without boundaries) and avoid refreshing in that case.

> This is the part that I'm not sure about. The current generator can create streams similar to the one in figure 4 in https://arxiv.org/pdf/2010.00373.pdf. I also looked at their code (in https://github.com/chenzeno/FOO-VB/blob/main/datasets.py), they define a linear decay in the number of samples between experiences. Similar smooth transitions can be generated with linear transition function in the current version of this PR.

For the moment, I think it's sufficient to implement the easiest one. But it's important to follow a real paper setup.

> If the goal is to implement that paper, we can implement a strategy with a dataloader that uses a customized sampler similar to the one in their code. 

This is not possible in Avalanche because the dataloader is controlled by the strategy and not by the benchmark.

> But if we want to define the gradual transition as a feature for AVL streams, we need to properly formulate the problem first. This part that is missing. We need a reference paper.

A stream with smooth boundaries is a stream where every experience:
- is an online experience
- boundaries are unknown (i.e. we don't have the boundary attributes)
- for logging purposes, we have a reference to the list of experiences that have been used to generate them (I would keep a list since we may have more than two)

That's it. I suggest to avoid guessing what we may need in the future. We can always add things later. Removing them is much more difficult.

HamedHemati:
> This is not possible in Avalanche because the dataloader is controlled by the strategy and not by the benchmark.

I meant you can give the whole stream to the strategy, then the strategy can concatenate datasets from all experiences and use its own customized dataloader with a probabilistic sampler as in the code repo to simulate smooth transitions. 
 

> A stream with smooth boundaries is a stream where every experience:
> 
> * is an online experience
> * boundaries are unknown (i.e. we don't have the boundary attributes)
> * for logging purposes, we have a reference to the list of experiences that have been used to generate them (I would keep a list since we may have more than two)
> 
> That's it. I suggest to avoid guessing what we may need in the future. We can always add things later. Removing them is much more difficult.

But the basic version is already achievable using the existing `OnlineCLScenario` in avalanche if you set `access_boundaries=False`. The purpose of this PR was to add the feature to generate the more complicated "task-free" streams where there is also smooth transitions between (original) experiences.

AntonioCarta:
> I meant you can give the whole stream to the strategy, then the strategy can concatenate datasets from all experiences and use its own customized dataloader with a probabilistic sampler as in the code repo to simulate smooth transitions.

I understand, but we already decided to go in the opposite direction to allow the strategy to control the dataloader.

> But the basic version is already achievable using the existing OnlineCLScenario in avalanche if you set access_boundaries=False. 

This is slightly incorrect. The OCL stream currently assumes a single "parent experience". Basically, it assumes hard boundaries between tasks, even when they are unknown. This is not true when you have smooth transitions.

> The purpose of this PR was to add the feature to generate the more complicated "task-free" streams where there is also smooth transitions between (original) experiences during stream generation instead of pushing it to the strategy side.

It would be nice to have a general generator, but I think your PR and our discussion shows that we don't have a clear idea of how it should be yet. I think having a specific generator is still useful because users can copy/paste it if they want to modify it. If the super-general things is too complex they are not going to use it anyway.

HamedHemati:
> This is slightly incorrect. The OCL stream currently assumes a single "parent experience". Basically, it assumes hard boundaries between tasks, even when they are unknown. This is not true when you have smooth transitions.

We can extend the exiting OCL scenario to also support "task-free"/"task-agnostic" streams. 
By adding an additional option we can set the `origin_experience` to either `None`, or a list of origin experience ( `None` probably makes more sense?). 

Something like this:
```python
exp = OnlineCLExperience(
                origin_experience=None,
                subexp_size=experience_size,
                is_first_subexp=False,
                is_last_subexp= False,
                sub_stream_length=sub_stream_length,
                access_task_boundaries=False,
            )
``` 
With a couple of  other changes it should do what we want. This way the strategy will not have access to the boundaries and each sub-experience will be independent from its parent/origin experience. 

Do you have any preferences for the name?
I'm a bit confused about which one is correct: task-free/task-agnostic/(or something else?) :D


> It would be nice to have a general generator, but I think your PR and our discussion shows that we don't have a clear idea of how it should be yet. I think having a specific generator is still useful because users can copy/paste it if they want to modify it. If the super-general things is too complex they are not going to use it anyway.

I will first extend the existing OCL scenario to support streams without any task information. In parallel I will work on smooth transitions based on a a setup from a paper in a different branch. I will share the paper before working on it.

HamedHemati:
Here is an update regarding setting a reference paper for working on task-agnostic scenarios:

I will use this  paper, which is a different version of the same paper in the issue on gradual transitions:
https://arxiv.org/pdf/1803.10123.pdf

I find this version more intuitive and easier to follow. The chart below from the paper shows different scenarios based on the task boundary type  and the access to task labels during training and testing:

 
task-agnostic

The two scenarios that we want to focus on for this PR are:
- Discrete task-agnostic
- Continuous task-agnostic

So, in our previous discussion, the scenario where we had abrupt transitions between tasks (in the absence of task labels), would be categorized as "discrete task-agnostic", and in the case of smooth transitions where some expereinces may have samples from different tasks, or samples from different tasks may appear without clear boundaries in the stream, we have "continuous task-agnostic".

I'm going to adapt their dataloader, and prepare an example code for both discrete and continuous task-agnostic scenarios based on the experiements in the sections 5.1 and 5.2 of the paper.

I will first need to add the Bayesian Gradient Descent (BGD) strategy, which will also be the basis for its meta-learning-based version for the strategy in #1319.

I hope this makes the objective more clear.

HamedHemati:
Here are some of the recent (major) changes that I've made  in this PR:

This PR is now fully based on the reference paper ( https://arxiv.org/pdf/1803.10123.pdf). I've also changed the scenario generator's name to "Continuous Task-Agnostic" accordingly.

I've adapted the code from one of the official repositories (https://github.com/chenzeno/FOO-VB/tree/main), and the scenario generator now uses the default linear decay as in the paper, with two control parameters which are the "overlap factor - beta" and the number of virtual iterations in each epoch.

Here are few examples for different beta values 0 --> 10:

gradual_transition_beta_0

gradual_transition_beta_1 1

gradual_transition_beta_2

gradual_transition_beta_5

gradual_transition_beta_10


The paper tests the scenario with their method based on Online Variational Bayes, which I'm planning to push in a separate PR.

avalanche/benchmarks/scenarios/continuous_task_agnostic_scenario.py
```diff
@@ -0,0 +1,330 @@
...
+                yield idx
+            self._reset_indices()
+
+    def __len__(self):
+        return self.n_samples
+
+
+def create_sub_exp_from_multi_exps(
+    online_benchmark: "OnlineCLScenario",
+    original_stream: Iterable[DatasetExperience[TClassificationDataset]],
```
AntonioCarta, line 73: can we remove original_stream from the arguments? original streams should already be in `online_benchmark` anyway...

avalanche/benchmarks/scenarios/continuous_task_agnostic_scenario.py
```diff
@@ -0,0 +1,330 @@
...
+    )
+
+    # For visualization purposes only
+    exp.n_samples_from_each_exp = n_samples_from_each_exp
+
+    return exp
+
+
+def _fixed_size_linear_decay_stream(
+    online_benchmark: "OnlineCLScenario",
```
AntonioCarta, line 135: it's taking a benchmark but returning a stream? I think it would be more intuitive if it took a benchmark and returned a new benchmark. Also, why is this an `online_benchmark`? I think it should take large batches and split them.

avalanche/benchmarks/scenarios/continuous_task_agnostic_scenario.py
```diff
@@ -0,0 +1,330 @@
...
+    stream_name: str = getattr(original_stream, "name", "train")
+    return CLStream(
+        name=stream_name,
+        exps_iter=exps_iter(),
+        set_stream_info=True,
+        benchmark=online_benchmark,
+    )
+
+
+class ContinuousTaskAgnosticScenario(
```
AntonioCarta, line 254: Do we need a new scenario? I think this is just an OCL scenario without task boundaries.

examples/continuous_task_agnostic_naive.py
```diff
@@ -0,0 +1,97 @@
...
+        train_passes=1,
+        train_mb_size=10,
+        eval_mb_size=32,
+        device=device,
+        evaluator=eval_plugin,
+    )
+
+    # Create streams using the continuous task-agnostic scenario
+    batch_streams = benchmark.streams.values()
+    cta_benchmark = ContinuousTaskAgnosticScenario(
```
AntonioCarta, line 79: Again, it seems that it would be easier to have a function instead of a class. Something like: `cta_bm = continuous_task_agnostic_benchmark(original_benchmark, ...)`

AntonioCarta (changes requested):

AntonioCarta:
Hi Hamed. I have two main comments:
- why do we need a new scenario? I think the OCL scenario is already able to handle this setup
- can you add some tests?

HamedHemati:
Latest updates: 

- Continuous linear-decay splitter is now adapted according to the latest API
- I had to add two new parameters to `OnlineCLScenario` for controlling the amount of "overlap", and number of "virtual iteration" based on the definitions in the reference paper. I think they can also be used for future similar splitters that use a different type of decay.
- I tested the example with the naive strategy and works fine.

AntonioCarta:
Two quick comments:
- please add test for the splitter method
- I would avoid using `OCLSCenario` (which I want to remove actually).  you can use a method to split the stream directly. Look at the [online example](https://github.com/ContinualAI/avalanche/blob/fca0ca04969aa7222b8b82f51730fcade3cfa7b1/examples/online_replay.py#L121), which uses `split_online_stream` directly. I would expect a similar method `split_with_linear_decay(train_stream, ...)`

HamedHemati:
> * please add test for the splitter method
> * I would avoid using `OCLSCenario` (which I want to remove actually).  you can use a method to split the stream directly. Look at the [online example](https://github.com/ContinualAI/avalanche/blob/fca0ca04969aa7222b8b82f51730fcade3cfa7b1/examples/online_replay.py#L121), which uses `split_online_stream` directly. I would expect a similar method `split_continuous_linear_decay(train_stream, ...)`

Latest changes:
+ I've added the test function under the OCL test
+ I've fixed the example. Now it directly generates the continuous OCL stream from the train stream of the main benchmark with the splitter function.

AntonioCarta:
Implementation looks good.
- can you expand the doc? what is a virtual epoch? maybe put a reference to the original paper too.
- can you check the classes in the test are correct?

HamedHemati:
> Implementation looks good.
> 
> * can you expand the doc? what is a virtual epoch? maybe put a reference to the original paper too.
> * can you check the classes in the test are correct?

Done

PR #1338, opened 2023-04-12
Faster checkpointing and simpler API

AntonioCarta:
continues my work on #1256 

- speed up checkpointing by removing useless attributes
- proposes new functional API

Example:
```
    # STEP 1: SET THE RANDOM SEEDS to guarantee reproducibility
    RNGManager.set_random_seeds(1234)

    # Nothing new here...
    # Create the strategy (as usual)
    strategy = Naive(...)

    # STEP 2: TRY TO LOAD THE LAST CHECKPOINT
    # if the checkpoint exists, load it into the newly created strategy
    # the method also loads the experience counter, so we know where to
    # resume training
    fname = './checkpoint.pkl'  # name of the checkpoint file
    strategy, initial_exp = maybe_load_checkpoint(strategy, fname)

    # STEP 3: USE THE "initial_exp" to resume training
    for train_exp in benchmark.train_stream[initial_exp:]:
        strategy.train(train_exp, num_workers=4, persistent_workers=True)
        strategy.eval(benchmark.test_stream, num_workers=4)

        # STEP 4: SAVE the checkpoint after training on each experience.
        save_checkpoint(strategy, fname)
```

@lrzpellegrini what do you think?

Personally, I would prefer to remove the plugin and leave only the functional API for the moment. The plugin does some strong assumptions about how eval and train are called. These assumptions are simply not true in practice. With this functional API, the user calls the save method explicitly, removing the problem.


examples/checkpointing.py
lrzpellegrini, line 1: I think it's better to have two different example scripts checkpointing_plugin.py and checkpointing_standalone.py. Maybe we can add, in the description of the plugin example, that the standalone version is more flexible and to refer to that example.

avalanche/training/templates/base_sgd.py
```diff
@@ -238,6 +238,14 @@ def _before_training_exp(self, **kwargs):
 
         super()._before_training_exp(**kwargs)
 
+    def _after_training(self, **kwargs):
+        super()._after_training(**kwargs)
+        # reset for faster serialization
```
lrzpellegrini, line 243: This is a good idea, but it seems that it is missing its "eval" equivalent. Also, maybe it's better to have a separate "cleanup" method called by both _after_train and _after_eval.

avalanche/training/templates/base.py
```diff
@@ -154,6 +158,10 @@ def eval(
 
         self._after_eval(**kwargs)
 
+        # reset for faster serialization
+        self.current_eval_stream = None
```
lrzpellegrini, line 162: I think it's better if we put everything inside a cleanup method called from _after_train and _after_eval

avalanche/training/plugins/checkpoint.py
```diff
@@ -245,7 +246,13 @@ def after_training(self, strategy, *args, **kwargs):
     def save_checkpoint(checkpoint_data, fobj: Union[BinaryIO, IO[bytes]]):
         # import dill.detect
         # dill.detect.trace(True)
+
+        # compute checkpoint duration
+        print('Saving checkpoint...')
+        start_time = time.time()
         torch.save(checkpoint_data, fobj, pickle_module=dill)
+        end_time = time.time()
+        print('Checkpoint duration:', end_time - start_time)
```
lrzpellegrini, line 255: ... created in X seconds.

avalanche/training/checkpoint.py
```diff
@@ -0,0 +1,89 @@
...
+    The file can be loaded using `maybe_load_checkpoint`.
+
+    For efficiency, the user can specify some attributes to `exclude`.
+    For example, if the optimizer is static and doesn't change during training,
+    it can be safely excluded. These helps to speed up the serialization.
+
+    WARNING: the method cannot be used inside the training and evaluation loops of the strategy.
+
+    :param strategy:
+    :param fname:
```
lrzpellegrini, line 72: Missing the description for strategy and fname

lrzpellegrini:
In general, I think that the PR is good. The two major concerns are splitting the examples (if we want to keep the plugin) and moving strategy cleanup operations to an ad-hoc method. I think that the plugin can be removed, but a descriptive error message should show up if the user tries to use it (telling the user that it should switch to the standalone mode + a link to the standalone example).

AntonioCarta:
I added everything you asked. I'm not convinced that the cleanup method really helps because (1) we need different cleanup ops for train/eval and (2) we may need to clean some variable that are initialized in the inner loops at some point, and it's not clear whether that method is the best place to do it. But I added it anyway.

PR #1339, opened 2023-04-12
Cumulative accuracies

AlbinSou:
Adds the cumulative accuracy metric from "On the importance of cross-task features in class-incremental learning" https://arxiv.org/abs/2106.11930.

For now I only put the accuracies, maybe later I will add the forgetting metric derived from this accuracy metric.

I don't know if we should change the name a bit or not, I am afraid that it could be confounded with Average Anytime Accuracy that I know some works also call it similarly (like cumulative or accumulative accurracy).

I tested it on my baselines branch and checked that it worked, I'm still thinking of a nice unittest I could add.

### DIFF
--- a/avalanche/evaluation/metrics/cumulative_accuracies.py
+++ b/avalanche/evaluation/metrics/cumulative_accuracies.py
@@ -1,0 +1,140 @@
+################################################################################
+# Copyright (c) 2021 ContinualAI.                                              #
+# Copyrights licensed under the MIT License.                                   #
+# See the accompanying LICENSE file for terms.                                 #
+#                                                                              #
+# Date: 30-12-2020                                                             #
+# Author(s): Lorenzo Pellegrini                                                #
+# E-mail: contact@continualai.org                                              #
+# Website: www.continualai.org                                                 #
+################################################################################
+
+from collections import defaultdict
+from typing import Dict, List, Union
+
+import torch
+from torch import Tensor
+
+from avalanche.benchmarks import OnlineCLExperience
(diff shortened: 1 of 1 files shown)


AlbinSou:
Yes, I will add some tests but also I'm trying to add the corresponding forgetting measure

AlbinSou:
Added cumulative forgetting, I will add all the tests asap.

There is one thing I wanted to comment, the way the cumulative accuracies are reported now is the following:

CumulativeAccuracy/eval_phase/test_stream/Task000: 0.88
CumulativeAccuracy/eval_phase/test_stream/Task001: 0.69
...

However, this can be a bit confusing since the task ID is never changing in class-incremental learning (always 0), but I still want to show the per-experience CumulativeAccuracy. Is this the correct way to do so or should I change it ?

AntonioCarta:
> However, this can be a bit confusing since the task ID is never changing in class-incremental learning (always 0), but I still want to show the per-experience CumulativeAccuracy. Is this the correct way to do so or should I change it ?

I would call them by exp, like `CumulativeAccuracy/eval_phase/test_stream/Exp000`.
Can you change it? the rest looks ok.

AlbinSou:
> > However, this can be a bit confusing since the task ID is never changing in class-incremental learning (always 0), but I still want to show the per-experience CumulativeAccuracy. Is this the correct way to do so or should I change it ?
> 
> I would call them by exp, like `CumulativeAccuracy/eval_phase/test_stream/Exp000`. Can you change it? the rest looks ok.

I changed it, I also prefer it.

AntonioCarta:
cool, you just need to pull the master branch to fix the merge conflict.

Issue #1340, opened 2023-04-12
Training speed of MIR


HamedHemati:
Hi @AlbinSou,

I've tagged you in this issue because you have implemented the strategy in avalanche and have the experience with training MIR.

I realized that the training speed of MIR is very slow, probably due to the virtual update and the buffer sample ranking. For my case training speed goes down from ~24 it/sec in the first experience to ~3 it/sec in the second experience on. Is this normal?
I'm setting the `batch_size_mem` and `subsample` to 100.  I also noticed that `subsample` is never used. I guess it should be used as the batch size of the replay loader, right?

AlbinSou:
I made a PR to fix the unused subsample argument. As for the speed, I also notice a drop but it's lower than the one you report. I drop from 37~ish it/s to 14~ish it/sc, which is a big drop indeed but I think it's expected and not so big compared to the one you describe. Could you precise what benchmark you were using ?

HamedHemati:
> I made a PR to fix the unused subsample argument. As for the speed, I also notice a drop but it's lower than the one you report. I drop from 37 ish it/s to 14 ish it/sc, which is a big drop indeed but I think it's expected and not so big compared to the one you describe. Could you precise what benchmark you were using ?

I see. It makes sense for the training speed to be reduced by ~half since you need to backprop twice. But in my case the drop is significant. I'm using a batch training setup instead of online, but it shouldn't affect the loss calculation speed. The benchmark is class-incremental CIFAR-100 with 20 tasks. Train and buffer batch sizes are set to 64 and the buffer size and subsample are both set to 100.

HamedHemati:
Ok, after some tests I noticed that it's a dataloading bottleneck issue and it's not related to the strategy itself. The slowdown caused by the startegy itself is normally  2x, but if the buffer size is large and the number of transformations applied to the dataset  is high, the dataloading process will be slow with num_workers=0.

AntonioCarta:
Ok, I'm closing this because we already have some issues about dataloading. If you want you can comment on those or open a new one specific for OCL.

PR #1341, opened 2023-04-12
fixed mir unused subsample argument

AlbinSou:
Maybe wait before merging since this is linked to issue #1340 and @HamedHemati also reported a huge slowdown. If I find out where it comes from we might fix it also in this PR.


HamedHemati:
As discussed in #1340 the strategy functions normally (for smaller buffer sizes), and this PR can be merged.

PR #1350, opened 2023-04-24
New Question Answering example. New wrapper for Huggingface Datasets

MickHardins:
Examples on how to use Avalanche for Natural Language Processing. Question answering example added. HuggingFace dataset wrapper added.


examples/hf_datawrapper.py
```diff
@@ -0,0 +1,268 @@
+import datasets as ds
```
AntonioCarta, line 1: can you add a description here?

MickHardins:
`nlp.py` removed as it's redundant as correctly pointed out.

AntonioCarta:
Thanks!

Issue #1353, opened 2023-04-25
freeze_transform does not work for datasets with child AvalancheDataset in _datasets


HeLehm:
🐛 **Description**
If an AvalancheDataset (parent) contains other AvalancheDatasets (children) (in _datasets) and  .freeze_transforms() is being called on the parent, the children's transforms are not frozen.

🐜 **To Reproduce**
```
def test_avalanche_inherit_groups_freeze_transforms(self):
            original_dataset = MNIST(
                root=default_dataset_location("mnist"), download=True
            )
    
            transform_groups = dict(
                train=(RandomCrop(16), None), eval=(None, None)
            )
            dataset = make_classification_dataset(
                original_dataset, transform_groups=transform_groups
            )
    
            dataset_inherit = make_classification_dataset(dataset)
            x, *_ = dataset_inherit[0]
    
            dataset_frozen = dataset_inherit.freeze_transforms()
            x2, *_ = dataset_frozen[0]
    
            dataset_frozen_reset = dataset_frozen.replace_current_transform_group(
                None
            )
            x3, *_ = dataset_frozen_reset[0]
    
            self.assertEqual(x.size, (16, 16))
            self.assertEqual(x2.size, (16, 16))
            # should be (16, 16) here as RandomCrop(16) should be frozen
            self.assertEqual(x3.size, (16, 16))
```

🐝 **Expected behavior**
I would expect the transforms of the children to also be frozen.
In the example x3 should be an Image of size (16, 16), but it is of size (28, 28), the original MNIST image size.

🦋 **Additional context**
I am working on a fix right now.

PR #1354, opened 2023-04-25
Fix for Issue #1353

HeLehm:
When calling .freeze_transfroms() on a dataset all child datasets will also freeze their transforms now.


AntonioCarta:
Thank you! This is a very important bug.

PR #1356, opened 2023-04-28
Supervised Contrastive Replay implementation.

AndreaCossu:
This PR adds Supervised Contrastive Replay (SCR). This is a draft implementation, open to discussion. The example is working in the sense that it runs without errors.
I have not used any online learning APIs so far.

Some caveats:

1. The accuracy cannot be measured at training time, since during training the SCR does not use any classifier (just a custom loss).
2. The SCR loss cannot be measured at test time since it requires building a multiviewed batch with augmentations. I switched to the CrossEntropy at eval time but people need to look at the APIDoc to know that. Also, I need to check if this switch may break something or if it is better to override `self._criterion` directly.
3. Transformations to build the multiviewed batch are supported only if they can work on (batched) tensors, not on single PIL images. This really simplifies the implementation. I think it could be feasible to remove this constraint, but we would probably need a custom dataset to apply transformations and augment the mini-batch.


AlbinSou:
> The accuracy cannot be measured at training time, since during training the SCR does not use any classifier (just a custom loss).

- I think this is not a problem, we are not really interested in the accuracy on the train stream (I mean, it's not a problem if we don't have it for free during training like what happens in supervised learning), but more on the validation stream. And I think this can be computed by using the nearest mean classifier on the buffer samples as they describe in the paper.

AndreaCossu:
Yes, @AlbinSou I agree. I meant that if applied blindly, the user may expect to get the training accuracy and this would raise an error. But we cannot do much more than clarifying that in the APIdoc :smile:

examples/supervised_contrastive_replay.py
```diff
@@ -0,0 +1,71 @@
+import torchvision.transforms
```
AntonioCarta, line 1: I would remove the example. I don't think it adds anything. We will have a script on continual-learning-baselines

AndreaCossu:
I already updated the APIdoc. As for the TrainEvalModel, I will convert it to a standard pytorch Module.

avalanche/models/dynamic_modules.py
```diff
@@ -455,10 +455,28 @@ def eval_adaptation(self, experience: CLExperience = None):
         self.classifier = self.eval_classifier
 
 
+class NormalizedTrainEvalModel(TrainEvalModel):
```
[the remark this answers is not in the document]
AndreaCossu: Fixed

examples/supervised_contrastive_replay.py
AntonioCarta: You can do whatever you want in your local branch, as long as you don't push it ;)

> In case, we should remove all examples like this one, because we have a lot right now that just show how to use a certain strategy.

I agree that we should remove them. You can remove all the examples that are trivial and already available in continual-learning-baselines.

avalanche/models/dynamic_modules.py
```diff
@@ -30,7 +29,7 @@ class DynamicModule(Module):
     `model_adaptation`, which adapts the model given the current experience.
     """
 
-    def adaptation(self, experience: CLExperience):
+    def adaptation(self, experience: CLExperience = None):
```
AntonioCarta, line 32: Why do you need this to be None?

avalanche/training/supervised/supervised_contrastive_replay.py
```diff
@@ -0,0 +1,180 @@
...
+    Accuracy cannot be monitored during training (no NCM classifier).
+    During training, NCRLoss is monitored, while during eval
+    CrossEntropyLoss is monitored.
+
+    The original paper uses an additional fine-tuning phase on the buffer
+    at the end of each experience (called review trick, but not mentioned
+    in the paper). This implementation does not implement the review trick.
+    """
+    def __init__(self,
+                 model: Module,
```
AntonioCarta, line 34: fix type? should be `SCRModel`

AntonioCarta: actually, this should also be checked explicitly, because otherwise the method doesn't work. Raising a `ValueError` may be the best choice here.

avalanche/models/dynamic_modules.py
AndreaCossu: that's definitely a typo, I never intended to modify that class.

avalanche/training/supervised/supervised_contrastive_replay.py
AndreaCossu: Agreed

AntonioCarta:
@AndreaCossu you marked my first comment as fixed but I still see the old change

AndreaCossu:
@AntonioCarta  I want to change the NCM Classifier because it is not very easy to use right now, but I will create a new PR for that, this can be merged if it looks ok for you

PR #1358, opened 2023-05-01
Update .gitignore

HamedHemati:
Minor update: adds .DS_Store metafiles on macOS to `.gitignore`. You have them inside every folder. Adding them to the ignore list makes tracking changes easier.


AntonioCarta:
Thanks. FYI git allows you to define user-level gitignores that apply to all the projects.

PR #1363, opened 2023-05-06
GSA strategy

HamedHemati:
This PR adds the GSA strategy. The implementation is based on the implementation in:
[https://github.com/gydpku/GSA/blob/main/GSA_CVPR/test_cifar100.py](https://github.com/gydpku/GSA/blob/main/GSA_CVPR/test_cifar100.py)

Here are some important changes that I’ve made to make the code simpler and easier to follow:

- Variable names are changed to improve the readability of the code.
- Unused losses are removed: the code computes SimCLR loss for some reason which is never used in the final loss and not discussed in the paper.
- Equivalent `torchvision` transforms are used instead of the user-defined augmentations for the augmentation function. I compared both in a notebook to make sure they are the same.
- Input masking via `index_select` is used instead of using class sorted (re-mapped) online streams.
- The current buffer is based on the reservoir sampling storage policy in Avalanche with a customized task-based sampler similar to the original code/paper.
- `random_flip` is replaced with `torch.flip` and is applied for only the final dimension. In the original code they compute three different flips but only use the flip for the final dimension. I also checked this transformation and compared both.
- Batch accuracy during training is calculated for the first half of  logits (since it’s extended and augmented random cropping and random-grayscaling).
- Also removed some other parts that are never used: e.g. disabled gradients for inputs since the inputs are never changed (they are not supposed to be changed).

Current issues:
- The accuracies is not even close to the ones reported in the paper: ~20% in CIFAR-100 (10-tasks) compared to ~31%.
- The focus of the strategy seems to be mainly on the first experience as the accuracy remains high for the first experience.
- The accuracy of the current experience is 0% after it is training on the current experience, but then the accuracy goes up for around ~30 after training for another experience. It's somehow one experience behind. The individual loss values (`loss-p` and `loss_n`) look normal and seem to be correct though.
- The loss computed is the reference code is not exactly the same as the final objective in the paper. This is the part that I need to investigate further.


neuperc:
Update: I've stopped working on this strategy for now until there is an updated version of the paper.

HamedHemati:
Closing this PR due to reproducibility issues.

PR #1372, opened 2023-05-11
Fix issue #1371

### LINKED ISSUE #1371: classes_in_this_experience fails on certain type of experiences
🐛 **Describe the bug**
Accessing the experience attribute classes_in_this_experience fails on certain types of experience (DatasetExperience in my case).

🐜 **To Reproduce**

```{python}
#!/usr/bin/env python3                                                                                      
import unittest                                                                                             
from avalanche.benchmarks.classic import SplitCIFAR10                                                       
from avalanche.benchmarks.generators import benchmark_with_validation_stream                               
                                                                                                            
class TestBug(unittest.TestCase):                                                                           
    def test_load_scenario(self):                                                                           
        scenario = SplitCIFAR10(5)                                               
        scenario = benchmark_with_validation_stream(scenario, validation_size=0.05)                         
        classes = scenario.train_stream[0].classes_in_this_experience  
```

🦋 **Additional context**
I already did a gitbisect on the above test and here is the output:

a1ee5e19362b171bbf295c8e9ba0ec46700ccf1a is the first bad commit                                           
commit a1ee5e19362b171bbf295c8e9ba0ec46700ccf1a                                                            
Author: @lrzpellegrini                                            
Date:   Fri Apr 14 17:14:13 2023 +0200                                                                     
                                                                                                           
    Simplified generics chain. Better naming of ClassificationScenario.

lrzpellegrini:
The recent typing fixes switched the factory validation and data incremental generators to a generic DatasetScenario, which is problem-agnostic. With this PR, we try to return a benchmark of the same class of the input benchmark, allowing experiences to be created with the correct fields.

Fixes #1371

### DIFF
--- a/avalanche/benchmarks/generators/benchmark_generators.py
+++ b/avalanche/benchmarks/generators/benchmark_generators.py
@@ -32,13 +32,17 @@ from typing import (
 )
 
 import torch
+from avalanche.benchmarks.scenarios.classification_scenario import \
+    ClassificationScenario
 
 from avalanche.benchmarks.scenarios.dataset_scenario import (
     DatasetScenario,
     DatasetStream,
     FactoryBasedStream,
     StreamDef,
+    TStreamsUserDict,
 )
+from avalanche.benchmarks.scenarios.detection_scenario import DetectionScenario
 from avalanche.benchmarks.scenarios.generic_benchmark_creation import *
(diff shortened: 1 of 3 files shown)


AntonioCarta:
Thanks

PR #1375, opened 2023-05-16
Fix issue #1232. Added unit tests.

### LINKED ISSUE #1232: torchmetrics v0.11 incompatiblity
Hello guys :),
so just two days ago, torchmetrics relrease a new version V0.11 ([see here](https://pypi.org/project/torchmetrics/#history)) which redefined their class torchmetrics.functional.accuracy to require definition of task type (see link : [Click Here](https://torchmetrics.readthedocs.io/en/stable/classification/accuracy.html#accuracy)).

this causes an error to be raised when using accuracy metrics , since it uses accuracy class without defining task positional argument.
for now I managed to make my current code work by specifying the previous version of torchmetrics as follows:
`!pip install torchmetrics==0.10.3`

lrzpellegrini:
Fixes an issue with torchmetrics >= 0.11.0 and the top-k accuracy

Fixes #1232

### DIFF
--- a/avalanche/evaluation/metrics/topk_acc.py
+++ b/avalanche/evaluation/metrics/topk_acc.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, List, Union, Dict
 
 import torch
 from torch import Tensor
+import torchmetrics
 from torchmetrics.functional import accuracy
 
 from avalanche.evaluation import Metric, GenericPluginMetric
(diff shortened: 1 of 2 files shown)


lrzpellegrini:
@AntonioCarta There is something wrong with an unrelated test here :'(

AntonioCarta:
@AlbinSou can you check this? The cumulative acc test is failing

PR #1383, opened 2023-05-23
Faster attributes

### LINKED ISSUE #1357: `concat_datasets` is slow when used for grouped ReservoirSampling buffers with attributed dataset
This issue is partially related to #1348, but since it happens only in datasets with attributes, I'm creating a new issue.

The problem happens when trying to concatenate buffer groups with attributed datasets in a class-balanced (or probably other grouped) storage policies. One of the strategies that is directly affected by this issue is DER, which adds sample "logits" as attributes to the buffer dataset. In scenarios such as CIR or DI where you update most or all buffer groups after each experience, the concatenation of datasets become much longer over time. This problem happens only when new attributes are added to the dataset.  Here are some numbers 

Concatenation of group datasets **without** added attributes:
- after 1 buffer update: ~ 0.002 sec
- after 10 buffer updates: ~ 0.002 sec
- after 20 buffer updates: ~ 0.002 sec

Concatenation of group datasets **with** added attributes:
- after 1 buffer update: ~ 0.002 sec
- after 10 buffer updates: ~ 6 secs
- after 20 buffer updates: ~24 secs

The results are for domain-incremental MNIST. The concatenation duration is much longer for CIFAR benchmarks.

The duration is calculated only for accessing the storage policy's buffer via `storage_policy.buffer`.

Here is the code to reproduce the issue on the master branch. I've taken the code from DER's storage policy and modified the update function to enable buffer updates via dataset directly:

```python


import torch
import time

from avalanche.benchmarks import SplitMNIST, SplitCIFAR100
from avalanche.benchmarks.utils import make_avalanche_dataset
from avalanche.benchmarks.utils.data_attribute import TensorDataAttribute
from avalanche.training.storage_policy import (BalancedExemplarsBuffer,
                                               ReservoirSamplingBuffer)


class ClassBalancedBufferWithLogits(BalancedExemplarsBuffer):
    """
    ClassBalancedBuffer that also stores the logits
    """

    def __init__(
        self,
        max_size: int,
        adaptive_size: bool = True,
        total_num_classes: int = None,
    ):
        if not adaptive_size:
            assert (
                total_num_classes > 0
            ), """When fixed exp mem size, total_num_classes should be > 0."""

        super().__init__(max_size, adaptive_size, total_num_classes)
        self.adaptive_size = adaptive_size
        self.total_num_classes = total_num_classes
        self.seen_classes = set()

    def update(self, dataset, add_attributes=True):
        logits = [torch.randn(1, 10) for _ in range(len(dataset))]
        if add_attributes:
            new_data_with_logits = make_avalanche_dataset(
                dataset,
                data_attributes=[
                    TensorDataAttribute(logits, name="logits", use_in_getitem=True)
                ],
            )
        else:
            new_data_with_logits = dataset
        # Get sample idxs per class
        cl_idxs = {}
        for idx, target in enumerate(dataset.targets):
            if target not in cl_idxs:
                cl_idxs[target] = []
            cl_idxs[target].append(idx)

        # Make AvalancheSubset per class
        cl_datasets = {}
        for c, c_idxs in cl_idxs.items():
            subset = new_data_with_logits.subset(c_idxs)
            cl_datasets[c] = subset
        # Update seen classes
        self.seen_classes.update(cl_datasets.keys())

        # associate lengths to classes
        lens = self.get_group_lengths(len(self.seen_classes))
        class_to_len = {}
        for class_id, ll in zip(self.seen_classes, lens):
            class_to_len[class_id] = ll

        # update buffers with new data
        for class_id, new_data_c in cl_datasets.items():
            ll = class_to_len[class_id]
            if class_id in self.buffer_groups:
                old_buffer_c = self.buffer_groups[class_id]
                # Here it uses underlying dataset
                old_buffer_c.update_from_dataset(new_data_c)
                old_buffer_c.resize(None, ll)
            else:
                new_buffer = ReservoirSamplingBuffer(ll)
                new_buffer.update_from_dataset(new_data_c)
                self.buffer_groups[class_id] = new_buffer

        # resize buffers
        for class_id, class_buf in self.buffer_groups.items():
            self.buffer_groups[class_id].resize(None, 
                                                class_to_len[class_id])
            

benchmark = SplitMNIST(
    n_experiences=1,
    dataset_root=expanduser("~") + "/.avalanche/data/",
)

storage_policy = ClassBalancedBufferWithLogits(max_size=2000)
dataset = benchmark.train_stream[0].dataset
num_updates = 20
for i in range(num_updates):
    print("Update ", i + 1)
    storage_policy.update(dataset, add_attributes=True)

start = time.time()
print("Buffer size: ", len(storage_policy.buffer))
end = time.time()
duration = end - start
print("Buffer access duration: ", duration)

```

You can change `num_updates`  and `add_attributes` variables to see the effect.

AntonioCarta:
closes #1357 

the high level idea is that before data attributes were part of the "dataset tree". However, we can manage them independently. Now an AvalancheDataset is a DatasetWithTransforms + a set of attributes. Overall easier to manage and speeds up #1357, which is now on par with the normal concat.

@lrzpellegrini as you can see there are two tests failing
- top-k I think is a change in the API, so it doesn't allow to check k=4 anymore
- I have no idea what the equality check in `custom_streams_name_and_length` is supposed to do. Shouldn't we check equality of the samples?

@neuperc @AlbinSou this fixes the DER slowdown.

### DIFF
--- a/avalanche/benchmarks/utils/classification_dataset.py
+++ b/avalanche/benchmarks/utils/classification_dataset.py
@@ -100,11 +100,11 @@ class ClassificationDataset(AvalancheDataset[T_co]):
     
     def subset(self, indices):
         data = super().subset(indices)
-        return data.with_transforms(self._transform_groups.current_group)
+        return data.with_transforms(self._flat_data._transform_groups.current_group)
 
     def concat(self, other):
         data = super().concat(other)
-        return data.with_transforms(self._transform_groups.current_group)
+        return data.with_transforms(self._flat_data._transform_groups.current_group)
 
     def __hash__(self):
         return id(self)
--- a/avalanche/benchmarks/utils/data.py
+++ b/avalanche/benchmarks/utils/data.py
@@ -167,7 +167,6 @@ class AvalancheDataset(IDataset[T_co]):
                         "has length {}".format(da.name, len(da), ld)
                     )
 
-        datasets = list(filter(lambda d: len(d) > 0, datasets))
         self._data_attributes: Dict[str, DataAttribute] = OrderedDict()
         first_dataset = datasets[0] if len(datasets) > 0 else None
         if isinstance(
@@ -185,7 +184,7 @@ class AvalancheDataset(IDataset[T_co]):
                 for d2 in datasets[1:]:
                     if hasattr(d2, attr.name):
                         acat = acat.concat(getattr(d2, attr.name))
-                    else:
+                    elif len(d2) > 0:  # if empty we allow missing attributes
                         found_all = False
                         break
                 if found_all:
@@ -263,7 +262,7 @@ class AvalancheDataset(IDataset[T_co]):
         :param indices: The indices to keep.
         :return: A new dataset.
         """
-        return self.__class__(self, indices=indices)
+        return self.__class__([self], indices=indices)
 
     @property
     def transform(self):
--- a/avalanche/benchmarks/utils/data.py
+++ b/avalanche/benchmarks/utils/data.py
@@ -44,9 +44,10 @@ from collections import OrderedDict
 
 T_co = TypeVar("T_co", covariant=True)
 TAvalancheDataset = TypeVar("TAvalancheDataset", bound="AvalancheDataset")
+TDataWTransform = TypeVar("TDataWTransform", bound="_FlatDataWithTransform")
 
 
-class AvalancheDataset(FlatData[T_co]):
+class AvalancheDataset(IDataset[T_co]):
     """Avalanche Dataset.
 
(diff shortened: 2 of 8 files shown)


lrzpellegrini:
Please correct me if I'm wrong. With this PR AvalancheDataset will now have two main components/fields:
- flat_data: in charge of managing transformations
- data_attributes: in charge of data attributes

with each field managing its own subset operations. If this is the case, it seems reasonable.

As for the tests, `custom_streams_name_and_length` seems to fail correctly: an object should be equal to itself, especially if it is the very same object (with the same memory address). The equality check method may need a fix before merging.

PR #1387, opened 2023-05-25
added topk patch

scumatteo:
If top-k metric is called with num_classes < k, top-k metric returns 0.


AntonioCarta:
It doesn't make a lot of sense to return 0 to mask an error, especially at the metric-level. IMO the plugin (not the metric) should not emit values when `num_classes < k`. If called, the metric should still fail.

PR #1388, opened 2023-05-25
Add a variable to control whether to retain the graph or not in the backward pass

BenCrulis:
With the addition of this variable, you can decide to keep the graph by setting `strategy.retain_graph = True` so that plugins can do their own backward pass after to the one triggered automatically without having to execute the forward pass again.

Unsatisfactory alternatives:
- subclass BaseSGDTemplate and change the implementation of its `backward()` method
- monkey patch the `backward()` method
- call `strategy.forward()` each time a second backward pass has to be made in the same iteration

Note that perhaps the name of this variable should be changed to something else.

### DIFF
--- a/avalanche/training/templates/base_sgd.py
+++ b/avalanche/training/templates/base_sgd.py
@@ -114,6 +114,9 @@ class BaseSGDTemplate(
         )
         """ Eval mini-batch size. """
 
+        self.retain_graph: bool = False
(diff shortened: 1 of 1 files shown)


AntonioCarta:
Thanks. I didn't consider the fact that plugin may need to set retain_graph, which makes subclassing useless. I think your solution is ok.

PR #1396, opened 2023-05-30
BugFix OnlineTemplate missing model adaptation call

AlbinSou:
When access_task_boundaries = False, the model adaptation was not called, I fixed it.

@AntonioCarta Another thing that I could fit there -> We had spoken about an option to reset the optimizer state or not. Right now by default it's False but their is no real way to change it apart when calling self.make_optimizer. Should I add this option to the SupervisedTemplate ?

Also I would like to fix the unit tests that don't pass sometimes but I'm not able to reproduce the bugs locally ...

### DIFF
--- a/avalanche/evaluation/metrics/cumulative_accuracies.py
+++ b/avalanche/evaluation/metrics/cumulative_accuracies.py
@@ -58,6 +58,10 @@ class CumulativeAccuracy(Metric[Dict[int, float]]):
             raise ValueError("Size mismatch for true_y " 
                              "and predicted_y tensors")
         for t, classes in classes_splits.items():
+            # This is to fix a weird bug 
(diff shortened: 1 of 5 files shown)


AntonioCarta:
> We had spoken about an option to reset the optimizer state or not. Right now by default it's False but their is no real way to change it apart when calling self.make_optimizer. Should I add this option to the SupervisedTemplate ?

makes sense to add it as an option. You can add it as a `train` argument as we do for dataloader arguments.

AlbinSou:
I tried to fix the test for cumulative accuracies ... But it's more that I'm waiting for magic to happen honestly. I have no idea why this fails sometimes. Let's see if the new one also fails. I replaced the assertDictEqual with multiple assertEquals.

PR #1397, opened 2023-05-31
Specialize benchmark creation helpers

### LINKED ISSUE #774: Core50 benchmark with ICARL strategy
ICarl strategy returns an error  with Core50 benchmark:
`scenario = CORe50(scenario="nc", run=9)`

The error is:
`AttributeError: 'GenericCLScenario' object has no attribute 'n_classes_per_exp'`

Other nc benchmarks belong to "NCScenario"  class and this issue does not appear.

lrzpellegrini:
This PR introduces the ability to customize the class of the benchmark objects returned by generic benchmark creation functions (such as `dataset_benchmark`, `paths_benchmark`, ...).

In addition, new problem-specific functions (`dataset_classification_benchmark`, `dataset_detection_benchmark`, ...) are introduced to explicitly cover classification and detection setups. Generic functions will still return `ClassificationScenario` instances if datasets with classification targets are detected (ints), but they now will display a warning suggesting the use of its classification counterpart.

This PR also fixes a problem with CORe50-NC, which did not have fields found in NCScenario.

Fixes #774


avalanche/benchmarks/scenarios/generic_benchmark_creation.py
```diff
@@ -28,35 +31,284 @@
...
+        benchmark_factory_compat = benchmark_factory
+
+    return dataset_factory_compat, benchmark_factory_compat
 
 
 def create_multi_dataset_generic_benchmark(
-    train_datasets: Sequence[SupportedDataset],
-    test_datasets: Sequence[SupportedDataset],
+    train_datasets: Sequence[GenericSupportedDataset],
+    test_datasets: Sequence[GenericSupportedDataset],
```
AntonioCarta, line 290: can we simplify this as require an AvalancheDataset? this way we remove the necessity to pass transform/task labels...

AntonioCarta:
IMO reusing the same organization of the classification generators is a bit clunky for the generic ones. If we assume that the input is an AvalancheDataset we can remove the transform arguments. IMO the generic generators should:
- create a stream/benchmark from a list of AvalancheDatasets
- create a stream/benchmark given an AvalancheDataset and an attribute to use for splitting it (e.g. a domain label which is a DataAttribute)

avalanche/benchmarks/utils/detection_dataset.py
```diff
@@ -138,37 +140,6 @@ def targets_task_labels(self) -> DataAttribute[int]:
...
-    task_labels: Union[int, Sequence[int]],
-    targets: Sequence[TTargetType],
-    collate_fn: Optional[Callable[[List], Any]] = None
-) -> SupervisedDetectionDataset:
-    ...
-
-
-@overload
 def make_detection_dataset(
     dataset: SupportedDetectionDataset,
```
AntonioCarta, line 144: do we need this? DetectionDataset should be enough. We have the others for classificaiton mostly to keep it compatible with the old API.

avalanche/benchmarks/utils/detection_dataset.py
```diff
@@ -359,43 +300,6 @@ def _detection_class_mapping_transform(class_mapping, example_target_dict):
...
-    task_labels: Union[int, Sequence[int]],
-    targets: Sequence[TTargetType],
-    collate_fn: Optional[Callable[[List], Any]] = None
-) -> SupervisedDetectionDataset:
-    ...
-
-
-@overload
 def detection_subset(
     dataset: SupportedDetectionDataset,
```
AntonioCarta, line 304: do we need this? DetectionDataset should be enough. We have the others for classificaiton mostly to keep it compatible with the old API.

avalanche/benchmarks/utils/detection_dataset.py
```diff
@@ -660,7 +472,7 @@ def concat_detection_datasets(
         Sequence[TTargetType], Sequence[Sequence[TTargetType]]
     ]] = None,
     collate_fn: Optional[Callable[[List], Any]] = None
-) -> Union[DetectionDataset, SupervisedDetectionDataset]:
+) -> DetectionDataset:
```
AntonioCarta, line 475: do we need this? DetectionDataset should be enough. We have the others for classificaiton mostly to keep it compatible with the old API.

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
         return self.data.targets_task_labels  # type: ignore
 
 
+def numpy_is_sequence_int(numpy_tensor: np.ndarray) -> bool:
```
AntonioCarta, line 677: private?

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+    return issubclass(numpy_tensor.dtype.type, np.integer)
+
+
+def numpy_is_single_int(numpy_tensor: np.ndarray) -> bool:
```
AntonioCarta, line 681: private?

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+    try:
+        single_value = numpy_tensor.item()
+        return isinstance(single_value, int)
+    except ValueError:
+        return False
+
+
+def torch_is_sequence_int(torch_tensor: Tensor) -> bool:
```
AntonioCarta, line 689: private?

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+    return not torch.is_floating_point(torch_tensor) and \
+        not torch.is_complex(torch_tensor)
+
+
+def torch_is_single_int(torch_tensor: Tensor) -> bool:
```
AntonioCarta, line 694: private?

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+    try:
+        single_value = torch_tensor.item()
+        return isinstance(single_value, int)
+    except ValueError:
+        return False
+    
+
+def element_is_single_int(element: Any):
```
AntonioCarta, line 702: private?

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+    if isinstance(element, (int, np.integer)):
+        return True
+    if isinstance(element, Tensor):
+        return torch_is_single_int(element)
+    else:
+        return False
+
+
+def is_int_iterable(iterable: Iterable[Any]):
```
AntonioCarta, line 711: private?

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+        for t in iterable:
+            if not element_is_single_int(t):
+                return False
+        return True
+    
+
+AnyT = TypeVar('AnyT', bound=Iterable)
+
+
+def to_int_list(iterable: AnyT, force: bool = True) -> Union[AnyT, List[int]]:
```
AntonioCarta, line 726: private?

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+    else:
+        return None
+
+    if indices is not None:
+        found_targets = SubSequence(found_targets, indices=indices)
+
+    return found_targets
+
+
+def make_generic_dataset(
```
AntonioCarta, line 831: do we need it? also, it's not generic because it's still asks for targets

avalanche/benchmarks/utils/utils.py
```diff
@@ -673,12 +674,293 @@ def _get_task_labels_field(self) -> DataAttribute[int]:
...
+        collate_fn=collate_fn,
+    )
+    
+    if initial_transform_group is not None:
+        return data.with_transforms(initial_transform_group)
+    else:
+        return data
+
+
+def make_generic_tensor_dataset(
```
AntonioCarta, line 904: do we need it?

AntonioCarta (changes requested):

AntonioCarta:
the same goes for the dataset methods. We don't really need those. Instead of doing
```
data = make_generic_tensor_data(...)
```
I think it's easier to do
```
data = TensorDataset(...) # pytorch data
data = AvalancheDataset(data)
```
We don't need to know how the dataset is made internally.

avalanche/benchmarks/utils/utils.py
lrzpellegrini: I think it's still useful to load targets and task labels if they are available. Targets and task labels are optional (it's a best effort search).

lrzpellegrini:
We have `DatasetScenario`, which is quite generic. The `ClassificationScenario` is a subclass of DatasetScenario with a couple of additional fields regarding the classes timeline.

`DatasetScenario` -> `ClassesTimelineCLScenario` (adds class timeline fields) -> `ClassificationScenario`

similarly, for detection:

`DatasetScenario` -> `ClassesTimelineCLScenario` -> `DetectionScenario`

In theory, one could create a regression/"other problem" benchmark using `DatasetScenario` instead of `ClassificationScenario`, which is more correct.

avalanche/benchmarks/utils/detection_dataset.py
lrzpellegrini: It's also a matter of allowing:

- On the Avalanche side, to automatically fetch the `targets` from the dataset
- On the user side, to set the targets and task labels

The alternative option is setting the targets and task labels through `data_attributes`, but that seems an advanced topic.

AntonioCarta:
> We have DatasetScenario, which is quite generic

This is a class, I'm referring to benchmark generators. We don't have generic ones because they all expect targets and task labels.

> In theory, one could create a regression/"other problem" benchmark using DatasetScenario instead of ClassificationScenario, which is more correct.

This is the kind of API that I want to avoid. For every type of dataset/problem and every type of CL scenario we will need a different generator, or a super general and super complex one.

> ClassesTimelineCLScenario (adds class timeline fields) 

keep in mind that we don't really have a use case for timeline fields. I think it's best to keep the benchmark itself as simple as possible (i.e. a dumb collection of streams). Same for streams (a dumb collection of experiences).

The idea of splitting dataset and stream creation tries to reduce this complexity by splitting separate concerns in different methods. I'm open to other proposals, but only in the direction of reducing complexity.

lrzpellegrini:
I think it may be a good idea to re-design the benchmark generation part a bit. I'm putting this PR on hold until we devise a more general solution.

PR #1405, opened 2023-06-05
NCM Classifier uses dictionary and can update means

AndreaCossu:
The NCM Classifier was quite cumbersome to use. Now it explicitly keeps a dictionary of means for each class and has a method to update the means. 
The means are vectorized in a tensor after each update.

Tests and strategies using NCM have been updated.

### DIFF
--- a/avalanche/models/ncm_classifier.py
+++ b/avalanche/models/ncm_classifier.py
@@ -1,5 +1,6 @@
 import torch
-from torch import nn
+from torch import nn, Tensor
+from typing import Dict
 
 
 class NCMClassifier(nn.Module):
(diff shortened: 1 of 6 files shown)


avalanche/models/ncm_classifier.py
```diff
@@ -10,22 +10,82 @@ class NCMClassifier(nn.Module):
...
+            classes must be zero-indexed.
         :param normalize: whether to normalize the input with
             2-norm = 1 before computing the distance.
         """
         super().__init__()
-        self.class_means = class_mean
+        assert isinstance(class_means_dict, dict), \
+            "class_means_dict must be a dictionary mapping class_id " \
+            "to mean vector"
+        self.class_means_dict = class_means_dict
```
AntonioCarta, line 24: tensors in `class_means_dict` should be registered as buffers

avalanche/models/ncm_classifier.py
```diff
@@ -10,22 +10,82 @@ class NCMClassifier(nn.Module):
...
+
+        # (num_classes, batch_size)
+        sqd = torch.cdist(self.class_means.to(x.device), x)
+        # (batch_size, num_classes)
         return (-sqd).T
 
+    def update_class_means_dict(self, class_means_dict):
+        """
+        Update dictionary of class means.
+        If class already exists, the average of the two mean vectors
```
AntonioCarta, line 74: shouldn't NCM keep track of counters for each class?

avalanche/models/ncm_classifier.py
AndreaCossu: I can do that. The dictionary itself cannot be registered though. Is this a problem for check-pointing? 
In case, I guess I can implement a method that fills the dictionary with the current buffers after each NCM update.

avalanche/models/ncm_classifier.py
AndreaCossu: The method can accept both a dictionary of means and a dictionary of class counters. If the user wants to compute the arithmetic mean I can simply multiply the current class mean by the current class counter to obtain the original sum and then update the mean with the new mean and class counter provided by the user. 
However this works only for arithmetic mean. For example, SCR uses L2-normalized means, not arithmetic means.

avalanche/models/ncm_classifier.py
AntonioCarta: > Is this a problem for check-pointing?

I'm not sure, maybe add a test that uses `load_state_dict` and `save_state_dict` to make sure buffers are detected correctly.

AndreaCossu: I had to override the default load but now loading/saving works.

AntonioCarta: can you add a test just for safety?

AndreaCossu: Done

AntonioCarta:
@AndreaCossu there is a broken test but I'm not sure how it's related to your PR.

@lrzpellegrini it looks like a distributed test. Can you give it a look?

AndreaCossu:
I ran again the failed test and it worked this time.

AntonioCarta:
ok, I guess we have to investigate this problem separately.

PR #1408, opened 2023-06-08
removed drop_last forwarding from make_eval_dataloader

AlbinSou:
Before that, when calling

strategy.train(experience, eval_streams=[scenario.val_stream], drop_last=True)

The drop_last argument was used both in the training and in the evaluation dataloader. This is a quickfix to fix this behavior, it would be nice to find a better way of distinguishing kwargs that are forwarded to eval/train loaders, or even to functions outside of loaders.

### DIFF
--- a/avalanche/training/templates/base_sgd.py
+++ b/avalanche/training/templates/base_sgd.py
@@ -430,7 +430,6 @@ class BaseSGDTemplate(
         shuffle=False,
         pin_memory=None,
         persistent_workers=False,
-        drop_last=False,
         **kwargs
     ):
         """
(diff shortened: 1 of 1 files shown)


AlbinSou:
> looks like `test_cumulative_accuracy` is failing again

I know and I still have no clue why ... Still works perfectly locally (contrarily to the topk test which fails locally in my case). If only I could reproduce the error locally that would help a lot.

AntonioCarta:
please try to fix this. You can try to create the same env that we use in the CI by creating a fresh environment (or using the docker image we use for tests).

PR #1414, opened 2023-06-13
Replay flattening

AntonioCarta:


AntonioCarta:
a very hacky fix but should solve the performance issues that @AlbinSou and @AndreaCossu have noticed.

PR #1416, opened 2023-06-14
faster indices for FlatData with lazy operations

AntonioCarta:
Lazy indices for FlatData to support faster operations.

### DIFF
--- a/avalanche/benchmarks/utils/flat_data.py
+++ b/avalanche/benchmarks/utils/flat_data.py
@@ -104,6 +104,36 @@ class LazyIndices:
             return sum(len(ll) for ll in self._lists)
 
 
+class LazyRange(LazyIndices):
+    """Avoid 'eagerification' step for ranges."""
+
+    def __init__(self, start, end, offset=0):
+        self._start = start
+        self._end = end
+        self._offset = offset
+        self._known_length = end - start
+        self._eager_list = self
+
+    def _to_eager(self):
+        # LazyRange is eager already
+        pass
+
+    def __iter__(self):
+        for i in range(self._start, self._end):
+            yield self._offset + i
+
+    def __getitem__(self, item):
+        assert item >= self._start and item < self._end, "LazyRange: index out of range"
+        return self._start + self._offset + item
+
+    def __add__(self, other):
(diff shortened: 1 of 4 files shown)


AntonioCarta:
@lrzpellegrini do you have any suggestion on how to check for performance (runtime) regression with automated tests?

lrzpellegrini:
Everything seems ok!

As for the performance, it's hard to check them in an uncontrolled environment (such as GH Actions), where the processor may change. Our best shot is to run performance checks on ContinualAI servers. There are ways to tell GitHub to run actions on external runners, but I don't know if it is hard to set up.

As for the PR, I wonder if we can further improve performance by storing the eager indices in PyTorch or NumPy tensors.

avalanche/benchmarks/utils/flat_data.py
```diff
@@ -44,6 +44,66 @@
...
+            whenever we generate them from the lazy sequence.
+            """
+            self._offset = offset
+
+        self._known_length = known_length
+
+    def _to_eager(self):
+        if self._eager_list is not None:
+            return
+        self._eager_list = [el + self._offset for el in self._lazy_sequence]
```
lrzpellegrini, line 84: Maybe we can further optimize using NumPy both as a storage for _eager_list and both to optimize summing the _offset to each index.

AntonioCarta:
> As for the PR, I wonder if we can further improve performance by storing the eager indices in PyTorch or NumPy tensors.

Maybe. Most of the cost was the concatenation of large lists for each concat operation. It should be more efficient in numpy.
For example, I added a `LazyRange` which should help avoid the "eagerification" step in some common cases.

I think we can speed this up further in the future, either pushing more on the lazyness or using numpy tensors.

AntonioCarta:
I will leave further improvements as future work.

avalanche/benchmarks/utils/flat_data.py
AntonioCarta: This is useful if we do it every time we offset (there are many places). Here most cost is due to the full iteration of the iteration. Not sure numpy would help a lot. But in general, converting everything to tensors may speed up general subsets and concat ops.

PR #1424, opened 2023-06-19
Rar merge

JuliousHurtado:
This is the first version of the implementation of Retrospective Adversarial Replay (RAR) for Continual Learning.

I tested on CIFAR10 using RAR, and compared it with ER (as in the original paper). RAR improves performance over simple ER.

The exact results shown in the paper could not be replicated, mainly due to the fact that the ER shows much higher accuracy and no original code is provided.

| Memory size | ER             | ER + RAR       |
|-------------|----------------|----------------|
| 200         | 32.76% + 4.03% | 37.62% + 2.46% |
| 500         | 37.85% + 2.61% | 43.43% + 2.40% |


AntonioCarta:
Thanks @JuliousHurtado. Can you fix the syntax error and adjust the doc a little bit?

PR #1426, opened 2023-06-20
small refactor and cumulative accuracy fix

AlbinSou:
I noticed cumulative accuracy was not working when the benchmark option class_idx_from_zero_from_first_exp was not activated. I fixed it. I also added some functions that I was using here and there to the training/utils.py as well as a new function to detect if the experience is at task boundary (and thus needs to be used for adaptation) depending on several attributes (access_task_boundaries, is_fist_subexp, is OnlineCLExperience etc.. ).


AntonioCarta:
Looks like the test is still failing.

AlbinSou:
> Looks like the test is still failing.

Yes, for the same reason than before. The fix I did was not related to the test failure. I still don't know why it fails since I still cannot reproduce the error locally.

PR #1428, opened 2023-06-22
Fixed icarl by converting back TrainEvalModel to DynamicModule

AndreaCossu:
The behavior of TrainEvalModel was not always consistent and the classifier sometimes failed to switch.
I brought back the DynamicModule version, this fixes iCarl performance.

### DIFF
--- a/avalanche/models/dynamic_modules.py
+++ b/avalanche/models/dynamic_modules.py
@@ -14,7 +14,7 @@ networks, ...).
 """
 import torch
 from torch.nn import Module
-import numpy as np
+from typing import Optional
 
 from avalanche.benchmarks.utils.flat_data import ConstantSequence
 from avalanche.benchmarks.scenarios import CLExperience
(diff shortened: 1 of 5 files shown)


AntonioCarta:
This is still broken though. You shouldn't have to adapt again on already seen experiences.

AndreaCossu:
I agree, I am working on fixing this but for the moment I just wanted to make iCARL work. I need to further inspect why the flag `self.training` was not switching properly.

AndreaCossu:
The last commits should have fixed the missing switch between train and eval in iCarl. The behavior is pretty strange. Basically, before the change the model started in training mode, then it switched to eval mode after the first 'strategy.eval' call and then it stayed in eval forever.
I had to remove both the 'old_model.eval()' call in the icarl loss and add the 'strategy.model.train()' in the icarl plugin. 

This fixes the switch but it is suspicious, the model should automatically switch back to training mode each time 'strategy.train' is called.

AntonioCarta:
> the model should automatically switch back to training mode each time 'strategy.train' is called.

It's a bit different, `strategy.eval` stores the mode for each module and reset them at the end of the method.

PR #1429, opened 2023-06-23
Streamline data loaders

lrzpellegrini:
As part of the effort to add support for FFCV data loading, this PR brings an overall change in the way data loaders are created.

Avalanche data loaders are used to return batches which are the concatenation of smaller mini-batches from multiple datasets. 

Up to now, this was accomplished by:
- Creating a data loader for each dataset (each with the same parameters!)
- Manually looping over all loaders, fetching the sub-mini-batches  
- Collate the sub-mini-batches before returning them (in the main thread!)

This led to issues regarding `num_workers` (which were spawned for each sub-dataset thus leading to spawning `N * num_workers`), bottlenecks related to the collate on the main thread, etc, not to mention that the code of each loader replicated operations already implemented in other loaders.

With the improvements found in this PR, a single PyTorch DataLoader is created based on a newly implemented Sampler that is used to return the indices in the desired order. The new mechanism supports distributed sampling, ensures that the correct number of workers is created, and that flags like `persistent_workers` and `pin_memory` have appreciable results. In addition, minibatch collate is done at the DataLoader level, thus removing a problematic bottleneck.

Note: experiment results may slightly change w.r.t. the previous implementation. This is not a bug and is related to the way random number generators are consumed, which may lead to a different choice of replay examples in the ReplayPlugin. Setting `shuffle=False` or setting a manual seed in the ReplayPlugin update method leads to the same results as the previous loader implementations.

### DIFF
--- a/avalanche/benchmarks/utils/classification_dataset.py
+++ b/avalanche/benchmarks/utils/classification_dataset.py
@@ -87,7 +87,7 @@ def lookup(indexable, idx):
 class ClassificationDataset(AvalancheDataset[T_co]):
 
     @property
-    def task_pattern_indices(self):
+    def task_pattern_indices(self) -> Dict[int, Sequence[int]]:
         """A dictionary mapping task ids to their sample indices."""
         return self.targets_task_labels.val_to_idx  # type: ignore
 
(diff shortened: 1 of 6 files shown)


AntonioCarta:
Thanks, this is much cleaner than the previous solution.

PR #1446, opened 2023-07-04
Fix MER

AlbinSou:
MER template was not updated and a test was missing. Unfortunately, the test fails even after fixing the template. @HamedHemati could you look into this ?

Right now I get the following error:

```
  File "/home/albin/Documents/Projects/avalanche_dev/avalanche/training/supervised/mer.py", line 119, in _inner_updates
    x, y, t = self.buffer.get_batch(x, y, t)
  File "/home/albin/Documents/Projects/avalanche_dev/avalanche/training/supervised/mer.py", line 37, in get_batch
    [self.storage_policy.buffer[i][0].unsqueeze(0) for i in rnd_ind]
  File "/home/albin/Documents/Projects/avalanche_dev/avalanche/training/supervised/mer.py", line 37, in <listcomp>
    [self.storage_policy.buffer[i][0].unsqueeze(0) for i in rnd_ind]
  File "/home/albin/Documents/Projects/avalanche_dev/avalanche/benchmarks/utils/data.py", line 376, in __getitem__
    elem.append(da[idx])  # type: ignore
AttributeError: '_FlatDataWithTransform' object has no attribute 'append'
```

But I think there might be more once this is fixed.

I also renamed parameters linked to memory to mem_size and batch_size_mem to be consistent with other methods using memory.

### DIFF
--- a/avalanche/training/supervised/__init__.py
+++ b/avalanche/training/supervised/__init__.py
@@ -14,3 +14,4 @@ from .der import DER
 from .l2p import LearningToPrompt
 from .supervised_contrastive_replay import SCR
 from .expert_gate import ExpertGateStrategy
+from .mer import MER
--- a/avalanche/training/supervised/mer.py
+++ b/avalanche/training/supervised/mer.py
@@ -138,7 +138,6 @@ class MER(SupervisedMetaLearningTemplate):
 
             self.model.load_state_dict(load_dict, strict=False)
 
-
     def _outer_update(self, **kwargs):
         w_aft = self.model.state_dict()
 
--- a/avalanche/training/supervised/mer.py
+++ b/avalanche/training/supervised/mer.py
@@ -1,22 +1,20 @@
-from typing import Callable, Sequence, Optional, Union
+from copy import deepcopy
+from typing import Callable, Optional, Sequence, Union
 
 import torch
 import torch.nn.functional as F
-from torch.nn import Module, CrossEntropyLoss
+from torch.nn import CrossEntropyLoss, Module
 from torch.optim import Optimizer
-from copy import deepcopy
 
-from avalanche.training.plugins import SupervisedPlugin, EvaluationPlugin
-from avalanche.training.plugins.evaluation import default_evaluator
-from avalanche.training.templates import SupervisedMetaLearningTemplate
 from avalanche.models.utils import avalanche_forward
+from avalanche.training.plugins import EvaluationPlugin, SupervisedPlugin
+from avalanche.training.plugins.evaluation import default_evaluator
 from avalanche.training.storage_policy import ReservoirSamplingBuffer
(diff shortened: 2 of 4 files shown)


HamedHemati:
Hey @AlbinSou . The problem should be fixed by converting `rnd_ind` in the `get_batch()` function  of `MERBuffer` to a list:
`rnd_ind = torch.randperm(len(self))[:bsize].tolist()`
`FlatData` indexing doesn't work with PyTorch tensors since an update from approx. 1 month ago.

AlbinSou:
> Hey @AlbinSou . The problem should be fixed by converting `rnd_ind` in the `get_batch()` function of `MERBuffer` to a list: `rnd_ind = torch.randperm(len(self))[:bsize].tolist()` `FlatData` indexing doesn't work with PyTorch tensors since an update from approx. 1 month ago.

Ok, thank you @HamedHemati it fixed it indeed. I also fixed another issue since so far the outer updates were performed by loading the state dict and sometimes it was causing some issues related to non-learnable keys (active units of the multiclass classifier in that case). So I changed it by only updating learnable keys that are in model.named_parameters, I guess this would also have been a problem with the batch normalization running parameters.

By the way, about batch normalization, I think this method still needs a change if we want it to work with batch normalization, because I am not sure it can work with the current implementation, the main problem being that examples are forwarded one by one into the model, using potentially very volatile batch statistics. I think it would be nicer to set the batch statistics to the same value for all of the internal updates and then unset it, but it's a bit tricky to do ... I don't know if it's worth doing this change.

HamedHemati:
> I also fixed another issue since so far the outer updates were performed by loading the state dict and sometimes it was causing some issues related to non-learnable keys (active units of the multiclass classifier in that case). So I changed it by only updating learnable keys that are in model.named_parameters, I guess this would also have been a problem with the batch normalization running parameters.

I see, thanks for the fix. I had tested it only with static models.

> By the way, about batch normalization, I think this method still needs a change if we want it to work with batch normalization, because I am not sure it can work with the current implementation, the main problem being that examples are forwarded one by one into the model, using potentially very volatile batch statistics. I think it would be nicer to set the batch statistics to the same value for all of the internal updates and then unset it, but it's a bit tricky to do ... I don't know if it's worth doing this change.

That's true. In general that's a problem with any OCL strategy that uses `batch_size=1` with a model that requires batch statistics. I would keep the AVL implementation as it is, and add the batch-norm correction in a subclass of MER if needed.

AlbinSou:
@AntonioCarta I also included a fix to SCR (setting the model and dataset transform to eval mode before updating the ncm classifier)

AntonioCarta:
@AlbinSou did you check the results for SCR? I remember @AndreaCossu had some issues with the train/eval mode but I thought it was fixed already.

AlbinSou:
@AntonioCarta Yes, I checked the results and they are much better than with my previous trials.

PR #1447, opened 2023-07-04
Fix integration with IncrementalClassifier

JuliousHurtado:
When testing the RAR implementation using a model with `IncrementalClassifier` class, @AlbinSou found a new error.

The problem arises between an incompatibility between `IncremetnalClassifier` and the `feature_extraction` functions from pytorch. Because of this, I changed the mask function in the `IncrementalClassifier` class to be an in_place operation. This solves the first problem.

A second problem was a `KeyError` problem with `optimizer.state_dict()`. Because of this, I simplified the RAR implementation. This can help with the performance but sometimes take more time to run.

### DIFF
--- a/avalanche/models/dynamic_modules.py
+++ b/avalanche/models/dynamic_modules.py
@@ -265,7 +265,8 @@ class IncrementalClassifier(DynamicModule):
         """
         out = self.classifier(x)
         if self.masking:
-            out[..., torch.logical_not(self.active_units)] = self.mask_value
+            mask = torch.logical_not(self.active_units)
+            out.masked_fill_(mask=mask, value=self.mask_value)
         return out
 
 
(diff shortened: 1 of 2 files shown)


AntonioCarta:
Is the in-place operation compatible with backpropagation? Can you use the [not in-place](https://pytorch.org/docs/stable/generated/torch.Tensor.masked_fill_.html#torch.Tensor.masked_fill_) version of the operation instead?

JuliousHurtado:
I don't know if it is compatible con back-propagation, I don't see why not.
But yes, I change it. I run some small experiments with both versions and I don't see any difference in the performance.

Something that Albin suggest was to change the RAR plugins to something similar to ICarl, and having the feature extractor and classifier independently. This would help us remove the `name_ext_layer` parameters, since we will have a feature extractor that can return the vector. Don't know what you think?

PR #1464, opened 2023-07-19
Speedup for long benchmarks iteration

lrzpellegrini:
A simple fix to speedup benchmark creation and stream iteration.

The impact is greatly noticeable in long benchmarks (such as SplitImagenet, n_experiences=1000), where the iteration time on the training stream drops from 1h to 1m. I also tried other optimizations regarding FlatData, but this seems to be the most straightforward solution.

### DIFF
--- a/avalanche/benchmarks/scenarios/lazy_dataset_sequence.py
+++ b/avalanche/benchmarks/scenarios/lazy_dataset_sequence.py
@@ -244,9 +244,11 @@ class LazyDatasetSequence(Sequence[TCLDataset]):
                 )
 
             self._loaded_experiences[exp_id] = generated_exp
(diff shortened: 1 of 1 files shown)


AntonioCarta:
Thanks, this is great! does it impacts also smaller benchmarks such as CIFAR100?

lrzpellegrini:
For SplitCIFAR100, n_exps=100, the following loop goes from 6.22 to 1.50 seconds:

```python
start_time = time.time()
benchmark_instance = SplitCIFAR100(100)
for exp in benchmark_instance.train_stream:
    dataset = exp.dataset
end_time = time.time()
print(end_time-start_time)
```

PR #1476, opened 2023-07-26
Update FZTH Notebooks for Release Version 0.4

HamedHemati:
All FZTH notebooks work with the master master.

@AntonioCarta Are there any new showcase examples for the new features that we might need to add? I can't remember if there is something new that needs to be added to the existing notebooks.


AntonioCarta:
No, it's okay as is.

PR #1487, opened 2023-07-31
Earlystopping patch

scumatteo:
Solved problem by moving the reset in before_training_exp, to reload always the model adapted at the current experience.


AntonioCarta:
Thank you

PR #1496, opened 2023-09-04
Implement PackNet

tachyonicClock:
- The PackNet plugin is implemented inline with the PackNet module implementation. Is that okay?

### DIFF
--- a/avalanche/models/packnet.py
+++ b/avalanche/models/packnet.py
@@ -1,7 +1,7 @@
 """
 Implements (Mallya & Lazebnik, 2018) PackNet algorithm for fixed-network
 parameter isolation. PackNet is a task-incremental learning algorithm that
-uses task identities to isolate parameters.
+uses task identities during testing.
 
 Mallya, A., & Lazebnik, S. (2018). PackNet: Adding Multiple Tasks to a
 Single Network by Iterative Pruning. 2018 IEEE/CVF Conference on Computer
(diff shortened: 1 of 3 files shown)


avalanche/models/packnet.py
```diff
@@ -0,0 +1,527 @@
...
+from torch.nn import functional as F
+from abc import ABC, abstractmethod
+from enum import Enum
+from avalanche.core import BaseSGDPlugin, Template
+from avalanche.models.dynamic_modules import MultiTaskModule
+from avalanche.models.simple_mlp import SimpleMLP
+from avalanche.training.templates.base_sgd import BaseSGDTemplate
+
+
+class PackNetModule(ABC):
```
AntonioCarta, line 26: when you call it `Module` you mean `nn.Module`? if so, it should inherit from it. Otherwise I suggest to rename it for clarity.

avalanche/models/packnet.py
```diff
@@ -0,0 +1,527 @@
...
+    def task_count(self) -> int:
+        return int(self._task_count.item())
+
+    @property
+    def device(self) -> torch.device:
+        return self.weight.device
+
+    def _remove_gradient_hook(self, grad: Tensor) -> Tensor:
+        """
+        Only the top layer should be trained. Todo so all other gradients
```
AntonioCarta, line 210: Can we do it differently if we only need to backprop on the classifier? for example, `h.detach()` when we don't need the gradients? it would be more efficient.

avalanche/models/packnet.py
```diff
@@ -0,0 +1,527 @@
...
+            w.bias,
+            w.stride,
+            w.padding,
+            output_padding,
+            w.groups,
+            w.dilation,
+        )
+
+
+def PackNetSimpleMLP(
```
AntonioCarta, line 343: since it's a function it's better to call it `packnet_simple_mlp`.

AntonioCarta: also add documentation

avalanche/models/packnet.py
```diff
@@ -0,0 +1,527 @@
...
+        elif isinstance(wrappee, nn.Sequential):
+            # Wrap each submodule
+            for i, x in enumerate(wrappee):
+                wrappee[i] = PackNet.wrap(x)
+        else:
+            for submodule_name, submodule in wrappee.named_children():
+                setattr(wrappee, submodule_name, PackNet.wrap(submodule))
+        return wrappee
+
+    def __init__(self, wrappee: nn.Module) -> None:
```
AntonioCarta, line 391: add docs

avalanche/models/packnet.py
```diff
@@ -0,0 +1,527 @@
...
+            # applications to their children
+            if isinstance(module, PackNetModule) and not isinstance(module, PackNet):
+                func(module)
+
+        self.apply(__pn_apply)
+
+    def prune(self, to_prune_proportion: float) -> None:
+        """
+        Prunes the layer by removing the smallest weights and freezing them.
+        Biases are frozen as a side-effect.
```
AntonioCarta, line 412: add docs for all the arguments (for all the other methods too).

AntonioCarta (changes requested):

AntonioCarta:
Hi @tachyonicClock, thanks for the PR. To add new strategies we require a script reproducing the paper's results on some benchmark to add to continual-learning-baselines.

Please remember to add the doc to all the arguments and methods.

> The PackNet plugin is implemented inline with the PackNet module implementation. Is that okay?

That is ok. It is a bit convoluted because the training logic is mixed between the module wrapper and the plugin but I don't think that's an issue.

avalanche/models/packnet.py
tachyonicClock: I don't believe this is possible, because we still need many of the gradients i.e `task_index` is the same shape as the grad/weights.

tachyonicClock:
Hi @AntonioCarta thanks for taking the time to code review. Regarding the "continual-learning-baselines" project would it be reasonable to match a survey rather than the original paper? The original paper uses a non-standard multi-dataset problem where each task is an entire dataset.

AntonioCarta:
Hey @tachyonicClock, let me know when the benchmark is ready and I can do another review of the PR.

tachyonicClock:
Absolutely.  I haven't forgotten about it :D, I've just been busy with other things.

tachyonicClock:
I have made a pull request in the continual-learning-baselines project (https://github.com/ContinualAI/continual-learning-baselines/pull/63). My implementation (46% on Split Tiny ImageNet) does not match the performance achieved in [1] (49% on Split Tiny ImageNet).  Perfectly matching [1] is challenging because their benchmark used the Continual Hyperparameter Selection Framework, rather than published static hyper-parameters.

[1] Delange, M., Aljundi, R., Masana, M., Parisot, S., Jia, X., Leonardis, A., Slabaugh, G., & Tuytelaars, T. (2021). A continual learning survey: Defying forgetting in classification tasks. IEEE Transactions on Pattern Analysis and Machine Intelligence, 1–1. https://doi.org/10.1109/TPAMI.2021.3057446

AntonioCarta:
Thanks, I think your result is ok. Hyperparameters chosen with the "Continual Hyperparameter Selection" can be optimized for each timestep, so it makes sense that your results are slightly lower.

I'm going to merge this, please open the PR at continual-learning-baselines for the reproducibility script.

PR #1511, opened 2023-10-05
WIP: New benchmarks API

AntonioCarta:
As we discussed in our internal meetings. 
Still a WIP. Right now, you can take a look at the fzth benchmark notebook to see how it looks and comment on the API.
Let me know if you have some feedback.

If there are no particular concerns, the rest should be ready in ~2/3 weeks.

Regarding the internal changes, I will say more later but notice:
- old scenarios/generators have been moved to `benchmarks.scenarios.deprecated`
- the old `benchmarks.generators` module is remove, there is no strict distinction between benchmarks and scenarios right now

Lots of changes, but most are just due to moving stuff around. Again, I suggest to just check the notebook for the moment.


HamedHemati:
Hi @AntonioCarta 

As you suggested, I tried to run the FZTH notebook for the benchmarks. Here are a few points/questions from my side:

- I saw that the attributes of an experience can have different access modes for each phase: train/eval /logging
I also tried  `exp.current_experience`, and it still works. Do we want to keep the "public" access  to the attributes for backward compatibility reasons? 

- I encountered an issue when trying to run `class_incremental_benchmark({'train': train_MNIST, 'test': test_MNIST}, num_experiences=5)`. I  get the error message that "ValueError: AvalancheDataset already has task labels. Use `benchmark_from_datasets` instead or set `reset_task_labels=True`". 
`AvalancheDataset` doesn't seem to have a `reset_task_label` attribute. Is it possible to wrap a dataset with targets but without task-labels directly? or do should we always first create a classification dataset and then remove the task labels?

- The command `split_online_stream(bm.train_stream, experience_size=10)` also doesn't work. Is it supposed to changed/re-implemented?

 One more minor comment:
 I still don't understand the definition of  task-boundary as given in the notebook. Which strategy requires a direct access to the task boundaries? the "Task-Free" setting in the literature is just online task-agnostic. Why would a strategy need the direct access to "task boundaries"? do I understand it correctly?

AntonioCarta:
Thanks @HamedHemati. The errors are due to missing implementations, I'm working on them.

> I saw that the attributes of an experience can have different access modes for each phase: train/eval /logging
I also tried exp.current_experience, and it still works. Do we want to keep the "public" access to the attributes for backward compatibility reasons?

This was an old change, I just added it to the doc. Right now the default is logging mode for backward compatibility, but at some point we will change it.

> AvalancheDataset doesn't seem to have a reset_task_label attribute. Is it possible to wrap a dataset with targets but without task-labels directly? or do should we always first create a classification dataset and then remove the task labels?

I want to distinguish the operation of adding task labels to a task-agnostic scenario and the operation of setting new labels. I don't want the user to accidentally override previously set task labels.
I think I will change `make_classification_dataset` to not add task labels.

> I still don't understand the definition of task-boundary as given in the notebook. Which strategy requires a direct access to the task boundaries? the "Task-Free" setting in the literature is just online task-agnostic. Why would a strategy need the direct access to "task boundaries"? do I understand it correctly?

Example: LwF needs task boundaries to know when to update the teacher model. Many OCL methods have similar operations done at task boundaries. To do these operations, you need explicit boundary labels.

AlbinSou:
I quickly reviewed the changes, here are a few questions:

1) If I understand well, you replaced the ClassificationDataset with TaskAwareClassificationDataset, does it change anything from the user point of view (i.e do we want to change the fact that a class-incremental benchmark returns 0 as task id or does it remain the same ?)

2)  I see you removed benchmark_with_validation_stream, is it replaced by something similar ?

Otherwise I don't have much to say, good job. I think the addition of all these protocols will come in handy. I just find the "taskaware" naming a bit confusing but it's not a big deal.

AntonioCarta:
> If I understand well, you replaced the ClassificationDataset with TaskAwareClassificationDataset, does it change anything from the user point of view (i.e do we want to change the fact that a class-incremental benchmark returns 0 as task id or does it remain the same ?)

Exactly. The old one is kept for backwarrd compatibility (even though it's renamed). The new `ClassificationDataset` will not add task labels by default.

> I see you removed benchmark_with_validation_stream, is it replaced by something similar ?

it's [here](https://github.com/AntonioCarta/avalanche/blob/4300ffaaee209287deedddfedb08df6ac886c9e1/avalanche/benchmarks/scenarios/dataset_scenario.py#L256).


> I just find the "taskaware" naming a bit confusing but it's not a big deal

Do you have any alternative? task-aware just means you have task labels. I think it's the most used nomenclature.

PR #1513, opened 2023-10-09
Some minor fixes in batch_observation and utils

AlbinSou:
Two fixes,

One is some useless code that was left in the batch_observation. This code was slowing down a bit in the case of using an online setting with access to task boundaries. I had mistakenly left this part of code when switching to using util function at_task_boundary.

Another is about some formatting and changing one feature in the at_task_boundary function. So far the function was only using is_first_subexp attribute which was leading to some problems when using it in after_training_task, it was not detecting the boundary if the experience was the last experience of the task. Now it is detecting the boundary both right before the shift and right after the shift.

### DIFF
--- a/avalanche/training/utils.py
+++ b/avalanche/training/utils.py
@@ -15,15 +15,15 @@ General utility functions for pytorch.
 
 """
 from collections import defaultdict
-from typing import Dict, NamedTuple, List, Optional, Tuple, Callable, Union
+from typing import Callable, Dict, List, NamedTuple, Optional, Tuple, Union
 
 import torch
 from torch import Tensor
-from torch.nn import Module, Linear
-from torch.utils.data import Dataset, DataLoader
+from torch.nn import Linear, Module
+from torch.utils.data import DataLoader, Dataset
 
-from avalanche.models.batch_renorm import BatchRenorm2D
 from avalanche.benchmarks import OnlineCLExperience
+from avalanche.models.batch_renorm import BatchRenorm2D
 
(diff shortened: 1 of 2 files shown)


avalanche/training/utils.py
```diff
@@ -45,7 +45,10 @@ def at_task_boundary(training_experience) -> bool:
 
     if isinstance(training_experience, OnlineCLExperience):
         if training_experience.access_task_boundaries:
-            if training_experience.is_first_subexp:
+            if (
+                training_experience.is_first_subexp
+                or training_experience.is_last_subexp
+            ):
```
AntonioCarta, line 51: can you explain this again? I'm a bit confused. With the current condition, we are going to update model and optimizer twice, at the final experience and at the new one.

AlbinSou: Ah, that's right. I had not thought about this. The idea is that this util function should be able to return both sides of the boundary. Maybe I should add an option to say which side to detect ?

For instance, I had the problem recently that to implement LwF for a compatibility with online, I also need to use this function but in the after_training_exp, so I need to know the right side of the boundary, whereas when it is called in before_training_exp I need to know the left side of the boundary.

AntonioCarta: Ok, but make it private (`_at_task_boundary(before=True)`). It's ok as an internal utility but I think it would be a bit confusing for external users

AntonioCarta:
if you fix the lint error I can merge this.

PR #1524, opened 2023-10-19
proper conversion of Tensor to list

malashinroman:
`list(torch.Tensor) ` doesn't convert Tensor to list of scalars, but to list of Tensors.

I checked this is True for pytorch versions 1.12 and 2.0

```
[ins] In [6]: a = torch.Tensor([0,1])
[ins] In [7]: a.tolist()
Out[7]: [0.0, 1.0]
[ins] In [8]: list(a)
Out[8]: [tensor(0.), tensor(1.)]
```
 
That lead to the KeyError when I tried to used combined pytorch datsets, like this

```
    if args.reproduce_bugs:
        # substitute labels 0-9 with 10-19
        fmnist_test.targets += 10
        fmnist_train.targets += 10
        full_test = mnist_test + fmnist_test
        full_train = mnist_train + fmnist_train
        n_classes = 20
    else:
        full_test = mnist_test
        full_train = mnist_train
        n_classes = 10

    benchmark = nc_benchmark(
        full_train,
        full_test,
        n_experiences=n_classes // 2,
        task_labels=list(range(10)),
    )
```

I probably will open an issue, because the same probably will repeat at least for similar usage in 

> avalanche/benchmarks/utils/dataset_traversal_utils.py:191

### DIFF
--- a/avalanche/benchmarks/utils/utils.py
+++ b/avalanche/benchmarks/utils/utils.py
@@ -318,9 +318,11 @@ def _traverse_supported_dataset(
         result = []
         if indices is None:
             for c_dataset in dataset.datasets:
-                result += list(
-                    _traverse_supported_dataset(c_dataset, values_selector, indices)
-                )
+                result_data = _traverse_supported_dataset(c_dataset, values_selector)
+                if isinstance(result_data, Tensor):
+                    result += result_data.tolist()
+                else:
+                    result += list(result_data)
             return result
 
         datasets_to_indexes = defaultdict(list)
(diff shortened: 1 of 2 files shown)


AntonioCarta:
Thanks! Looks good to me. 

can you add a small unit test so that we can avoid a regression?

@lrzpellegrini can you also check this? Do you know if we have other places where we may have the same bug?

lrzpellegrini:
Not that I am aware of. Alas, this is a common issue when dealing with elements that *may* be PyTorch Tensors.

malashinroman:
@AntonioCarta, I will add unittest tomorrow. 

*frankly saying in environment managed as documented I wasn't able to pass couple of fast tests myself

AntonioCarta:
> *frankly saying in environment managed as documented I wasn't able to pass couple of fast tests

If you think this is a general issue and not a problem with your environment, feel free to open a separate issue.

For this PR, you can run just your test with `python -m unittest tests.your.module.TestClass.test_method`

malashinroman:
@AntonioCarta I've found that scenarios I used to found bugs are now deprecated, so I added test to verify type and content of TaskAwareSuperivsedClassificationDataset.targets_val_to_idx

PR #1528, opened 2023-10-31
CIL Examplar Free components

AlbinSou:
I start this PR but it's not finished. Feel free to discuss the things that should be changed. 

For now, here is what I added

- FeCAM (NCM-like but using malahanobis distance (stores per-class covariance matrices) [arxiv](https://arxiv.org/abs/2309.14062)
- FeCAM update utils (current, memory and oracle)
- NCM update utils (current, memory and oracle)
- Cosine Linear first draft

Here is what I want to add:

- Tests for FeCAM
- Tests for all the update utils
- Change Cosine Linear so that it can have new number of new classes (for now it handles only cases with same number of classes in each exp)
- Test for Cosine Linear
- FeatureAdapters [arxiv](https://arxiv.org/abs/2004.00713) (for NCM, for now could not make it work for LinearSVC) + tests
- Resnet18 adapted for Cifar and Tinyimagenet (without maxpool layer) -> @AntonioCarta Do we have this already ?
- Maybe some examples of using all this in continual-learning-baselines
- A lot of regularization based methods use higher lr and different scheduling strategy in the first task and in the remaining ones. I have made a plugin inheriting from LRSchedulerPlugin to handle this kind of things, I could adapt it a bit and propose it (I was thinking of smt general like a plugin that can switch the scheduler based on the task)

### DIFF
--- a/avalanche/models/cosine_layer.py
+++ b/avalanche/models/cosine_layer.py
@@ -13,8 +13,24 @@ Implementation of Cosine layer taken and modified from https://github.com/G-U-N/
 
 
 class CosineLinear(nn.Module):
+    """
+    Cosine layer defined in 
+    "Learning a Unified Classifier Incrementally via Rebalancing" 
+    by Saihui Hou et al.
+
+    Implementation modified from https://github.com/G-U-N/PyCIL
+
+    This layer is aimed at countering the task-recency bias by removing the bias 
+    in the classifier and normalizing the weight and the input feature before 
+    computing the weight-feature product
+    """
     def __init__(self, in_features, out_features, sigma=True):
-        super(CosineLinear, self).__init__()
+        """
+        :param in_features: number of input features
+        :param out_features: number of classes
+        :param sigma: learnable output scaling factor
+        """
+        super().__init__()
         self.in_features = in_features
         self.out_features = out_features
         self.weight = nn.Parameter(torch.Tensor(self.out_features, in_features))
@@ -42,8 +58,9 @@ class CosineLinear(nn.Module):
 
 class SplitCosineLinear(nn.Module):
     """
-    This class keeps two Cosine Linear layers, without sigma, and handles the sigma parameter
-    that is common for the two of them. One CosineLinear is for the old classes and the other
+    This class keeps two Cosine Linear layers, without sigma scaling, 
+    and handles the sigma parameter that is common for the two of them. 
+    One CosineLinear is for the old classes and the other
     one is for the new classes
     """
 
@@ -85,8 +102,10 @@ class CosineIncrementalClassifier(DynamicModule):
             # Do not adapt
             return
         self.num_current_classes = max_class
-        fc = self.generate_fc(self.feature_dim, max_class + 1)
+        fc = self._generate_fc(self.feature_dim, max_class + 1)
         if experience.current_experience == 1:
+            # First exp self.fc is CosineLinear 
+            # while it is SplitCosineLinear for subsequent exps
             fc.fc1.weight.data = self.fc.weight.data
             fc.sigma.data = self.fc.sigma.data
         else:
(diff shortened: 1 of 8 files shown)


avalanche/models/cosine_layer.py
```diff
@@ -0,0 +1,107 @@
...
+
+from avalanche.models import DynamicModule
+
+
+"""
+Implementation of Cosine layer taken and modified from https://github.com/G-U-N/PyCIL
+"""
+
+
+class CosineLinear(nn.Module):
```
AntonioCarta, line 15: doc

avalanche/models/cosine_layer.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class SplitCosineLinear(nn.Module):
+    """
+    This class keeps two Cosine Linear layers, without sigma, and handles the sigma parameter
+    that is common for the two of them. One CosineLinear is for the old classes and the other
+    one is for the new classes
+    """
+
+    def __init__(self, in_features, out_features1, out_features2, sigma=True):
```
AntonioCarta, line 50: doc

avalanche/models/cosine_layer.py
```diff
@@ -0,0 +1,107 @@
...
+
+        return out
+
+
+class SplitCosineLinear(nn.Module):
+    """
+    This class keeps two Cosine Linear layers, without sigma, and handles the sigma parameter
+    that is common for the two of them. One CosineLinear is for the old classes and the other
+    one is for the new classes
+    """
```
AntonioCarta, line 48: what is sigma?

avalanche/models/cosine_layer.py
```diff
@@ -0,0 +1,107 @@
...
+
+        if self.sigma is not None:
+            out = self.sigma * out
+
+        return out
+
+
+class CosineIncrementalClassifier(DynamicModule):
+    # WARNING Maybe does not work with initial evaluation
+    def __init__(self, in_features, num_classes):
```
AntonioCarta, line 76: num_classes should be optional in the incremental version? (1 is a good default)

avalanche/models/cosine_layer.py
```diff
@@ -0,0 +1,107 @@
...
+            fc.fc1.weight.data[:prev_out_features1] = self.fc.fc1.weight.data
+            fc.fc1.weight.data[prev_out_features1:] = self.fc.fc2.weight.data
+            fc.sigma.data = self.fc.sigma.data
+        del self.fc
+        self.fc = fc
+
+    def forward(self, x):
+        return self.fc(x)
+
+    def generate_fc(self, in_dim, out_dim):
```
AntonioCarta, line 103: internal method? use `_generate_fc`. If it's public add doc.

AntonioCarta (changes requested):

AntonioCarta:
Thanks @AlbinSou. Can we split this in multiple PRs? I think FeCAM and CosineLayer are ready once you add tests and doc. We can add the rest later.

I'm not sure about the plugins. The decoupling between the modules and their logic makes it a bit more complex to use them but I don't have a solution for this. If there is no alternative solution, I would at least add a reference to their implementation in the Module's doc.

avalanche/models/cosine_layer.py
AlbinSou: About that, I am not sure right now whether this behavior can be implemented with the current code state. In this code, whenever new classes arrive, the classifier switchs from using one CosineLinear layer to a SplitCosineLinear one (containing two CosineLinear). So, in the case we initialize with 1 class, this SplitCosineLinear will contain the old layer (with 1 logit) plus a new layer containing the new seen classes (let say 9 new ones for the first task). I think this behavior is a bit weird. But I agree that it would be nice to be consistent with the IncrementalClassifier behavior, I will make some more tests to make sure we can get a version that is both correct and consistent with the current IncrementalClassifier

AntonioCarta: If I understand correctly, this means that you should not have a `num_classes` parameter at all. You must always start with 0 units.

PR #1535, opened 2023-11-23
Feature replay and feature distillation (CIL Exfree)

AlbinSou:
Adding a few more methods for CIL exfree, based on feature distillation and feature replay.

Few things,

I don't know what is the canonical citation for feature replay, we should add it if it exists.

I want to include also generative feature replay (model features with multivariate gaussian, with options to do it per class / per task)

I also included Feature distillation, in particular, I did some checks and in my case the performance of Feature distillation + Feature replay is similar (a bit better but by less than 1%) to the performance of LwF (distillation on the logits). So, it leads me to think that I don't know if it's worth to include this approach but at the same time it can be useful in other circumstances (i.e  for distillation losses that need to be applied on the features directly and do not regularize the classifier).

TODO:
- Add Tests
- Add Doc
- Add Generative feature replay
- Add Feature adapters


AlbinSou:
Some update, I added test for both feature replay and feature distillation. However when came the time to test on multitask scenarios, I realised that it was not so obvious to make these strategies work in that setting for now.

One of the blocking point is the wrapper that I use, but also in general all the wrappers, that prevent the use of multitask modules inside, that is why I created issue #1542.

Another blocking point is when I create the feature dataset in feature replay, I don't understand why the created dataset already returns task labels although I always update it with datasets that don't contain task labels. The problem is not to pass the test in that case but rather that the task label is always 0 when in a real scenario it would be a different task label for a different feature. Do you know why this happens @AntonioCarta  ? If I try to put a task_labels attribute in FeatureDataset and return it it then returns too many elements (among which the 0 task label).

AntonioCarta:
Hi @AlbinSou, the code looks good.

> I don't know what is the canonical citation for feature replay, we should add it if it exists.

I don't know, maybe @lrzpellegrini has a good suggestion.

> If I try to put a task_labels attribute in FeatureDataset and return it it then returns too many elements (among which the 0 task label).

The DataAttribute for task labels is unfortunately called `targets_task_labels`. We should probably rename it but I never did it because I don't know how many things can subtly break with a rename. I think you are adding task labels somewhere (defaulted to zero), and your attribute is an additional one.

AlbinSou:
One test is failing on the server but not locally. It seems like it's the test_icarl, do you know if they are some known problems with it ?

AntonioCarta:
I'm not sure what's happening. The CI on the master branch seems to be working. Maybe it's an environment issue?

PR #1550, opened 2023-12-19
Major BiC fix/reimplementation

lrzpellegrini:
This PR is a major re-implementation/global fix for [BiC (paper)](https://openaccess.thecvf.com/content_CVPR_2019/papers/Wu_Large_Scale_Incremental_Learning_CVPR_2019_paper.pdf)

The implementation found in Avalanche (and in other CL libraries as well) has various issues, mainly connected to the understanding of how the algorithm works. It's an honest mistake that I made too when reading the paper.

The official implementation of BiC can be found here (TF 1.x :/ ): https://github.com/wuyuebupt/LargeScaleIncrementalLearning/tree/master

BiC works like that:
1. Distillation phase: classic training loop with distillation loss (on logits of old classes) + classification loss (logits of old and new classes)
2. Bias correction phase: learn alpha and beta correction parameters for new classes

The main issue in Avalanche implementation is that BiC does not actually keep a separate bias correction layer for each experience, it only keeps single alpha and beta parameters (a single bias correction layer) which is then applied only to the "last" experience.

In practice:

1. Distillation phase: distill on logits (after the FC layer) by applying the bias correction on the activations of the old model (used for distillation) for the classes found in experience `current_experience-1`. Bias correction is not applied to other classes (such as the ones in `current_experience-t` t>=2, nor `current_experience`. Bias correction is not applied to the activations from the current model. The usual classification loss is applied on the logits of the current model without bias correction.
2. Discard the previous bias correction layer and learn new alpha and beta values for the classes in `current_experience`
3. (eval) During the test phase, apply the bias correction only to the logits of classes from `current_experience`

What Avalanche was doing (and I suspect FACIL too) is to keep a separate (permanent) bias correction layer for each experience. Those were then used in all phases (computing the distillation loss, classification loss, bias correction for new classes, test phase).

I noticed these problems when fixing other accidental bugs related to the freezing of the old model and bias layers, which are of course fixed in this version :).

To recap:

- Fix the distillation loss computation procedure (proper freezing of the old model, bias correction only on classes of `current_experience-1`, ...)
- Fix the loss function (see: https://colab.research.google.com/drive/1eLQbfxe6i0V7tu5spMxLSopebuP7Wc9l?usp=sharing)
- Fix the bias correction procedure (freezing of the current model, post-training bias layer freezing, ...)
- Fix the BiasLayer implementation (the forward function was bugged as the beta-derived tensor was initialized to 1s instead of 0s)

### DIFF
--- a/tests/benchmarks/scenarios/test_task_aware.py
+++ b/tests/benchmarks/scenarios/test_task_aware.py
@@ -16,7 +16,8 @@ class TestsTaskAware(unittest.TestCase):
     def test_taskaware(self):
         """Common use case: add tas labels to class-incremental benchmark."""
         n_classes, n_samples_per_class, n_features = 10, 3, 7
-        while True:
+
+        for _ in range(10000):
             dataset = make_classification(
                 n_samples=n_classes * n_samples_per_class,
                 n_classes=n_classes,
@@ -25,6 +26,9 @@ class TestsTaskAware(unittest.TestCase):
                 n_redundant=0,
             )
 
+            # The following check is required to ensure that at least 2 exemplars
+            # per class are generated. Otherwise, the train_test_split function will
+            # fail.
             _, unique_count = np.unique(dataset[1], return_counts=True)
             if np.min(unique_count) > 1:
                 break
--- a/tests/benchmarks/scenarios/test_task_aware.py
+++ b/tests/benchmarks/scenarios/test_task_aware.py
@@ -9,19 +9,26 @@ from avalanche.benchmarks.scenarios.task_aware import task_incremental_benchmark
 from avalanche.benchmarks.utils.classification_dataset import ClassificationDataset
 from avalanche.benchmarks.utils.data_attribute import DataAttribute
 from torch.utils.data import TensorDataset
+import numpy as np
 
 
 class TestsTaskAware(unittest.TestCase):
     def test_taskaware(self):
         """Common use case: add tas labels to class-incremental benchmark."""
         n_classes, n_samples_per_class, n_features = 10, 3, 7
-        dataset = make_classification(
-            n_samples=n_classes * n_samples_per_class,
-            n_classes=n_classes,
-            n_features=n_features,
-            n_informative=6,
-            n_redundant=0,
-        )
+        while True:
+            dataset = make_classification(
+                n_samples=n_classes * n_samples_per_class,
+                n_classes=n_classes,
+                n_features=n_features,
+                n_informative=6,
+                n_redundant=0,
+            )
+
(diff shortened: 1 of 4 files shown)


AntonioCarta:
Thanks, the code looks ok. I see that we are missing a BiC reproducibility example in continual-learning-baselines. If you have time maybe you can add something there too.

lrzpellegrini:
I'm trying to reproduce a CIFAR experiment. Alas, I also found a bug in the `_ParametricSingleBuffer` (to apply the iCarl herding). I'll try to include a fix for that, too.

lrzpellegrini:
The results are far better than the ones obtained by using the previous version.

Once tests are finished, I recommend merging.

lrzpellegrini:
@AntonioCarta In the code I used `_concat_taskaware_classification_datasets`. Is there a better way to do this without using a private function?

AntonioCarta:
> @AntonioCarta In the code I used `_concat_taskaware_classification_datasets`. Is there a better way to do this without using a private function?

The `Avalanche.concat` should work without issues. `_concat_taskaware_classification_datasets` is the old method, kept for compatibility because it also handles targets, transform, and collate.

lrzpellegrini:
I switched to `concat_datasets`. Let me know if it makes sense to you.

Apart from the Python 3.11 tests (which, if I'm correct, are still being fixed), everything works fine.

tests/benchmarks/scenarios/test_task_aware.py
```diff
@@ -9,19 +9,26 @@
...
+            dataset = make_classification(
+                n_samples=n_classes * n_samples_per_class,
+                n_classes=n_classes,
+                n_features=n_features,
+                n_informative=6,
+                n_redundant=0,
+            )
+
+            _, unique_count = np.unique(dataset[1], return_counts=True)
+            if np.min(unique_count) > 1:
```
AntonioCarta, line 29: What's going on here? How is this condition not always true? I'm afraid this issue may break other tests too.

AntonioCarta (approved):

AntonioCarta:
Everything is ok. Just a minor comment about the test change.

tests/benchmarks/scenarios/test_task_aware.py
lrzpellegrini: It's a corner case in which only 1 exemplar is created for one of the classes. In that case, the train test split function fails.

AntonioCarta: ok, then it's fine. Can you add a comment about it? I hope this is not causing issues somewhere else...

PR #1561, opened 2024-01-23
WIP: Remove conda files

AntonioCarta:


.github/workflows/environment-update.yml
lrzpellegrini, line 1: I think it could make sense to add requirements.txt and setup-related files (such as setup.py) to the paths that trigger this action. We currently have `'.github/workflows/environment-update.yml'` only

docs/gitbook/getting-started/how-to-install.md
lrzpellegrini, line 1: We can add that, by installing the [all] and [extra] versions, the PyTorch version may be limited to <2.* due to the dependencies of those additional packages.

lrzpellegrini (approved):
Removing conda env files makes sense as it's currently becoming harder to support an environment with both conda and pip dependencies.

I left a couple of comments to check.

PS: I think we can safely remove the constraint on typing_extensions in requirements.txt, too.

PS 2: after further investigation, it seems that versions of typing-extensions `<4.6.0` do not work well with dill. It seems reasonable to add a `>=4.6.0` constraint.

.github/workflows/docker-nightly-release.yml
```diff
@@ -21,6 +21,9 @@ on:
     paths:
       - 'docker/nightly/**'
       - '.github/workflows/docker-nightly-release.yml'
+      - 'requirements.txt'
+      - 'setup.py'
+      - 'extra_dependencies.py'
```
lrzpellegrini, line 26: I think it's extra_dependencies.txt

.github/workflows/environment-update.yml
```diff
@@ -17,6 +17,9 @@ on:
       - master
     paths:
       - '.github/workflows/environment-update.yml'
+      - 'requirements.txt'
+      - 'setup.py'
+      - 'extra_dependencies.py'
```
lrzpellegrini, line 22: I think it's extra_dependencies.txt

.github/workflows/unit-test.yml
```diff
@@ -26,6 +26,8 @@ on:
     paths:
       - '**.py'
       - '.github/workflows/unit-test.yml'
+      - 'requirements.txt'
+      - 'extra_dependencies.py'
```
lrzpellegrini, line 30: I think it's extra_dependencies.txt

lrzpellegrini (commented):
I think that the correct file name is `extra_dependencies.txt'

PR #1587, opened 2024-01-29
Fix multiple inheritance in Python 3.11

lrzpellegrini:
This fixes the multiple inheritance issue in Python 3.11.

Strategy constructors now accept all arguments as keyword-only. The main elements of the class hierarchy now use a cooperative constructor mechanism. The deprecation of the previous positional-based mechanism is handled to allow users to use existing code (Avalanche will print a warning if positional arguments declared/used).

This also has an additional change to the coveralls action to align it to the unit-test one.


avalanche/training/supervised/er_ace.py
```diff
@@ -47,6 +49,7 @@ def __init__(
         ] = default_evaluator,
         eval_every=-1,
         peval_mode="epoch",
+        **kwargs
```
AntonioCarta, line 52: Why is **kwargs needed also for top-level classes? There shouldn't be any extra arguments. I'm worried because unused kwargs would go unnoticed with this solution.
My understanding is that we need it for classes that are not top-level so that the constructor works for any MRO.

avalanche/training/templates/observation_type/batch_observation.py
```diff
@@ -12,8 +6,8 @@
 
 
 class BatchObservation(SGDStrategyProtocol):
-    def __init__(self):
-        super().__init__()
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
AntonioCarta, line 10: if we are using only kwargs shouldn't we have only kwargs in the method signature and super call?
btw, maybe add a small comment to explain this in the template classes.

avalanche/training/supervised/er_ace.py
lrzpellegrini: Yes, makes sense. But consider that it's still a WIP and I still need to work on this PR!

tests/models/test_packnet.py
```diff
@@ -26,16 +29,17 @@ def _test_PackNetPlugin(self, expectations, optimizer_constructor, lr):
...
-            model,
-            prune_proportion=0.5,
-            post_prune_epochs=1,
-            optimizer=optimizer,
-            train_epochs=2,
-            train_mb_size=10,
-            eval_mb_size=10,
-            device="cuda" if use_gpu else "cpu",
-        )
+        with self.assertWarns(PositionalArgumentDeprecatedWarning):
```
AntonioCarta, line 32: move test?

PR #1596, opened 2024-02-13
Generative Replay with weighted loss for replayed data

gogamid:
First attempt to add support for Generative Replay with weighted loss for replayed data. 
Added three arguments:
```python
is_weighted_replay=True,
weight_replay_loss_factor=2.0,
weight_replay_loss=0.001,
```


avalanche/training/plugins/generative_replay.py
```diff
@@ -52,11 +52,14 @@ class GenerativeReplayPlugin(SupervisedPlugin):
...
+            generator_strategy=None,
+            untrained_solver: bool = True,
+            replay_size: Optional[int] = None,
+            increasing_replay_size: bool = False,
+            is_weighted_replay: bool = False,
+            weight_replay_loss_factor: float = 1.0,
+            weight_replay_loss: float = 0.0001,
     ):
         """
         Init.
```
AntonioCarta, line 65: can you add the documentation for the arguments?

avalanche/training/plugins/generative_replay.py
```diff
@@ -101,19 +107,67 @@ def before_training_exp(
...
         """
         Set untrained_solver boolean to False after (the first) experience,
         in order to start training with replay data from the second experience.
         """
         self.untrained_solver = False
 
+    def before_backward(self, strategy: Template, *args, **kwargs) -> Any:
+        super().before_backward(strategy, *args, **kwargs)
+        """
+        Generate replay data and calculate the loss on the replay data.
```
AntonioCarta, line 121: docstring must be before super() call

avalanche/training/plugins/generative_replay.py
```diff
@@ -101,19 +107,67 @@ def before_training_exp(
...
+        # get labels for replay data
+        if not self.model_is_generator:
+            with torch.no_grad():
+                replay_output = self.old_model(replay_data).argmax(dim=-1)
+        else:
+            # Mock labels:
+            replay_output = torch.zeros(replay_data.shape[0])
+
+        # make copy of mbatch
+        mbatch = deepcopy(strategy.mbatch)
```
AntonioCarta, line 154: do you need a deepcopy here?

examples/generative_replay_splitMNIST_weighted.py
```diff
@@ -0,0 +1,97 @@
+################################################################################
```
AntonioCarta, line 1: please modify one of the existing examples instead of creating a new one

AntonioCarta:
Thank you for your contribution. I left some comments.

It's not clear to me why you multiply `weight_replay_loss_factor` after each iteration. Is this really what you want?

FYI, you can fix the lint issues with by running `black .`.

gogamid:
Thank you for reviewing, I am trying to make it better.

> It's not clear to me why you multiply `weight_replay_loss_factor` after each iteration. Is this really what you want?


According to this section of [API Docu](https://avalanche-api.continualai.org/en/v0.5.0/generated/avalanche.training.plugins.GenerativeReplayPlugin.html#avalanche.training.plugins.GenerativeReplayPlugin):
"Another way to implempent the algorithm is by weighting the loss function and give more importance to the replayed data as the number of experiences increases."
Therefore, I increase the weight of loss by multiplying it with a factor to give more importance to the replay data as the number of experiences increases. Is there better suggestion that you can make?

AndreaCossu:
Thanks @gogamid , this looks good!

PR #1598, opened 2024-02-14
fix issue #1597

### LINKED ISSUE #1597: benchmark_with_validation stream removes or hides classes_in_experience attribute
🐛 **Describe the bug**

When using benchmark_with_validation_stream, the classes_in_experience attribute is removed or hidden and cannot be caught during model adaptation.

🐜 **To Reproduce**

```{python}
import torch
import argparse
from torch.nn import CrossEntropyLoss
from torch.optim import SGD

from avalanche.benchmarks.classic import SplitCIFAR100
from avalanche.benchmarks import benchmark_with_validation_stream
from avalanche.models import SlimResNet18
from avalanche.models.dynamic_modules import IncrementalClassifier
from avalanche.training.supervised import Naive

def main(args):
    # Device config
    device = torch.device(
        f"cuda:{args.cuda}" if torch.cuda.is_available() and args.cuda >= 0 else "cpu"
    )

    # model
    model = SlimResNet18(1)
    model.linear = IncrementalClassifier(model.linear.in_features, model.linear.out_features)

    benchmark = SplitCIFAR100(n_experiences=5, seed=1, dataset_root="/DATA/data")
    benchmark = benchmark_with_validation_stream(benchmark, validation_size=0.05)

    # Than we can extract the parallel train and test streams
    train_stream = benchmark.train_stream
    test_stream = benchmark.test_stream

    # Prepare for training & testing
    optimizer = SGD(model.parameters(), lr=0.001, momentum=0.9)
    criterion = CrossEntropyLoss()

    # Continual learning strategy with default logger
    cl_strategy = Naive(
        model,
        optimizer,
        criterion,
        train_mb_size=32,
        train_epochs=0,
        eval_mb_size=32,
        device=device,
    )

    # train and test loop
    results = []
    for train_task in train_stream:
        print("Current Classes: ", train_task.classes_in_this_experience)
        cl_strategy.train(train_task)
        results.append(cl_strategy.eval(test_stream))


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--cuda",
        type=int,
        default=0,
        help="Select zero-indexed cuda device. -1 to use CPU.",
    )
    args = parser.parse_args()

    main(args)
```

🐝 **Expected behavior**

Should work with and without calling benchmark_with_validation_stream, but here gives an error

```
    print("Current Classes: ", train_task.classes_in_this_experience)
  File "/home/albin/Documents/Projects/avalanche_dev/avalanche/benchmarks/scenarios/generic_scenario.py", line 170, in __getattribute__
    v = super().__getattribute__(item)
AttributeError: 'DatasetExperience' object has no attribute 'classes_in_this_experience'
```

AntonioCarta:
@AlbinSou this should fix your issue.

Closes #1597

### DIFF
--- a/avalanche/benchmarks/scenarios/__init__.py
+++ b/avalanche/benchmarks/scenarios/__init__.py
@@ -8,3 +8,4 @@ from .task_aware import *
 from .dataset_scenario import *
 from .exmodel_scenario import *
 from .online import *
+from .validation_scenario import *
--- a/avalanche/benchmarks/scenarios/dataset_scenario.py
+++ b/avalanche/benchmarks/scenarios/dataset_scenario.py
@@ -14,7 +14,6 @@
 import random
 from avalanche.benchmarks.utils.data import AvalancheDataset
 import torch
-from itertools import tee
 from typing import (
     Callable,
     Generator,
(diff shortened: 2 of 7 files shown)


AntonioCarta:
@lrzpellegrini I think the distributed tests are failing.

lrzpellegrini:
Yes, classic distributed issues. The problem is not on the code itself, but how torch shuts down the distributed module when exiting the process. I'll add a final sync + a different way to detect if a test failed.

Issue #1591, opened 2024-02-10
IncrementalClassifiers inside MultiHeadClassifiers are adapted on all experiences


AlbinSou:
🐛 **Describe the bug**
MultiHead classifier contains a dictionnary of Incremental classifiers, both MultiHeadClassifier and IncrementalClassifier are DynamicModules, so they get adapted through the following loop:

```{python}
def avalanche_model_adaptation(model: nn.Module, experience: CLExperience):
    if isinstance(model, DistributedDataParallel):
        raise RuntimeError(
            "The model is wrapped in DistributedDataParallel. "
            "Please unwrap it before calling this method."
        )
    for module in model.modules():
        if isinstance(module, DynamicModule):
            module.adaptation(experience)
```

However, we would like the IncrementalClassifiers inside the MultiHeadClassifier to be adapted only on experiences that contain the task it corresponds to, just like it is called inside the MultiHeadClassifier 

One solution would be to add an option "adaptable" in Incremental classifiers and turn this off when they are created in MultiHeadClassifiers.

AntonioCarta:
This is an issue indeed. I agree with your solution of disabling adaptation.

Another issue that we have is that the model adaptation must be called on each module separately. This is a frequent source of errors that I see in the github discussions (calling `model.adaptation` instead of `model_adaptation` or having a for loop). Maybe the `adaptation` method should automatically call its children? I'm thinking of an API like this:

```
class DynamicModule:
    def adaptation(self, exp):  # called by users and avalanche strategies. Adapts children too
    def _module_adaptation(self, exp):  # adapt itself
    def enable_adaptation(self, is_enabled):  # enable/disable adaptation  
```

and maybe IncrementalClassifier could have a method `grow_classifier(new_units)` that can be called by the `MultiTaskClassifier` that can be called avoiding the adaptation?

AlbinSou:
> This is an issue indeed. I agree with your solution of disabling adaptation.
> 
> Another issue that we have is that the model adaptation must be called on each module separately. This is a frequent source of errors that I see in the github discussions (calling `model.adaptation` instead of `model_adaptation` or having a for loop). Maybe the `adaptation` method should automatically call its children? I'm thinking of an API like this:
> 
> ```
> class DynamicModule:
>     def adaptation(self, exp):  # called by users and avalanche strategies. Adapts children too
>     def _module_adaptation(self, exp):  # adapt itself
[...]

PR #1600, opened 2024-02-16
Changes in Dynamic Modules

AlbinSou:
I tried to implement the changes that would fix bugs described in Issue #1591 and also allow users to use Dynamic modules more easily outside of avalanche strategies.

To do so, I made several major changes.

- Remove old avalanche_model_adaptation method and replace it with a new version which does things differently. Instead of iterating on all modules, it iterates only on childrens, and does so recursively in order to explore the whole hierarchy. It still iterates on the hierarchy of nn.Modules and catches DynamicModules to adapt them just like before.
- Added a method adapt() to dynamicmodule that will call the new "avalanche_model_adaptation" on itself and recursively on it's submodules
- Added an _auto_adapt attribute to all DynamicModules that triggers or not the adaptation inside the avalanche_model_adaptation recursion (defaults to True). In particular, this attribute does not prevent the user to call directly module.adapt() if he wants, however if the module needs to be adapted as a second level or more module inside this recursion, it will not get adapted. This is the part that fixes the bug described in #1591, since that way we can instantiate IncrementalClassifiers inside the MultiHeadClassifier that will not be automatically adapted but rather adapted manually by the MultiHeadClassifier only if the experience used for adaptation contains the task id related to this IncrementalClassifier.
- Made a few clarification changes in IncrementalClassifier and Multiheadclassifier by replacing occurences of adaptation() implementation by train_adaptation and eval_adaptation, where if the eval_adaptation needs to be the same than train_adaptation I just call train_adaptation inside eval_adaptation (nested adaptation() override were a bit complex to understand and could lead to confusion)

For now tests do not pass, but some of them I am not sure what behaviour they were checking and if we really want this behaviour. I will also add some more tests to check that the behaviour in #1591 is solved (my local test says it is fixed for now).

### DIFF
--- a/avalanche/models/dynamic_modules.py
+++ b/avalanche/models/dynamic_modules.py
@@ -173,11 +173,9 @@ class MultiTaskModule(DynamicModule):
         self.max_class_label = max(self.max_class_label, max(curr_classes) + 1)
         super().adaptation(experience)
 
-    def eval_adaptation(self, experience: CLExperience):
-        pass
-
     def train_adaptation(self, experience: CLExperience):
         """Update known task labels."""
+        super().train_adaptation(experience)
         task_labels = experience.task_labels
         self.known_train_tasks_labels = self.known_train_tasks_labels.union(
             set(task_labels)
@@ -277,6 +275,7 @@ class IncrementalClassifier(DynamicModule):
         :param experience: data from the current experience.
         :return:
         """
+        super().train_adaptation(experience)
         device = self._adaptation_device
         in_features = self.classifier.in_features
         old_nclasses = self.classifier.out_features
@@ -304,6 +303,7 @@ class IncrementalClassifier(DynamicModule):
         self.classifier.bias[:old_nclasses] = old_b
 
     def eval_adaptation(self, experience):
+        super().eval_adaptation(experience)
         self.train_adaptation(experience)
 
     def forward(self, x, **kwargs):
@@ -404,6 +404,7 @@ class MultiHeadClassifier(MultiTaskModule):
         :param experience: data from the current experience.
         :return:
         """
+        super().train_adaptation(experience)
         device = self._adaptation_device
         curr_classes = experience.classes_in_this_experience
         task_labels = experience.task_labels
@@ -460,6 +461,7 @@ class MultiHeadClassifier(MultiTaskModule):
                     self._buffers[au_name][curr_classes] = 1
 
     def eval_adaptation(self, experience):
+        super().eval_adaptation(experience)
         self.train_adaptation(experience)
 
     def forward_single_task(self, x, task_label):
--- a/avalanche/models/dynamic_modules.py
+++ b/avalanche/models/dynamic_modules.py
@@ -12,12 +12,39 @@
 to allow architectural modifications (multi-head classifiers, progressive
 networks, ...).
 """
+from typing import List, Optional
+
 import torch
 from torch.nn import Module
-from typing import Optional
 
-from avalanche.benchmarks.utils.flat_data import ConstantSequence
 from avalanche.benchmarks.scenarios import CLExperience
+from avalanche.benchmarks.utils.flat_data import ConstantSequence
+
+
+def adapt_recursive(
+    module: Module,
+    experience: CLExperience,
+    _visited: List = None,
+    _initial_call: bool = True,
+):
+    if _visited is None:
+        _visited = []
+
+    if module in _visited:
+        return
+
+    _visited.append(module)
+
+    if isinstance(module, DynamicModule):
+        if (not _initial_call) and (not module._auto_adapt):
+            # Some modules don't want to be auto-adapted
+            return
+        else:
+            module.adaptation(experience)
+
+    # Iterate over children
+    for name, submodule in module.named_children():
+        adapt_recursive(submodule, experience, _visited=_visited, _initial_call=False)
 
 
 class DynamicModule(Module):
@@ -28,6 +55,22 @@ class DynamicModule(Module):
     Compared to pytoch Modules, they provide an additional method,
     `model_adaptation`, which adapts the model given the current experience.
     """
+
+    def __init__(self, auto_adapt=True):
+        """
+        :param auto_adapt: If True, will be adapted in the recursive adaptation loop
+                           else, will be adapted by a module in charge
+                           (i.e IncrementalClassifier inside MultiHeadClassifier)
+        """
+        super().__init__()
+        self._auto_adapt = auto_adapt
+
+    def adapt(self, experience):
+        """
+        Calls self.adaptation recursively accross
+        the hierarchy of module children
+        """
+        adapt_recursive(self, experience)
 
     def adaptation(self, experience: CLExperience):
         """Adapt the module (freeze units, add units...) using the current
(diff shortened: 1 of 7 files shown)


tests/models/test_models.py
```diff
@@ -455,13 +456,23 @@ def test_multihead_head_selection(self):
         for x, y, t in DataLoader(benchmark.train_stream[0].dataset):
             y_mh = model(x, t)
             y_t = model_t0(x)
+
+            # We need to pad y_t to dim with zeros
+            # because y_mh will have max dim of all heads
+            y_t = F.pad(y_t, (0, y_mh.size(1) - y_t.size(1)))
+
```
AlbinSou, line 463: Previous test was assuming that y_t and y_mh are always of the same size. This is not true. The reason why is that when passing a vector as a target to MultiHeadClassifier, it will always return an output that is padded to the maximum size of one of it's head, this is because it assumes that more than one unique target can be inside the vector (unfortunately, most of the time we indeed only pass one task inside the vector, maybe this is a behaviour we want to change). However, because of the bug the output y_t was also of the size of the biggest task because all IncrementalClassifiers were updated on all experiments. Now it's not the case anymore so before checking equality I have to pad y_t with zeros.

avalanche/models/dynamic_modules.py
```diff
@@ -12,12 +12,41 @@
...
+    _visited=None,
+    _initial_call: bool = True,
+):
+    if _visited is None:
+        _visited = []
+
+    if module in _visited:
+        return
+
+    _visited.append(module)
```
AntonioCarta, line 36: it is better if `_visited` is a set for efficiency in the search.

avalanche/models/dynamic_modules.py
```diff
@@ -12,12 +12,41 @@
...
-from avalanche.benchmarks.utils.flat_data import ConstantSequence
 from avalanche.benchmarks.scenarios import CLExperience
+from avalanche.benchmarks.utils.flat_data import ConstantSequence
+
+
+def avalanche_model_adaptation(
+    module: Module,
+    experience: CLExperience,
+    _visited=None,
+    _initial_call: bool = True,
```
AntonioCarta, line 28: can you add a comment about why we need `_initial_call`?

tests/models/test_models.py
```diff
@@ -416,8 +417,8 @@ def test_multihead_head_creation(self):
             w.data_ptr() for group in optimizer.param_groups for w in group["params"]
         ]
 
-        assert w_ptr not in opt_params_ptrs  # head0 has been updated
-        assert b_ptr not in opt_params_ptrs  # head0 has been updated
+        # assert w_ptr not in opt_params_ptrs  # head0 has NOT been updated
+        # assert b_ptr not in opt_params_ptrs  # head0 has NOT been updated
```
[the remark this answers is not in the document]
AntonioCarta: Are you sure? I think it's if head0 is changed by SGD when training on task 4, which should not happen.

tests/models/test_models.py
AntonioCarta: Correct.

AntonioCarta:
> Made a few clarification changes in IncrementalClassifier and Multiheadclassifier by replacing occurences of adaptation() implementation by train_adaptation and eval_adaptation, where if the eval_adaptation needs to be the same than train_adaptation I just call train_adaptation inside eval_adaptation (nested adaptation() override were a bit complex to understand and could lead to confusion)

I think we could also remove `train_adaptation` and `eval_adaptation` and have a single method. After, you can detect train/eval mode with a single if when necessary, so it seems to be adding unnecessary complexity in the common case.

tests/models/test_models.py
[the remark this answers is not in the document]
AlbinSou: w_ptr contains the head0 data pointer after head0 has been adapted on task 0. This pointer should still be the one present in optimizer after adaptation is done on task 4, since it should not be touched by the task 4 adaptation (which modifies head 4 only, and not head 0).

AlbinSou:
@AntonioCarta  How would you like to manage the adapt() and adaptation() function ? I wanted to keep the name of the "adaptation" also to keep backward compatibility but now it's a bit confusing to have two methods adapt() and adaptation()

AntonioCarta:
> @AntonioCarta How would you like to manage the adapt() and adaptation() function ? I wanted to keep the name of the "adaptation" also to keep backward compatibility but now it's a bit confusing to have two methods adapt() and adaptation()

The advantage of the current solution is that it's backward compatible. If we don't want to rename the old method we could rename the new one as `rec_adapt` for recursive adaptation?

I was also thinking about a test case to check that the recursive calls don't break the #1591 fix:
```
model = ABasicDynamicModule(ABasicModule(MultiTaskClassifier(...)))
model.adapt(exp)  # here it's calling rec_adapt from the "base" layer, not the MultiTaskClassifier. It should still work.
# assert that wrong heads are not adapted
```

AlbinSou:
> > @AntonioCarta How would you like to manage the adapt() and adaptation() function ? I wanted to keep the name of the "adaptation" also to keep backward compatibility but now it's a bit confusing to have two methods adapt() and adaptation()
> 
> The advantage of the current solution is that it's backward compatible. If we don't want to rename the old method we could rename the new one as `rec_adapt` for recursive adaptation?
> 
> I was also thinking about a test case to check that the recursive calls don't break the #1591 fix:
> 
> ```
> model = ABasicDynamicModule(ABasicModule(MultiTaskClassifier(...)))
> model.adapt(exp)  # here it's calling rec_adapt from the "base" layer, not the MultiTaskClassifier. It should still work.
> # assert that wrong heads are not adapted
> ```

I added a test for that at the end of test_models.DynamicModulesTest. Yes, maybe I will rename to make it clearer

tests/models/test_models.py
```diff
@@ -511,6 +522,42 @@ def test_incremental_classifier_update_masking_only_during_training(self):
...
+        for t, exp in enumerate(benchmark.train_stream):
+            avalanche_model_adaptation(model, exp)
+
+        for t, s in sizes.items():
+            self.assertEqual(s, model.classifiers[str(t)].classifier.out_features)
+
+    def test_avalanche_adaptation(self):
+        model1 = torch.nn.Sequential(MultiHeadClassifier(in_features=6))
+        benchmark = get_fast_benchmark(use_task_labels=True, shuffle=True)
+        avalanche_model_adaptation(model1, benchmark.train_stream[0])
```
AntonioCarta, line 543: the test should check that only the correct modules are updated in this test case

AlbinSou: I was doing it in a test above, I changed them a bit now I have two tests, one for avalanche_model_adaptation() and one for recursive_adaptation, both test that the correct modules were adapted (through checking that the sizes of the heads do not change when doing 2 adaptations). I removed the first test_multihead_size because it was doing the same.

AlbinSou:
@AntonioCarta On my side everything is ok. I was thinking maybe to include an example on how to use dynamic modules outside and inside of a strategy (call to recursive_adaptation and explanation on how avalanche_model_adaptation works). Another thing is that I changed the location of avalanche_model_adaptation for now it's inside dynamic_modules.py before it was in models/utils.py. For now I import it in utils so that ppl don't have to change their imports, but it's a bit weird. The reason I put it inside dynamic_modules is that this function is needed in recursive_adaptation() method so putting it to utils was leading to annoying recursive imports.

AntonioCarta:
Ok. I will wait to merge this after release to be on the safe side.

> I was thinking maybe to include an example on how to use dynamic modules outside and inside of a strategy 

we already have examples of that in the notebooks. Maybe you can just update them?

PR #1601, opened 2024-02-20
LaMAMLv2 buffer now uses subset instead of indexing

AndreaCossu:
Fix a bug in lamaml v2 buffer usage


AntonioCarta:
Looks good to me. @HamedHemati can you also give it a quick look?

HamedHemati (commented):
Also looks good to me!

PR #1629, opened 2024-03-21
WIP: Documentation & Typos

jneuendorf:
Still PR is still in progress and not "clean for review" - just some things, I stumbled across. I opened it already, so you know there is some progress on certain aspects that others don't have to worry about 😉

### DIFF
--- a/avalanche/benchmarks/datasets/dataset_utils.py
+++ b/avalanche/benchmarks/datasets/dataset_utils.py
@@ -48,7 +48,7 @@ def load_config_file():
 
 
 def maybe_init_config_file():
-    """Initialize Avalanche user's config file, if it does not exists yet.
+    """Initialize Avalanche user's config file, if it does not exist yet.
 
     The file is located in `~/.avalanche/config.json`
     """
--- a/avalanche/benchmarks/utils/data.py
+++ b/avalanche/benchmarks/utils/data.py
@@ -50,7 +50,7 @@ TDataWTransform = TypeVar("TDataWTransform", bound="_FlatDataWithTransform")
 class AvalancheDataset(IDataset[T_co]):
     """Avalanche Dataset.
 
-    Avlanche dataset are pytorch-compatible Datasets with some additional
+    Avalanche dataset are pytorch-compatible Datasets with some additional
     functionality such as:
     - management of transformation groups via :class:`AvalancheTransform`
     - support for sample attributes such as class targets and task labels
@@ -79,9 +79,9 @@ class AvalancheDataset(IDataset[T_co]):
 
     switching to a different transformation group by calling the ``train()``,
     ``eval()`` or ``with_transforms` methods always returns a new dataset,
-    levaing the original one unchanged.
+    leaving the original one unchanged.
 
-    Ttransformation groups can be manipulated by removing, freezing, or
+    Transformation groups can be manipulated by removing, freezing, or
     replacing transformations. Each operation returns a new dataset, leaving
     the original one unchanged.
     """
@@ -103,9 +103,7 @@ class AvalancheDataset(IDataset[T_co]):
             applied by this dataset.
         :param transform_groups: Avalanche transform groups.
         """
-        if issubclass(type(datasets), TorchDataset) or issubclass(
-            type(datasets), AvalancheDataset
-        ):
+        if isinstance(datasets, (TorchDataset, AvalancheDataset)):
             datasets = [datasets]  # type: ignore
 
         # NOTES on implementation:
--- a/avalanche/benchmarks/utils/flat_data.py
+++ b/avalanche/benchmarks/utils/flat_data.py
@@ -186,10 +186,10 @@ class FlatData(IDataset[T_co], Sequence[T_co]):
     """FlatData is a dataset optimized for efficient repeated concatenation
     and subset operations.
 
-    The class combines concatentation and subsampling operations in a single
+    The class combines concatenation and subsampling operations in a single
     class.
 
-    Class for internal use only. Users shuold use `AvalancheDataset` for data
+    Class for internal use only. Users should use `AvalancheDataset` for data
     or `DataAttribute` for attributes such as class and task labels.
 
     *Notes for subclassing*
--- a/docs/gitbook/from-zero-to-hero-tutorial/03_benchmarks.md
+++ b/docs/gitbook/from-zero-to-hero-tutorial/03_benchmarks.md
@@ -43,10 +43,10 @@ from avalanche.benchmarks.utils import as_classification_dataset, AvalancheDatas
 
 # Most datasets in Avalanche are automatically downloaded the first time you use them
 # and stored in a default location. You can change this folder by calling
-# avalanche.benchmarks.utils.set_dataset_root(new_location)
+# avalanche.benchmarks.datasets.dataset_utils.set_dataset_root(new_location)
 datadir = default_dataset_location('mnist')
 
-# As we would simply do with any Pytorch dataset we can create the train and 
+# As we would simply do with any Pytorch dataset we can create the train and
 # test sets from it. We could use any of the above imported Datasets, but let's
 # just try to use the standard MNIST.
 train_MNIST = MNIST(datadir, train=True, download=True)
@@ -65,14 +65,14 @@ eval_transforms = torchvision.transforms.Compose([
 train_MNIST = as_classification_dataset(
     train_MNIST,
     transform_groups={
-        'train': train_transforms, 
+        'train': train_transforms,
         'eval': eval_transforms
     }
 )
 test_MNIST = as_classification_dataset(
     test_MNIST,
     transform_groups={
-        'train': train_transforms, 
+        'train': train_transforms,
         'eval': eval_transforms
     }
 )
@@ -117,7 +117,7 @@ print(list(dsub.targets))
 ## 🏛️ Classic Benchmarks
 
 Most benchmarks will provide two streams: the `train_stream` and `test_stream`.
-Often, these are two parallel streams of the same length, where each experience is sampled from the same distribution (e.g. same set of classes). 
+Often, these are two parallel streams of the same length, where each experience is sampled from the same distribution (e.g. same set of classes).
 Some benchmarks may have a single test experience with the whole test dataset.
 
 Experiences provide all the information needed to update the model, such as the new batch of data, and they may be decorated with attributes that are helpful for training or logging purposes.
@@ -248,7 +248,7 @@ print(f"Experience {exp.logging().current_experience}")
 
 #### Classification
 
-classification benchmarks follow the `ClassesTimeline` protocol and provide attributes about the classes in the stream. 
+classification benchmarks follow the `ClassesTimeline` protocol and provide attributes about the classes in the stream.
 
 
 ```python
@@ -322,7 +322,7 @@ for exp in online_train_stream:
     print(f"\tsize: {len(exp.dataset)}")
 
     # in a training loop, here you would train on the online_train_stream
-    # here you would test on bm.valid_stream or bm.test_stream 
+    # here you would test on bm.valid_stream or bm.test_stream
 ```
 
 This completes the "_Benchmark_" tutorial for the "_From Zero to Hero_" series. We hope you enjoyed it!
--- a/docs/gitbook/from-zero-to-hero-tutorial/09_contribute-to-avalanche.md
+++ b/docs/gitbook/from-zero-to-hero-tutorial/09_contribute-to-avalanche.md
@@ -1,4 +1,4 @@
-]---
+---
 description: How to Contribute Back to the Avalanche Community
 ---
 
@@ -62,14 +62,14 @@ USE_GPU=False FAST_TEST=True python -m unittest discover tests -v
 
 **Contribute to the Avalanche documentation**
 
-Apart from the code, you can also contribute to the Avalanche documentation 📚!  We use [Jupyter notebooks](https://jupyter.org/) to write the documentation, so both code and text can be smoothly inserted, and, as you may have noticed, all our documentation can be run on [Google Colab](https://colab.research.google.com/notebooks/intro.ipynb)! 
+Apart from the code, you can also contribute to the Avalanche documentation 📚!  We use [Jupyter notebooks](https://jupyter.org/) to write the documentation, so both code and text can be smoothly inserted, and, as you may have noticed, all our documentation can be run on [Google Colab](https://colab.research.google.com/notebooks/intro.ipynb)!
 
 To contribute to the documentation you need to follow the steps below:
 
-1. The notebooks are contained in the folder `notebooks`. The folder structure is specular to the documentation, so do not create or delete any folder. 
-2. Detect the notebook that you want to edit and do all the modifications 📝 
-3. Commit the changes and open a pull request (PR). 
-4. If your pull request will be accepted, your edited notebooks will be automatically converted and uploaded to the official Avalanche website 🎊! 
+1. The notebooks are contained in the folder `notebooks`. The folder structure is specular to the documentation, so do not create or delete any folder.
+2. Detect the notebook that you want to edit and do all the modifications 📝
+3. Commit the changes and open a pull request (PR).
+4. If your pull request will be accepted, your edited notebooks will be automatically converted and uploaded to the official Avalanche website 🎊!
 
 
 


AntonioCarta:
Thank you! ping me once you are ready for a review.

jneuendorf:
@AntonioCarta Thanks for the quick reply!

I have a question not directly related to documentation: In a codebase using a past avalanche version there is this import `from avalanche.benchmarks.utils import make_classification_dataset`. From what I could see in the git history, this function was deprecated and no longer exists in avalanche. **What is the equivalent in the current version (0.5.0)?**

`from avalanche.benchmarks.utils import as_classification_dataset` seems to be the successor. However, it's not quite clear to me how it can be used equivalently. `_init_transform_groups` was used in `make_classification_dataset` and is still a "private" function in the current code, but there is no public function that uses the private one (only private ones, e.g `_make_taskaware_classification_dataset`). So the question is:

**What's the reason that only `transform_groups` are passed to `as_classification_dataset` instead of all the kwargs of `make_classification_dataset`?** Would you be fine with reintroducing the former kwargs and calling `_init_transform_groups`? Otherwise, I think the best alternative is to make `_init_transform_groups` public so users can call

```python
as_classification_dataset(
  dataset,
  transform_groups=init_transform_groups(
    target_transform=lambda x: x,  # or whatever
    ...
  ),
)
```

This could also be achieved via a class method (the constructor seems a bit complex already), e.g.

```python
as_classification_dataset(
  dataset,
  transform_groups=TransformGroups.from(
    target_transform=lambda x: x,  # or whatever
    ...
  ),
)
```

Kind regards 🙂 

PS: Providing a `dict` to `as_classification_dataset` as in the docs does not comply to the `TransformGroup` type.

AntonioCarta:
Thank you for highlighting these issues. The methods changed over time because at first `AvalancheDataset` only supported classification datasets. Now, a classification dataset is an AvalancheDataset where:
- the data returns triplets <x, y, t>
- it has a `targets` DataAttribute
- it has a `targets_task_labels` DataAttribute

Therefore, the goal of `as_classification_dataset` should be to make sure that the data has the correct attributes that are expected from classification datasets. Transformation groups in theory should be defined before this step. But I agree that it may be useful to define them at this stage as long as it's not too complex or confusing.
 
> Would you be fine with reintroducing the former kwargs and calling _init_transform_groups? Otherwise, I think the best alternative is to make _init_transform_groups public so users can call

I think it makes sense to add the arguments to `as_classification_dataset`. Transformation groups are a bit verbose to create from scratch.

> Otherwise, I think the best alternative is to make _init_transform_groups

I like this less. Classification transform groups are different from other transformations and TransformGroups tries to be agnostic to these differences (as much as possible). Also, It would not solve the verbosity problem.

> PS: Providing a dict to as_classification_dataset as in the docs does not comply to the TransformGroup type.

This is probably a mistake from the earlier API version.

AntonioCarta:
btw, if you have other comments about usability/pain point in the API/documentation do let me know. Fixing these issues takes a lot of time so we can't do everything quickly, but we try to keep track of it and improve over time.

AntonioCarta:
Hi @jneuendorf I see that the PR is still a draft but if there are no blocking issue I think we could merge it. Doc improvements can be easily split into multiple PRs.

jneuendorf:
> Hi @jneuendorf I see that the PR is still a draft but if there are no blocking issue I think we could merge it. Doc improvements can be easily split into multiple PRs.

Hi, your Github-comment email got lost between the other ones from Github like `coverall` etc. 🙈 You're correct in that some things could already be merged because later amendments could go into another PR.

_So feel free to review/cherry-pick the changes and decide which ones are ok for you._ 😉

jneuendorf:
I currently have a problem with training a benchmark from a custom dataset:

My custom dataset is a PyTorch `TensorDataset` that implements `ISupportedClassificationDataset`, i.e. the `targets` method. This way, it is compatible with `as_classification_dataset(...)`. However, when I try to train the following benchmark

```python
benchmark = class_incremental_benchmark(
    {
        'train': as_classification_dataset(train_dataset),
        'test': as_classification_dataset(test_dataset),
    },
    num_experiences=1,
)
```

I get the following error:

```
...
    cl_strategy.train(experience)
  File "avalanche/avalanche/training/templates/base_sgd.py", line 211, in train
    super().train(experiences, eval_streams, **kwargs)
  File "avalanche/avalanche/training/templates/base.py", line 162, in train
    self._before_training_exp(**kwargs)
  File "avalanche/avalanche/training/templates/base_sgd.py", line 291, in _before_training_exp
    self.make_train_dataloader(**kwargs)
  File "avalanche/avalanche/training/templates/base_sgd.py", line 456, in make_train_dataloader
    self.dataloader = TaskBalancedDataLoader(
                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "avalanche/avalanche/benchmarks/utils/data_loader.py", line 404, in __init__
    task_labels_field = getattr(data, "targets_task_labels")
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ClassificationDataset' object has no attribute 'targets_task_labels'
```

For me, this is a hint that there is some protocol mismatch: `class_incremental_benchmark` accepts a dataset created by `as_classification_dataset` but its training doesn't work. So am I correct that 

- either `ISupportedClassificationDataset` (and/or similar) should also required a `targets_task_labels` method
- or the base SGD should not rely on a data loader that depends on `targets_task_labels`

?

AntonioCarta:
> I think it would be helpful to extend the documentation with examples for creating datasets from typical sources:

yes, we need to update that notebook. Most of the time doing AvalancheDataset(your_data) should work. You don't need the old benchark builders (from tensors, from files,...) once you converted your data into an AvalancheDataset with appropriate attributes (needed to easily split it).


> For me, this is a hint that there is some protocol mismatch: class_incremental_benchmark accepts a dataset created by as_classification_dataset but its training doesn't work.

This is a missing feature. Basically, I implemented the datasets without task labels but many strategies still require them even when not strictly necessary. We have to fix this problem. Early version of datasets always had task labels, which was confusing for some people.

jneuendorf:
> yes, we need to update that notebook. Most of the time doing AvalancheDataset(your_data) should work. You don't need the old benchark builders (from tensors, from files,...) once you converted your data into an AvalancheDataset with appropriate attributes (needed to easily split it).

Okay. Based on your answer and [How to Create an AvalancheDataset](https://avalanche.continualai.org/how-tos/avalanchedataset/avalanche-datasets#how-to-create-an-avalanchedataset) I tried using 

```python
import torch
from torch.utils.data.dataset import TensorDataset
from avalanche.benchmarks.utils import AvalancheDataset

# Create a dataset of 100 data points described by 22 features + 1 class label
x_data = torch.rand(100, 22)
y_data = torch.randint(0, 5, (100,))

# Create the Dataset
torch_data = TensorDataset(x_data, y_data)

avl_data = AvalancheDataset(
  torch_data,
  data_attributes=[
    DataAttribute(y_data, "targets"),
    DataAttribute(ConstantSequence(0, len(y_data)), "targets_task_labels", use_in_getitem=True),
  ],
)
```

instead of

```python
# ...
avl_data = _make_taskaware_tensor_classification_dataset(x_data, y_data)
```

This is "with appropriate attributes" but quite cumbersome and typing does not work because `.targets` does not exist on `AvalancheDataset`, for example. Also the logic in `_make_taskaware_tensor_classification_dataset` is more powerful as it handles more cases.

## Suggestion 1: Dataset Hierarchy

So it would be really useful to add a dataset hierarchy to the documentation, i.e. the inheritance chain, along with the required data attributes (and more?).

## Suggestion 2: Dataset construction

We should think about whether the structure of _Suggestion 1_ is ideal. If so, it would be consequent to force the user to use the dataset class for the according scenario, e.g. 

```python
TaskAwareClassificationDataset(
  # ...
  task_labels=[1, 2, 3],
)
```

As with the "old benchmark builders", I would try to hide internal complexity from the user, such as `DataAttribute` or `ConstantSequence`.

More advanced benchmark builders could still exist with correct typing (e.g. via `@overload`) - also as `classmethod`s on `AvalancheDataset` for nicer imports.

AntonioCarta:
Thanks for the feedback.

The API for the benchmarks is quite cumbersome for a lot of reasons (changed over time, was supposed to be only an internal tool). We need a better API for attributes and transformations. If you have some proposals I would be happy to discuss it with you.
IMO it's better to discuss that in a separate issue so that we can track it separately.

> So it would be really useful to add a dataset hierarchy to the documentation, i.e. the inheritance chain, along with the required data attributes (and more?).

IMO the issue is that CL benchmarks are quite difficult and have an overloaded nomenclature. Many Avalanche users are new to CL so they don't have a complete understanding of the CL literature, and we assume they do. 

Avalanche models different benchmarks by adding attributes to datasets/experiences and splitting experiences appropriately. CL Benchmarks have 4 large properties: task-awareness, online, boundary-awareness, supervision. These are explained in the [FZTH](https://avalanche.continualai.org/from-zero-to-hero-tutorial/03_benchmarks). There is no strict hierarchy in Avalanche. Other CL frameworks tried to provide a strict hierarchy (e.g. sequoia) but we don't do it.

I would add other relevant examples to the FZTH notebook if you feel anything is missing or there needs to be more explanation.

> the logic in _make_taskaware_tensor_classification_dataset is more powerful as it handles more cases

The issue with that approach is that you are mixing the dataset creation with the benchmark creation and you need to make a lot of assumptions about the datasets. It is hard to generalize it to all possible settings (e.g. sequence classification, videos, graphs, ...). We also wanted to give the possibility of creating class-incremental benchmarks without task labels, which was confusing for many users. I agree with you that the new API is not as powerful yet, but I believe it's easier to provide high level methods to manage transforms/attributes than it is to provide truly general benchmark generators.

AntonioCarta:
> For me, this is a hint that there is some protocol mismatch: class_incremental_benchmark accepts a dataset created by as_classification_dataset but its training doesn't work. 

FYI #1652 fixes this error. Now you should be able to use class-incremental benchmarks without task labels with strategies (as long as they don't need task labels.

This is an example of the benchmark creation: 
```
    # create pytorch dataset
    train_data = MNIST(root=default_dataset_location("mnist"), train=True)
    test_data = MNIST(root=default_dataset_location("mnist"), train=False)

    # prepare transformations
    train_transform = Compose([ToTensor(), Normalize((0.1307,), (0.3081,))])
    eval_transform = Compose([ToTensor(), Normalize((0.1307,), (0.3081,))])
    tgroups = TransformGroups({"train": train_transform, "eval": eval_transform})

    # create Avalanche datasets with targets attributes (needed to split by class)
    da = DataAttribute(train_data.targets, "targets")
    train_data = make_avalanche_dataset(
        train_data, data_attributes=[da], transform_groups=tgroups
    )

    da = DataAttribute(test_data.targets, "targets")
    test_data = make_avalanche_dataset(
        test_data, data_attributes=[da], transform_groups=tgroups
    )

    # create benchmark
    bm = class_incremental_benchmark(
        {"train": train_data, "test": test_data}, num_experiences=5
    )

```

jneuendorf:
@AntonioCarta Thanks for your thorough reply. 😊 I will see, if the latest version works for me.

Generally, I agree on your thoughts. 👍 And I must admit that this issue's discussion has gone out of scope. Since I am neither an CL nor Avalanche expert, most of my comments were questions. 

So I thought Slack would be a better place for discussion and tried the [Slack Invite Link](https://join.slack.com/t/continualai/shared_invite/enQtNjQxNDYwMzkxNzk0LTBhYjg2MjM0YTM2OWRkNDYzOGE0ZTIzNDQ0ZGMzNDE3ZGUxNTZmNmM1YzJiYzgwMTkyZDQxYTlkMTI3NzZkNjU) but it appears to be dead. In case you still use Slack for discussion, could you provide a new invite link? 🙏

AntonioCarta:
> So I thought Slack would be a better place for discussion and tried the [Slack Invite Link](https://join.slack.com/t/continualai/shared_invite/enQtNjQxNDYwMzkxNzk0LTBhYjg2MjM0YTM2OWRkNDYzOGE0ZTIzNDQ0ZGMzNDE3ZGUxNTZmNmM1YzJiYzgwMTkyZDQxYTlkMTI3NzZkNjU) but it appears to be dead. In case you still use Slack for discussion, could you provide a new invite link? 🙏

where did you take this link? The one on the website seems to be working.

Anyway, I think it's better to discuss on github because slack only shows the last 90 days of messages. It's easier for us to track discussions here. I check them periodically and eventually close them if they don't lead to actionable items.

> I understand the difficulty/infeasibility of creating an abstraction for different benchmark scenarios. I meant the hierarchy of datasets only. Maybe adding (data)attributes like targets by benchmarks is common enough to result in a built-in subclass of AvalancheDataset (SupervisedDataset). If implemented correctly, the user could create a custom benchmark that utilizes a dataset which inherits from multiple built-in datasets.

For scenarios we have Protocols ([example](https://github.com/ContinualAI/avalanche/blob/d752103e838babda2e22e2771d5630561c9496c0/avalanche/benchmarks/scenarios/task_aware.py#L22)). Again, keep in mind that a scenario may need multiple protocols (e.g. task-aware + boundary-aware).

For datasets we would need some work to clean the hierarchy. We have some helpers ([example](https://github.com/ContinualAI/avalanche/blob/d752103e838babda2e22e2771d5630561c9496c0/avalanche/benchmarks/utils/utils.py#L185)) but I admit it's not clear for the user which one they should use.

jneuendorf:
> > So I thought Slack would be a better place for discussion and tried the [Slack Invite Link](https://join.slack.com/t/continualai/shared_invite/enQtNjQxNDYwMzkxNzk0LTBhYjg2MjM0YTM2OWRkNDYzOGE0ZTIzNDQ0ZGMzNDE3ZGUxNTZmNmM1YzJiYzgwMTkyZDQxYTlkMTI3NzZkNjU) but it appears to be dead. In case you still use Slack for discussion, could you provide a new invite link? 🙏
> 
> where did you take this link? The one on the website seems to be working.

I tried both, the one directly from the sidebar and the one at the bottom of [The People](https://avalanche.continualai.org/about-us/the-team) - also across different machines. But none have worked.

> For scenarios we have Protocols ([example](https://github.com/ContinualAI/avalanche/blob/d752103e838babda2e22e2771d5630561c9496c0/avalanche/benchmarks/scenarios/task_aware.py#L22)). Again, keep in mind that a scenario may need multiple protocols (e.g. task-aware + boundary-aware).
> 
> For datasets we would need some work to clean the hierarchy. We have some helpers ([example](https://github.com/ContinualAI/avalanche/blob/d752103e838babda2e22e2771d5630561c9496c0/avalanche/benchmarks/utils/utils.py#L185)) but I admit it's not clear for the user which one they should use.

I tried to get a better picture of the structure and relations between scenarios, streams experiences, and their interfaces/protocols, ABCs and implementations. Here is what I came up with



[Link to editable PlantUML](https://www.plantuml.com/plantuml/uml/jLZRRjiu4xxNKn3vB-eMZe3_LWYyTLC5QsphBcBj0J0CWfB6DhSIgIMf9kbktdqfYpQbng7aADQLHCxXc-5m3del5H6gp3DlIrD0a1Vg0UMGno2a9q4X2GKHH761VCzxS_dYntk3PlCelBsPNeSBJ_4Dg2q8z8-7z4EP0h4c2I2P02E2Skah8hzYSaS4-BD8lq-hLqDVFRlzZCIG1HN_aNKPe8nAjTIxgztse_UiDiuoohIc3xnaAH4FFsgZ4g-fa0hBCeRx8a0nvrbpDoFFRUf5B9K0akCCs4PjstX9ae2KU0SxhhLJa5Q62xKi2n3Vg8JKdrrh119aH7FGfZIn9_KEfWohhOQg8OBMmH9eESWyXO1lb9SI6xOsJLkY161OShmcuXZPkbIbW2E2JhQBhcb25EMiym2_i1-ivu9kA6iPRRasmFHgWlSh_lMdy41eK6gShN0rvXyFLHkfLrSvlQTiCa4qBpB8WIdP5Tv7fHlBR6Gwf3l3sKe_xkgsfvqgt6VaiC8SJq_uDNKlt3RoJmPkMRt7eJdkGpY__ldpzFONpm1sXTh2l5v5-dsN8H3SAs2fcy_GJamqrhb51YYbYLhghp4wo5jL5ZcFmgBGVakX490G1QbsNZyevdMetnRHRJZzRA5s6L-vpDxKVbWj_TvrZEGGL0wi7GbgdwWfM_D62eurqpOdukjFGKztQBUJM0iZYKATHYteTqXVoR9J8iXu0LTr2UFHsxemdGOT_YuhlUT3x_QMrnqo8Tc0E2i-jqkCbtK1aW5IPP71iWLYZ7pVNmrFFU6VluQtCvr_mcwVj9nWYbef133LAfHLEDWQQYv7q2UqbWl8rgkwmYj1AAkYgrghVV6DPAs5Z6-qWir-xOKrTyZHZjsoUvVcuI3QiWTKyE58t0fw5FgcoEhgVIIAI507DAMYcVHJlUfFTKmGdMR1q1oxQsczPI4RMfS73HBFWkqbS3Zg_hVFF0PCWAnAquBsm_w9V7-8ghlsEyaMHKQL0j4GRo-s3D1Cxv4uW-KJWsrpVbBNTQvWineydHM6viYFqsYw22FFJX3cB8qlKrXJHglOblxCkEOgmWcV76hb1XHLaEjwILaAzxiagTDMHNIycfwax0-gjX4HMedSwTMzNFqPFHIWlsryTSioa5KjBa3xmhHiYrprnhWeYu8B1Max3HXr1gyXaZDcByDfXlS3uy90pwmx0YQvCDmlE94TU-j4D6pRgxiIZ8b6EzdqLclhy9mpQDPXkt2dFO4r7PDhBd8vsWXU5lZG9ZT94ywqwqUygj7DKYxBM49x_M2O0-lpnsbFKd6HR7sprzC5H-MsuIGj4cp4k15rLu8-AcpdCaINFM0j3XT4AH0CKvRI1FPZ8MLgRCDDVwrMbgKkElfj0SeqReydmLySadQt8ODSChW0tFmsluvcN-OBpz5GtfGisOMF7edHw6PSKFRrBPgO6c0kYVtDEEBnNphI6vFql-ZG_5Uljj9PdrEMY5tqaWmVnijHd3VbMA5cf7IKupwOzx_jbAceBk0xBWPt6gXKH4EN3PXwO3c2pTsUT24yICThmNDisENetRjwx6WfNBQ6kYDDp6eeVjCzTg6MLP-5Qrep7nclEi3L-LDlAirJnjN8LHkTbvcYrQSq49_mDfGcd2dAIbvAN6ad0gUGa0U3mWXmeHWySTw6dwRrPUhrKmHT_Fy2lRkyH1VpYpupPB3_jUNr8EulF6mEk0g4vtN_6hjS_az0Eaw9t4AwkfoSD4AW09KIF7UgUMKLDZESIuNnrIkBTx2SNv5hZdeDdTxf_SPJlIUsK07AgLvb6-yziBJCi_y0)

This is neither complete nor mirroring the actual code base (`ClassesTimelineCLExperience`, conceptual interfaces according to the wording and for consistency).

### Some things, I figured out

1. Everything seems to be working with `DatasetExperience` rather than `CLExperience`
2. Some dataset definitions are not used
3. Some experience protocols are not used
4. Some protocols are basically ABCs as they provide method/property implementations.

### Remarks / Suggestions

- I like the idea of having protocols, as they let the user ship their own implementations.
- Avalanche experience properties should be ABCs/mixins implementing the protocols. This would very nicely support the idea "that a scenario may need multiple protocols (e.g. task-aware + boundary-aware)."
- Merge `CLExperience` and `DatasetExperience` to be used for implementations, since it is used as a single unit. There could be an experience protocol for typing, as well.
- Adjust dataset definitions according to the above points. The protocol/class structure could match the one of experiences. This way, it would be more obvious for the user what parts are required for a certain settings, e.g.
  - task-aware + boundary-aware $\leadsto$
    - scenario type `class Scenario(scenarios.TaskAware, scenarios.BoundaryAware): ...`
    - dataset type `class Dataset(datasets.TaskAware, datasets.BoundaryAware): ...`
  - Of course, this could also be implemented as factory functions like already done. With overloading, intersection typing should be doable (basically function composition instead multi-inheritance classes). 

### Let me know what you think 😉

AntonioCarta:
> Everything seems to be working with DatasetExperience rather than CLExperience

The CLExperience provides a bit of additional flexibility. We have some minimal support for reinforcement learning, which is a scenario where we don't have a dataset. We also have avalanche-rl, which was supported by an older avalanche version but it is currently unmantained. For example, methods like progressive neural networks should work with task-aware RL scenarios without any modifications.

> Avalanche experience properties should be ABCs/mixins implementing the protocols. This would very nicely support the idea "that a scenario may need multiple protocols (e.g. task-aware + boundary-aware)."

Yes, but I would keep this as an internal API. Users should be using high-level methods, because I think most people would be very confused by mixins.

> Adjust dataset definitions according to the above points. The protocol/class structure could match the one of experiences. This way, it would be more obvious for the user what parts are required for a certain settings, e.g.

I tend to agree with this, but keep in mind that some protocols don't need to be propagated. For example, scenarios/experiences can be BoundaryAware, but the dataset is not BoundaryAware since this is a property of the stream, not the data (unlike task and class labels).
There are also some older benchmarks that respect the protocols but were implemented before their definition so they have a slightly different class structure. I would not touch those to avoid introducing bugs.

> With overloading, intersection typing should be doable (basically function composition instead multi-inheritance classes)

I'm not an expert on python typing but last time I checked I remember that there were some serious limitations (e.g. protocols runtime checks not checking attributes or full signature of functions). We also have the additional challenge of having to support older python versions (for example right now the CI fails due to typing import errors).

Issue #1611, opened 2024-03-06
strategy API design


AntonioCarta:
@AlbinSou to continue the discussion that we had on Monday, this is a pseudocode of what I imagined a "standalone" Avalanche strategy would be like. The idea is that the whole training loop must fit in a single file. Components (optimizer, dataloaders, losses, models) are all independent and composable. Components are updatable object that receive the strategy state and current experience only at predefined points. This mitigates the issue that many have with Avalanche strategies, that is having to navigate the template hierarchy to understand the training loop.

If we go in this direction, it is important to do it in a manner that is cheap to maintain without ending up with broken code over the next few releases (like it happened with many examples).

Some things will not be easily definable as simple "updatable" objects. This is fine, it just means that they will be more integrated in their training loop.

Let me know what you think. Of course we don't need to convert what we already have, this would be mostly to support usage of Avalanche together with other frameworks. For example, this would make it possible to use Avalanche in Lightning.

```
def my_cool_method(exp, agent_state=None, **extra_args):
    # defining the training for a single exp. instead of the whole stream
    # should be better because this way training does not need to care about eval loop.
    # agent_state takes the role of the strategy object
    
    if agent_state is None:
        # first exp, init stuff
        # we could even do partial initialization (if not too tricky)
        # maybe_init does not touch already defined arguments
        agent_state.replay = maybe_init(MyReplayBuffer(memory_size=a_lot))
        agent_state.reg_loss = maybe_init(BetterLwF(with_tricks=True))
        agent_state.model = maybe_init(MyModel(...))
        
        # no partial init for optim and scheduler because they depend on the model
        agent_state.opt = MyCLOptimizer(model.parameters(), lr=0.001)  # could be GEM, OWM, ...
        agent_state.scheduler = MyScheduler(opt, ...)

    # this is the usual before_exp 
    updatable_objs = [agent_state.replay, agent_state.reg_loss, agent_state.model]
    [uo.pre_update(exp, agent_state) for uo in updatable_objs]

    dl = AvlDataLoader(exp.dataset, agent_state.replay.buffer, batch_size=128, ...)
    for x, y in dl:
        opt.zero_grad()
        yp = agent_state.model(x)
        l = CrossEntropyLoss(yp, y)
        l += agent_state.reg_loss()
        l.backward()
        l.step()

    # this is the usual after_exp 
    updatable_objs = [agent_state.replay, agent_state.reg_loss, agent_state.model]
[...]

PR #1633, opened 2024-04-03
Updatable objects

AntonioCarta:
attempt at #1611.

Right now there is:
- a single API for objects that require pre/post training experience hooks (models, optimizers, losses, buffers..)
- everything is encapsulated in the `Agent` object, which keeps track of CL state and provides some basic utilities to update them or add functional hooks
- `MetricCollector`: a new object to manage metrics. This is needed because the current metric system is tightly coupled to the templates. `MetricCollector` is a simplified metric system but it should provide the basic functionality with a simpler API.
- everything is "documented" via the `updatable_objects.py` example.

I'm opening the PR for comments right now, just to understand if this may be useful or not. I'm finding it conveniente for some simple experiments that I'm running.

Regarding compatibility, I didn't find any major roadblocks. Everything seems easy to integrate, though some things may need to be adapted. For example, some methods like GEM may be rewritten as optimizer to make their use transparent with this API.


avalanche/evaluation/collector.py
```diff
@@ -0,0 +1,173 @@
...
+        """Returns a metric value given its name and aggregation method.
+
+        :param name: name of the metric.
+        :param time_reduce: Aggregation over the time dimension. One of {None, 'last', 'mean'}, where:
+            - None (default) does not use any aggregation
+            - 'last' returns the last timestep
+            - 'mean' averages over time
+        :param exp_reduce: Aggregation over the experience dimension. One of {None, 'sample_mean', 'experience_mean'} where:
+            - None (default) does not use any aggregation
+            - `sample_mean` is an average weighted by the number of samples in each experience
```
AlbinSou, line 92: Nice ! Maybe we could add a class means too, or is that too specific ? I think that in general in continual learning this option is useful when experiences are not of the same size, and because we often use classical benchmarks that have the same number of samples per class in the test set we often think the sample mean and the class mean is the same, but I think what we want in reality is the class mean in that case. I don't know if you were thinking of another application case. 

I would put both and add something in the doc to say which one should be used in what situation

avalanche/evaluation/collector.py
```diff
@@ -0,0 +1,173 @@
...
+        # iterating over the stream may be expensive
+        self._stream_len[stream.name] = 0
+        coeffs = []
+        for exp in stream:
+            self._stream_len[stream.name] += 1
+            coeffs.append(len(exp.dataset))
+        coeffs = np.array(coeffs)
+        self._coeffs[stream.name] = coeffs / coeffs.sum()
+
+    def update(self, res, *, stream=None):
```
AlbinSou, line 55: Should we add a "step" argument here ? I see that we store all metrics inside this metric_res but if we don't know at what step one metric was logged the previous metrics are kind of hard to use (because we don't know where they come from). I don't know how we could handle that, usually what I do is that I have a global (atomic) step and then if I want to have another kind of step I log it as a metric which is also a function of the atomic step.

In the classical avalanche metrics we always have a "plot_x" argument in the MetricValue object

avalanche/core.py
```diff
@@ -1,13 +1,153 @@
...
+    Protocol, which requires the `pre_adapt` and `post_adapt` methods.
+    To call the pre/post adaptation you can implement your training loop
+    like in the following example:
+
+    .. code-block::
+
+        def train(agent, exp):
+            agent.pre_adapt(exp)
+            # do training here
+            agent.post_update(exp)
```
AlbinSou, line 42: post_adapt

AlbinSou (commented):
Looks good so far. I was a  bit concerned wether some objects would need to implement something else than "before_training_exp" and "after_training_exp" but I don't think so.  For instance for optimizers the "before_backward" call would be replaced by overriding the step() method. In general I think before_training_exp and after_training_exp are sufficient to account for every cases.

avalanche/core.py
```diff
@@ -1,13 +1,153 @@
...
+            is called during the adaptation. Useful for debugging.
+        """
+        self._updatable_objects = []
+        self.verbose = verbose
+        self._pre_hooks = []
+        self._post_hooks = []
+
+    def __setattr__(self, name, value):
+        super().__setattr__(name, value)
+        if hasattr(value, "pre_adapt") or hasattr(value, "post_adapt"):
```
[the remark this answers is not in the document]
AntonioCarta: We would need two protocols for pre-adapt and post-adapt to keep them independent. It's doable but I preferred to keep a single Protocol mostly for documentation purposes. I think the `hasattr` check is self-explanatory already when it's used.

avalanche/evaluation/collector.py
AntonioCarta: This is a good point. Maybe I can add an argument for a custom reduce function to support this use case. I prefer not to add reduce methods that depend on classification.

avalanche/evaluation/collector.py
AntonioCarta: My idea was that with this API the choice of when `update` is called is left to the user, so the `plot_x` is not necessary anymore. I admit that `plot_x` may still be useful in some cases. I'm also not sure how we should do the time aggregation if `plot_x` can skip timesteps.

AlbinSou:
As a side comment. Do you think it can happen that one updatable object depends on another ? I am thinking in particular about the optimizer. If we want to use the existing functions for updating optimizer we would need to encapsulate the optimizer into an Updatable object, but then it would need to be updated after the model. In that case we could add some dependency system to each object to make sure they are updated in the correct order.

AntonioCarta:
@AlbinSou I added the custom weights mode for the MetricCollector and a wrapper for the optimizers. I think the PR is ready to be merged if you agree.

avalanche/models/dynamic_optimizers.py
```diff
@@ -31,6 +33,55 @@
...
+
+    def step(self):
+        self.optim.step()
+
+    def pre_adapt(self, agent: Agent, exp: CLExperience):
+        """Adapt the optimizer before training on the current experience."""
+        update_optimizer(
+            self.optim,
+            new_params=dict(agent.model.named_parameters()),
+            optimized_params=dict(agent.model.named_parameters())
```
AlbinSou: This should be the previous dict of optimized params (without the potentially new parameters), as returned by this same function (See how it's used in batch_observation.py).

PR #1663, opened 2024-08-28
When import the library, appears an error said that cant import DwsCo…

skiaxtro:
…nvBlock.

The problem is at "mobilenetv1.py" file at line 25. There is  import "from pytorchcv.models.common import DwsConvBlock". The problem is that in common, there is not DwsConvBlock. Instead "DwsConvBlock" is in "pytorchcv.models.common.conv" . 

So to solve the problem must change line 25 with "from pytorchcv.models.common.conv import DwsConvBlock"

### DIFF
--- a/avalanche/models/mobilenetv1.py
+++ b/avalanche/models/mobilenetv1.py
@@ -22,7 +22,7 @@ from pytorchcv.models.mobilenet import mobilenet_w1
 try:
     from pytorchcv.models.mobilenet import DwsConvBlock
 except Exception:
-    from pytorchcv.models.common import DwsConvBlock
+    from pytorchcv.models.common.conv import DwsConvBlock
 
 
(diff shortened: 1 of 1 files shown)


AntonioCarta:
Thank you for the PR!
It seems that the current solution is not robust to different versions of pytorchcv. Tests are failing with `ModuleNotFoundError: No module named 'pytorchcv.models.common.conv'; 'pytorchcv.models.common' is not a package`.

samuelebodon:
> …nvBlock.
> 
> The problem is at "mobilenetv1.py" file at line 25. There is import "from pytorchcv.models.common import DwsConvBlock". The problem is that in common, there is not DwsConvBlock. Instead "DwsConvBlock" is in "pytorchcv.models.common.conv" .
> 
> So to solve the problem must change line 25 with "from pytorchcv.models.common.conv import DwsConvBlock"

I can confirm, this fix works!

PR #1672, opened 2024-12-04
Added SER Strategy

AlboAlby00:
I Implemented the SER (Strong Experience Replay) algorithm from the paper "Continual Learning with Strong Experience Replay" https://arxiv.org/pdf/2302.11346 .
It is a modified version of DER++, so my implementation of SER is strongly based on the DER++ implementation here on avalanche. 
I plan to create a PR on [continual-learning-baselines](https://github.com/ContinualAI/continual-learning-baselines) when I will be able to replicate the paper results. 
On my partial experiments so far SER outperforms DER++ on the SMNIST and CIFAR10 benchmarks in a class incremental setup.

### DIFF
--- a/avalanche/training/supervised/ser.py
+++ b/avalanche/training/supervised/ser.py
@@ -1,0 +1,216 @@
+from collections import defaultdict
+from typing import (
+    Callable,
+    List,
+    Optional,
+    Union,
+)
+
+import torch
+import torch.nn.functional as F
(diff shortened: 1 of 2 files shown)


AntonioCarta:
Hi, sorry for the late reply. The code looks ok. Did you reproduce the paper experiments? I'm not familiar with the paper so I can't evaluate if your implementation matches the paper exactly at the moment.

Issue #1678, opened 2025-01-22
Error when working with multimodal data: 'dict' object has no attribute 'to'


spartanjoax:
🐛 **Describe the bug**
I am working with a multimodal signal dataset. It is composed of time-series data (signals) and structured data (process parameters). The samples are returned as a dict with two keys (x and proc_data), which are then fed to the model. I am getting an error during training, as Avalanche attempts to move the minibatches to the GPU device. The error I am getting is the following:

File "/workspace/example/continual_training.py", line 325, in main
    strategy.train(scenario.train_stream)
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/supervised/joint_training.py", line 152, in train
    self._before_training(**kwargs)
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/templates/base.py", line 326, in _before_training
    trigger_plugins(self, "before_training", **kwargs)
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/utils.py", line 75, in trigger_plugins
    getattr(p, event)(strategy, **kwargs)
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/templates/base_sgd.py", line 616, in before_training
    self._peval(strategy, **kwargs)
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/templates/base_sgd.py", line 637, in _peval
    strategy.eval(el, **kwargs)
  File "/opt/conda/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/templates/base_sgd.py", line 228, in eval
    super().eval(exp_list, **kwargs)
  File "/opt/conda/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/templates/base.py", line 212, in eval
    self._eval_exp(**kwargs)
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/templates/base_sgd.py", line 232, in _eval_exp
    self.eval_epoch(**kwargs)
  File "/opt/conda/lib/python3.11/site-packages/avalanche/training/templates/base_sgd.py", line 270, in eval_epoch
[...]

PR #1685, opened 2025-02-24
Added regression metrics and support for multi-modal data using dictionaries

spartanjoax:
Hi all,

I have created a fork to contribute to Avalanche. My problem is a regression problem using multi-modal data and noticed that Avalanche did not support this natively. I have added RMSE and R2 metrics, as well as forgetting metrics for both, and modified also the library to handle batches that use dictionaries of tensors instead of just tensors, to handle multi-modal data. This second point address issue #1678. I have blackend the code as well. Please let me know if everything looks good.

Best regards,
Joaquín


spartanjoax:
It seems the readthedocs failed because it is using a prior version of sklearn which uses the deprecated mean_squared_error function to calculate the RMSE and my proposed metric uses the new version which calls directly the new root_mean_squared_error. Should I modify my code to use the deprecated function?

AntonioCarta:
Thanks, the changes look good. Can you also add the new classes to the API documentation?

> Should I modify my code to use the deprecated function?

is it possibile to support both? Either by checking the scikit-learn version or by try/catching the import. Otherwise we can add a version constraint, although it's better to avoid it if possible.

AntonioCarta:
Hi, it seems that there is an import error:
```
  File "/__w/avalanche/avalanche/avalanche/evaluation/metrics/regression_forgetting.py", line 19, in <module>
    from metrics import TaskAwareRMSE, TaskAwareR2
ModuleNotFoundError: No module named 'metrics'
```

spartanjoax:
I'll fix it right away

spartanjoax:
I have changed RMSE and R2 to be calculated directly with Torch functions. I also fixed an error in the joint training strategy, as it was not training.

spartanjoax:
Hi @AntonioCarta , ¿could you please help me understand what test failed? Thanks in advance.

AntonioCarta:
Hi, it's just the style checker. I will fix that in a separate commit. Thanks for your contibution

